PR #26, opened 2022-12-05
Adds some pagination tests to the SDK

mstanbCO:


pkg/pagination/pagination_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	userResourceType := "user"
+	secondPageToken := "page=2"
+	secondPageState := PageState{
+		Token:          secondPageToken,
+		ResourceTypeID: userResourceType,
+	}
+
+	bag.push(firstPageState)
+	bag.push(secondPageState)
```
laurenleach, line 47: i think we want to push `secondPageState` before `firstPageState` bc its a stack

laurenleach: why is `page=2`?

jirwin (approved):
Thanks!

PR #30, opened 2022-12-09
Builder Options for Entitlements, Grants, and Resources.

mstanbCO:
Adds options to the Entitlement, Grant, and Resources new object builders.


pkg/types/entitlement/entilement.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+}
+
+func WithDisplayName(displayName string) EntitlementOption {
+	return func(g *v2.Entitlement) {
+		g.DisplayName = displayName
+	}
+}
+
+func WithDescriptionName(description string) EntitlementOption {
```
jirwin, line 33: This should probably just be `WithDescription`

pkg/types/entitlement/entilement.go
```diff
@@ -0,0 +1,37 @@
+package entitlement
+
+import (
+	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
+	"github.com/conductorone/baton-sdk/pkg/annotations"
+	"google.golang.org/protobuf/proto"
+)
+
+type EntitlementOption func(*v2.Entitlement)
```
jirwin, line 9: What do you think of moving all of the `New*` functions into these smaller type specific packages so we wouldn't have any of this in the `sdk/` package?

mstanbCO: Yeah, I think that makes more sense. I'll get a commit added.

pkg/types/resource/resource_test.go
```diff
@@ -133,44 +63,28 @@ func TestNewGroupResource(t *testing.T) {
 	require.True(t, ok)
 	require.Equal(t, "v1", v1ID.Id)
 
-	roleTrait, err := GetRoleTrait(gr)
+	roleTrait, err := sdk.GetRoleTrait(gr)
```
jirwin, line 66: Probably better for another PR, but I think it makes sense to move these trait helpers to the types package as well. The `sdk` package doesn't make much sense I think, so moving more out of it feels good.

jirwin (approved):

PR #31, opened 2022-12-15
Reorganize trait helpers

jirwin:
Move the trait helpers out of the `sdk/` package into the `types/resource` package.

Update the trait helpers to use functional options for further customization.


mstanbCO (approved):
Looks great!

PR #34, opened 2022-12-27
Add account type to user trait, refactor options to be more flexible

jirwin:
When syncing user resources, we want some way to designate whether or not this user is a service account or not.


proto/c1/connector/v2/annotation_trait.proto
```diff
@@ -47,6 +47,7 @@ message UserTrait {
   ];
   google.protobuf.Struct profile = 3;
   c1.connector.v2.AssetRef icon = 4;
+  bool is_service_account = 5;
```
morgabra, line 50: Would this be better as an enum? Or something similar to `status` where we also give a text-field next to it?

Something that came up recently was `system` accounts, which aren't actually assumable by any humans but are still addressable/have permissions. (GCP has this as an example). So that's (at least) `human`, `service`, `system` - but I can imagine a bunch of other descriptors we might want to capture (`api`, `shared`, etc).

pquerna: yes i tend to like an enum, undefined defaults to human?

mstanbCO (approved):
Looks good to me :)

PR #36, opened 2023-01-26
Adds an option to set the principal resource for a grant when using NewGrant

mstanbCO:
This PR adds an option method to set the principal resource on a grant when using the `NewGrant` function.

### DIFF
--- a/pkg/types/grant/grant.go
+++ b/pkg/types/grant/grant.go
@@ -21,6 +21,12 @@ func WithAnnotation(msgs ...proto.Message) GrantOption {
 	}
 }
 
+func WithPrincipal(r *v2.Resource) GrantOption {
+	return func(g *v2.Grant) {
+		g.Principal = r
+	}
+}
+
 // NewGrant returns a new grant for the given entitlement on the resource for the provided principal resource ID.
 func NewGrant(resource *v2.Resource, entitlementName string, principal *v2.ResourceId, grantOptions ...GrantOption) *v2.Grant {
 	entitlement := &v2.Entitlement{


pkg/types/grant/grant.go
```diff
@@ -21,6 +21,12 @@ func WithAnnotation(msgs ...proto.Message) GrantOption {
 	}
 }
 
+func WithPrincipal(r *v2.Resource) GrantOption {
+	return func(g *v2.Grant) {
+		g.Principal = r
```
jirwin, line 26: Because `NewGrant()` requires a principal resource ID today, I wonder if we should check and compare the IDs here?

I think because a principal is required for every grant, we want the principal to stay part of the function call versus via an option. What if this function takes `principal interface{}`, and we type assert the value. If it is a resource ID, we set it to an empty resource with the ID set, else if we get a pointer to a resource, we could set that explicitly. 

Another option would be that we could add a field to the messages like `bool baton_resource = 1;` to the resource id and resource messages. This would generate a new `BatonResource() bool` method on those messages, and then we could make an interface to target that to keep some compile type safety.

The interface we match on could be something like:

```go
type BatonResourceLike interface {
  proto.Message // embed proto.Message so we're sure we are only getting protos setn
  BatonResource() bool
}
```

mstanbCO: Doing the principal interface{} seems pretty simple, and think it would do the trick. The second option also seems fine to me if that's the option you prefer. I wish there was a clean way to just have an option to set the trait on the principal, but being able to set the entire principal object is still useful.

jirwin:
Closed for #37

PR #37, opened 2023-01-31
Allow grants to be created with a resource ID or a resource

jirwin:
- Adds new `baton_resource` field to `Resource` and `ResourceId` messages. 
- Adds a new `GrantPrincipal` interface that matches on the new `baton_resource` field.
- Handle a `GrantPrincipal` being passed to `NewGrant`.


mstanbCO (approved):
Looks good to me! Thanks for adding this

PR #42, opened 2023-02-21
Update dependencies - Feb 2023

jirwin:
I ran `make update-deps`


Makefile
```diff
@@ -4,7 +4,7 @@ lint:
 
 .PHONY: update-deps
 update-deps:
-	GOPRIVATE=github.com/conductorone/baton-sdk go get -d -u ./...
+	go get -d -u ./...
```
jirwin, line 7: SDK is no longer private, so we can drop this env var.

pquerna (approved):

PR #43, opened 2023-02-24
Automatically cleanup old sync data in the c1z after a sync

jirwin:
In order to prevent c1z files from growing unbounded, this change will perform automatic cleanup on the contents of the local db after a sync.

By default it will keep the last sync performed, and the previous sync if available. This allows diffing to continue to work without storing all of the data.

In order to disable the automatic cleanup, the user can set `BATON_SKIP_CLEANUP=1` in the environment, and cleanup will not happen. Users can also set `BATON_KEEP_SYNC_COUNT=5` in order to increase the number of syncs from the default 2. Sync count is clamped to 1 as a minimum.

### DIFF
--- a/internal/dotc1z/sync_runs.go
+++ b/internal/dotc1z/sync_runs.go
@@ -5,11 +5,14 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
 	"time"
 
 	"github.com/doug-martin/goqu/v9"
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
 	"github.com/segmentio/ksuid"
+	"go.uber.org/zap"
 )
 
 const syncRunsTableVersion = "1"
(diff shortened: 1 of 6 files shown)


internal/dotc1z/sync_runs.go
```diff
@@ -351,3 +354,93 @@ func (c *C1File) EndSync(ctx context.Context) error {
...
+		syncLimit = int(customSyncLimit)
+	}
+
+	if len(ret) <= syncLimit {
+		return nil
+	}
+
+	l.Info("Cleaning up old sync data...")
+	for i := 0; i < len(ret)-syncLimit; i++ {
+		err = c.DeleteSyncRun(ctx, ret[i].ID)
```
morgabra, line 407: Is ret implied to be ordered? Maybe explicitly ordering it so we don't accidentally delete more recent syncs would be nice?

internal/dotc1z/sync_runs.go
```diff
@@ -351,3 +354,93 @@ func (c *C1File) EndSync(ctx context.Context) error {
...
+	if customSyncLimit, err := strconv.ParseInt(os.Getenv("BATON_KEEP_SYNC_COUNT"), 10, 64); err == nil && customSyncLimit > 0 {
+		syncLimit = int(customSyncLimit)
+	}
+
+	if len(ret) <= syncLimit {
+		return nil
+	}
+
+	l.Info("Cleaning up old sync data...")
+	for i := 0; i < len(ret)-syncLimit; i++ {
```
morgabra, line 406: small nit: this and the above logic might be a little cleaner with something like:
```
for _, run := range ret[syncLimit:] { ...
```

internal/dotc1z/sync_runs.go
```diff
@@ -351,3 +354,93 @@ func (c *C1File) EndSync(ctx context.Context) error {
...
+	var ret []*syncRun
+
+	pageToken := ""
+	for {
+		runs, nextPageToken, err := c.ListSyncRuns(ctx, pageToken, 100)
+		if err != nil {
+			return err
+		}
+
+		for _, sr := range runs {
```
morgabra, line 383: Is this desired behavior? You could start and never complete multiple syncs right? Maybe make sure to keep the current 'active' one but all the others could go?

morgabra (approved):

PR #45, opened 2023-03-15
Daemon Refactoring

jirwin:
The goal of this change is to refactor the CLI harness to take ownership of the process and introduce the idea of a 'daemon' mode for connectors. The SDK takes over the main run loop of the connector, working off of a serial work queue. Today this includes a single task type for Syncs, but you can imagine adding `Grant` and `Revoke` tasks as well. 

This is the first step on the path to allowing connectors to run in "service mode" where they will communicate with an upstream API in order to get work to do and submit the results of each task.

This is a breaking change for connectors consuming the SDK. They will be required to make updates to begin using this version of the SDK. 

## main loop changes
The primary change here is that the SDK cli package takes control of running the connector implementation instead of the connector implementation itself. Before this change, when using the `NewCmd()`, SDK users would pass a run function defined by the connector. This function looked something like this:
```
// run is where the process of syncing with the connector is implemented.
func run(ctx context.Context, cfg *config) error {
	l := ctxzap.Extract(ctx)

	c, err := getConnector(ctx, cfg)
	if err != nil {
		return err
	}

	r, err := sdk.NewConnectorRunner(ctx, c, cfg.C1zPath)
	if err != nil {
		l.Error("error creating connector runner", zap.Error(err))
		return err
	}
	defer r.Close()

	err = r.Run(ctx)
	if err != nil {
		l.Error("error running connector", zap.Error(err))
		return err
	}

	return nil
}
```

With this change, the SDK is in charge of the run command for the 'main loop' of the connector process, meaning that implementors only need to provide a factory method.

## on demand mode
in order to keep the existing functionality(running without the new `-d` flag), this also implements 'on-demand' mode. If we don't have the daemonize flag set, we will still bring up the connector as if we were daemonized, but we manually push a sync task into the work queue, and then explicitly quit after running the single task. 

## naive task manager
For testing and debugging, I've implemented a 'naive' task manager. It is very basic and implements the work queue as an in-memory slice. It also supports injecting a single sync task into its queue by signalling the process with `USR1`. Clearly not for production use, and will be removed eventually.

## connector lifecycle
Previous to this change, when the connector was run it would immediately spin up the connector server as a child process, run a sync, and then shut everything down. Now the connector server is only spun up to perform a task, and then is shutdown. A new process will be spun up to process each task.

### DIFF
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -1,73 +1,52 @@
-package sdk
+package connectorrunner
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"os"
 	"os/signal"
 	"time"
 
 	ratelimitV1 "github.com/conductorone/baton-sdk/pb/c1/ratelimit/v1"
-	"github.com/conductorone/baton-sdk/pkg/dotc1z/manager"
-	"github.com/conductorone/baton-sdk/pkg/sync"
+	"github.com/conductorone/baton-sdk/pkg/connectorrunner/tasks"
+	sdkSync "github.com/conductorone/baton-sdk/pkg/sync"
 	"github.com/conductorone/baton-sdk/pkg/types"
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	"github.com/conductorone/baton-sdk/internal/connector"
-	"github.com/conductorone/baton-sdk/pkg/connectorstore"
 )
 
 type connectorRunner struct {
-	syncer  sync.Syncer
-	store   connectorstore.Writer
-	manager manager.Manager
+	cw           types.ClientWrapper
+	dbPath       string
+	onDemandMode bool
+	tasks        tasks.Manager
 }
 
-func (c *connectorRunner) shutdown(ctx context.Context) error {
-	logger := ctxzap.Extract(ctx)
-
-	err := c.Close()
-	if err != nil {
-		// Explicitly ignoring the error here as it is possible that things have already been closed.
-		logger.Error("error closing connector runner", zap.Error(err))
-	}
-
-	err = c.manager.SaveC1Z(ctx)
-	if err != nil {
-		logger.Error("error saving c1z", zap.Error(err))
-		return err
-	}
-
-	err = c.manager.Close(ctx)
-	if err != nil {
-		logger.Error("error closing c1z manager", zap.Error(err))
-		return err
-	}
-
-	return nil
-}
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -7,7 +7,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.19.x
+          go-version: 1.20.x
```
jirwin, line 10: Using new 1.20 only features, so have to drop support for old versions. Will need to cascade this change down to each connector.

pkg/connectorrunner/runner.go
```diff
@@ -1,86 +1,154 @@
...
-	go func() {
-		for range sigChan {
-			cancel()
+		// No work for us to do, go to sleep for a bit and try again
+		if nextTask == nil {
+			l.Debug("no available tasks, going to sleep for 5 seconds")
+			select {
+			case <-ctx.Done():
+				return c.handleContextCancel(ctx)
+			case <-time.After(time.Second * 5):
```
jirwin, line 100: Unsure about the amount of time here, I think it could actually probably be a lot higher. We might want it to be configurable by the task manager? `Next()` could return the next time to check again instead of it being hardcoded.

pkg/connectorrunner/runner.go
```diff
@@ -1,86 +1,158 @@
...
+			if err != nil {
+				return err
+			}
+
+		default:
+			l.Debug("unknown task type, going to sleep for 10 seconds")
+			select {
+			case <-ctx.Done():
+				return c.handleContextCancel(ctx)
+			case <-time.After(time.Second * 10):
```
jirwin, line 146: Another randomly picked duration to wait to avoid spinning cores. I assume we want to be resilient to unknown things, so I don't exit here.

pkg/sync/syncer.go
```diff
@@ -761,15 +772,15 @@ func WithTransitionHandler(f func(s Action)) SyncOpt {
 }
 
 // NewSyncer returns a new syncer object.
-func NewSyncer(store connectorstore.Writer, c types.ClientWrapper, opts ...SyncOpt) Syncer {
+func NewSyncer(ctx context.Context, c types.ConnectorClient, dbPath string, opts ...SyncOpt) (Syncer, error) {
```
jirwin, line 775: Need to think more about this prototype and make sure it makes sense for SaaS use cases. I don't think this is actually what we want today.

pkg/connectorrunner/runner.go
pquerna: yes, this should be a lot higher

pquerna: well its complicated. maybe not.  something to think about for sure.

PR #46, opened 2023-03-23
Checkpoint more often while syncing resources and their children types

jirwin:
Before this change, we would sync all resources(and their children types) of a given resource type in a single sync run without checkpointing. This means that if the sync failed in the middle of syncing some resource type, the sync would have to begin again at the first page of resources for that type.

After this change, the syncer will checkpoint after each page of resources processed so that we can resume without having to repeat a large amount of work.

This is accomplished by refactoring how resources are synced. Now we will push individual state actions onto the sync stack as we discover pages and child resource types.

Previously we kept a slice of the resources we've seen so that we wouldn't find ourselves in an infinite loop of syncing resources. Because we are calling `SyncResources` many more times now, we don't have state for which resources we've visited across invocations. To get around this before processing a resource, we actually check the store to see if we've synced the resource already. If we have, we skip it and move onto the next one.

### DIFF
--- a/pkg/sync/state.go
+++ b/pkg/sync/state.go
@@ -16,6 +16,8 @@ type State interface {
 	NextPage(ctx context.Context, pageToken string) error
 	ResourceTypeID(ctx context.Context) string
 	ResourceID(ctx context.Context) string
+	ParentResourceID(ctx context.Context) string
+	ParentResourceTypeID(ctx context.Context) string
 	PageToken(ctx context.Context) string
 	Current() *Action
 	Marshal() (string, error)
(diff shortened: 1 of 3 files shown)


pkg/sync/state.go
```diff
@@ -95,10 +97,12 @@ const (
 
 // Action stores the current operation, page token, and optional fields for which resource is being worked with.
 type Action struct {
-	Op             ActionOp `json:"operation"`
-	PageToken      string   `json:"page_token"`
-	ResourceTypeID string   `json:"resource_type_id"`
-	ResourceID     string   `json:"resource_id"`
+	Op                   ActionOp `json:"operation,omitempty"`
```
jirwin, line 100: Added `omitempty` here to reduce the size of the serialized state stored.

pkg/sync/syncer.go
```diff
@@ -290,84 +284,70 @@ func (s *syncer) SyncResources(ctx context.Context) error {
 		return nil
 	}
 
-	visited := make(map[subResource]struct{})
-	subResources := []subResource{{resourceTypeId: s.state.Current().ResourceTypeID}}
+	return s.syncResources(ctx)
```
morgabra, line 287: I think you need to return on l277 after you finish fetching all the resource types? It lets you checkpoint and it's possible there aren't actually any resource types so we wouldn't push any resource actions on the stack.

pkg/sync/syncer.go
```diff
@@ -290,84 +284,70 @@ func (s *syncer) SyncResources(ctx context.Context) error {
...
-
-	pageToken := ""
-	for {
-		req := &v2.ResourcesServiceListResourcesRequest{
-			ResourceTypeId:   resourceTypeID,
-			ParentResourceId: parentResourceID,
-			PageToken:        pageToken,
+	for _, r := range resp.List {
+		// Check if we've already synced this resource, skip it if we have
+		_, err = s.store.GetResource(ctx, &reader_v2.ResourceTypesReaderServiceGetResourceRequest{
```
morgabra, line 324: Nice, I think this works.

pkg/sync/syncer.go
morgabra: Ah jeeze, I don't think I had this expanded all the way in the diff. I see that we do return. Disregard.

PR #53, opened 2023-04-05
Beginnings of windows support

jirwin:
The primary compatibility issues with windows was how the parent process prepares a listener for the child process.

In linux, the parent process creates and starts the listener, and then passes the fd of the listener onto the child process. The method of creating a new listener from a file isn't supported in windows golang.

On windows the flow looks more like:
1. parent process creates listener `tcp localhost:0`
2. parent process records the port the listener is started on. Because we picked `0`, the kernel is responsible for giving us some random high port.
3. parent process closes the listener
4. parent process starts up the child process, passing the port from the listener it created as part of the server config.
5. child process starts up its own listener on the specified port. The idea is that the kernel keeps track of what ports it hands out, and shouldn't automatically assign anything to that port.


internal/connector/connector_server.go
```diff
@@ -0,0 +1,69 @@
+//go:build !windows
```
pquerna, line 1: filename: maybe have _unix.go ?

PR #56, opened 2023-04-11
Service mode

jirwin:
Changes included in this PR:
- Service/Daemon mode
- ConductorOne API integration
- Windows and Windows service support

## Service Mode
The primary goal of this change is to refactor the CLI harness to take ownership of the process and introduce the idea of a 'daemon' mode for connectors. The SDK takes over the main run loop of the connector, working off of a serial work queue. Today this includes a single task type for Syncs, but you can imagine adding `Grant` and `Revoke` tasks as well. 

This is the first step on the path to allowing connectors to run in "service mode" where they will communicate with an upstream API in order to get work to do and submit the results of each task.

This is a breaking change for connectors consuming the SDK. They will be required to make updates to begin using this version of the SDK. 

### main loop changes
The primary change here is that the SDK cli package takes control of running the connector implementation instead of the connector implementation itself. Before this change, when using the `NewCmd()`, SDK users would pass a run function defined by the connector. This function looked something like this:
```
// run is where the process of syncing with the connector is implemented.
func run(ctx context.Context, cfg *config) error {
	l := ctxzap.Extract(ctx)

	c, err := getConnector(ctx, cfg)
	if err != nil {
		return err
	}

	r, err := sdk.NewConnectorRunner(ctx, c, cfg.C1zPath)
	if err != nil {
		l.Error("error creating connector runner", zap.Error(err))
		return err
	}
	defer r.Close()

	err = r.Run(ctx)
	if err != nil {
		l.Error("error running connector", zap.Error(err))
		return err
	}

	return nil
}
```

With this change, the SDK is in charge of the run command for the 'main loop' of the connector process, meaning that implementors only need to provide a factory method.

### on demand mode
in order to keep the existing functionality(running without the new `-d` flag), this also implements 'on-demand' mode. If we don't have the daemonize flag set, we will still bring up the connector as if we were daemonized, but we manually push a sync task into the work queue, and then explicitly quit after running the single task. 

### connector lifecycle
Previous to this change, when the connector was run it would immediately spin up the connector server as a child process, run a sync, and then shut everything down. Now the connector server is only spun up to perform a task, and then is shutdown. A new process will be spun up to process each task.

## ConductorOne API Integration
In order to have the work the connector performs managed by the API we need to integrate with the Baton API from the C1 platform. The typical flow of this API interaction goes as follows:
1. Authenticate connection
2. Process and submit an initial HelloTask
3. Check to see if there are new tasks for us to process
4. If there is, pass it to the task manager.
5. Once we finish processing the task, we call the FinishTask endpoint in the API to indicate that we've finished. We may report errors here as well.
6. We then wait some amount of time(determined by the API) before checking for a new task

### Authentication
In order to authenticate with the API, users provide a `client-id` and `client-secret`. These are obtained by creating a new Baton integration within the C1 platform.

## Windows Support
- Adds support for spinning up child processes and listeners in windows
-  Adds support for setup and configuring a windows process
-  Support for logging support for windows services.


internal/connector/connector.go
```diff
@@ -230,20 +225,25 @@ func (cw *wrapper) runServer(ctx context.Context, serverCred *tlsV1.Credential)
 
 // C returns a ConnectorClient that the caller can use to interact with a locally running connector.
 func (cw *wrapper) C(ctx context.Context) (types.ConnectorClient, error) {
+	// Check to see if we have a client already
+	cw.mtx.RLock()
```
btipling, line 229: `A` - First call to `C`
`B` - Second call to `C`
`C` - Third call to `C`

`A` happens, gets `RLock`
`B` happens, gets `RLock`

`A` & `B` see no client.

`A` gets `Lock` first.
`B` blocked on acquiring `Lock` by `A`'s lock on `Lock`

`C` happens, blocked on acquiring `RLock` by `A`'s lock on `Lock`

`A` runs through and creates client.
`B` unblocked, acquires `Lock` sees client created and early returns.
`C` unblocked from acquiring `RLock`, sees client and early returns.

How is this different than if you just got a `Lock` on line 229 instead of an `RLock`?

`A` acquires `Lock`
`B` blocked on acquiring `Lock`
`C` blocked on acquiring `Lock`

`A`sees no client and continues on and creates a client.
`B` unblocked,  acquires `Lock`, early returns on seeing client
`C` unblocked, acquires `Lock`, early returns on seeing client.

No `Rlock` needed?

Just trying to understand, thank you!

btipling: Oh I think understand, it's so that multiple calls can get a client once it's been created without one locking the other, but getting a client is blocked while one is being created.

pkg/tasks/c1api/service_client.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/conductorone/baton-sdk/pkg/ugrpc"
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	// API allows for up to 4MB chunks, but that is also the max gRPC message size. We use 3MB to be safe.
+	fileChunkSize = 3 * 1024 * 1024 // 3MB
```
pquerna, line 25: we changed this to 1mb i think on the c1 side?

pkg/tasks/c1api/service_client.go
```diff
@@ -0,0 +1,253 @@
...
+	return nil
+}
+
+func newServiceClient(ctx context.Context, clientID string, clientSecret string) (BatonServiceClient, error) {
+	credProvider, clientName, tokenHost, err := ugrpc.NewC1CredentialProvider(ctx, clientID, clientSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	if envHost, ok := os.LookupEnv("BATON_C1_API_HOST"); ok {
```
btipling, line 230: Should Baton reference C1? Are we releasing a Baton server since it's open source?

pkg/ugrpc/c1_credential_provider.go
```diff
@@ -0,0 +1,245 @@
...
+
+	return clientName, items[0], nil
+}
+
+func parseSecret(input []byte) (*jose.JSONWebKey, error) {
+	items := bytes.SplitN(input, []byte(":"), 4)
+	if len(items) != 4 {
+		return nil, ErrInvalidClientSecret
+	}
+
```
pquerna, line 72: missing check if items[2] == "v1"?

pkg/tasks/c1api/service_client.go
btipling, line 230: Should Baton reference C1? Are we releasing a Baton server since it's open source?

btipling (commented):
pretty cool

pkg/ugrpc/c1_credential_provider.go
```diff
@@ -0,0 +1,245 @@
...
+		Host:   tokenHost,
+		Path:   "auth/v1/token",
+	}
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodPost, tokenUrl.String(), strings.NewReader(body.Encode()))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	resp, err := http.DefaultClient.Do(req)
```
pquerna, line 161: could this use our uhttp client?  defaultclient has... tradeoffs? no user-agent?

pquerna: yeah actually the struct has an http.Client, should use that

pkg/tasks/c1api/service_client.go
jirwin: This is lowered to 512KB now.

pkg/ugrpc/c1_credential_provider.go
jirwin: This is fixed in 8700b1a260f38ac4fd892f0e608c3c4289919ecc

pkg/tasks/c1api/service_client.go
jirwin: No plan on releasing a server right now. This is specifically referenced within the implementation of the `c1api` though, so I think it is ok to reference it here.

pkg/ugrpc/c1_credential_provider.go
jirwin: Good call. We were already creating a client to use for other calls, just forgot to update this one. Fixed in cfad197.

internal/connector/connector.go
```diff
@@ -305,6 +304,9 @@ func (cw *wrapper) C(ctx context.Context) (types.ConnectorClient, error) {
 
 // Close shuts down the grpc server and closes the connection.
 func (cw *wrapper) Close() error {
+	cw.mtx.Lock()
```
pquerna, line 307: tiny: its weird how we have an C() which returns an instance, and caches it.

but then Close() unconditionally closes it.

it feels like maybe it should be a reference counting by default, and a HardClose()/Shutdown() if we're actually serious.

just a thought.

pquerna (approved):

PR #57, opened 2023-04-12
[service-mode] add basic log rotation

pquerna:


pkg/logging/logging.go
```diff
@@ -36,9 +39,48 @@ func WithLogFormat(format string) Option {
...
 func WithOutputPaths(paths []string) Option {
 	return func(c *zap.Config) {
-		c.OutputPaths = paths
+		p := make([]string, 0, len(paths))
+		for _, path := range paths {
+			switch path {
+			case "stdout", "stderr":
+				p = append(p, path)
+			default:
+				u := &url.URL{Scheme: rotatorrScheme, Path: path}
```
jirwin, line 52: does rotatorr deal with this path in a portable way?

pquerna: yes, i think they do.  all the CI is running on windows too:
https://github.com/golift/rotatorr/blob/main/.github/workflows/codetests.yml

PR #63, opened 2023-05-18
Etag support to enable re-using grant data from previous sync generations

jirwin:
# Why
For some connectors, it can be very expensive to calculate every grant for a given resource type. Sometimes upstream APIs are kind enough to return some information to know whether or not the relevant information has changed since we last tried to fetch it. If we can be confident that the data hasn't changed, and we have data available to us about the previous sync, we can use that data directly instead.

# What
This change introduces 3 new annotations for coordinating ETagging:
- `ETagMetadata`
- `ETag`
- `ETagMatch`

`ETagMetadata` is used to store information related to the upstream ETag mechanism. This might be a hash of the object, an actual ETag header, or even a timestamp of when the last update happened. When listing resources, the implementation can store an `ETagMetadata` annotation on the resource. This allows the data to be associated with the resource for the lifetime of the object, including when being referenced by future sync generations.  It contains a map that you can store arbitrary values in.

`ETag` is an annotation that contains the final computed ETag that the next sync generation can use to determine if work should happen. It stores an opaque value that is computed by the connector developer how they see fit.

`ETagMatch` is an annotation that can be returned as an annotation on a response. If the syncer detects this annotation, it will know to use the data from the previous sync generation instead of only what was included in the API response. 

# How
1. During `ListResources()`, `ETagMetadata` can be set and persisted on each resource individually. This annotation will be alongside the resource anytime it is passed to the connector implementation from that point forward.
2. Before the SDK invokes `ListGrants()` for a resource, it checks to see if that resource existed in the previous generation of data. If it does *and* it has a stored `ETag` annotation from the previous generation, the `ETag` is fetched and included as an annotation on the resource provided to `ListGrants()`.
3. During `ListGrants()`, the original resource with its fresh `ETagMetadata` annotation is provided. If it is available, the `ETag` from the previous sync is also provided on the resource.
4. The developer is able to use the `ETagMetadata` and the previous computed `ETag` to determine if new work needs to be done, or if we can reuse the data from the previous sync generation. 
5. If it is determined that we can use the previous generation, the developer can return an empty list with the `ETagMatch` annotation. The SDK looks for the `ETagMatch` message, and if it sees it *and* it is valid(e.g. there is a previous generation of data), the data from the previous sync will be used. Developers can still return new grants and have those merged with the data from the previous run if required.
6. If the work needs to be done(no `ETagMatch` message returned), the developer can include a new `ETag` message on the last page of results, and this will be persisted on the resource. This is the `ETag` value that will be provided to the next sync generation in order to determine if we can skip doing work.


pkg/dotc1z/sql_helpers.go
```diff
@@ -74,6 +78,33 @@ func (c *C1File) listConnectorObjects(ctx context.Context, tableName string, req
...
+	// If the request has a sync id annotation, use that
+	case hasSyncIdAnno && syncDetails.GetId() != "":
+		reqSyncID = syncDetails.GetId()
+
+	// We are currently syncing, so use the current sync id
+	case c.currentSyncID != "":
+		reqSyncID = c.currentSyncID
+
+	// We are viewing a sync, so use the view sync id
+	case c.viewSyncID != "":
```
jirwin, line 100: I think it might be worth dropping `viewSyncID` entirely now that we can specify which sync we care about as part of the request.

pkg/sync/syncer.go
```diff
@@ -734,7 +739,43 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+			}
+		}
+		var lastSyncResourceReqAnnos annotations.Annotations
+		lastSyncResourceReqAnnos.Update(&c1zpb.SyncDetails{Id: latestSyncID})
+		if latestSyncID != "" {
+			prevResource, err := s.store.GetResource(ctx, &reader_v2.ResourceTypesReaderServiceGetResourceRequest{
+				ResourceId:  resourceID,
+				Annotations: lastSyncResourceReqAnnos,
+			})
+			// FIXME(jirwin): Don't fail if this resource didn't exist in the previous sync
```
jirwin, line 762: Need to fix this before merging

pkg/sync/syncer.go
```diff
@@ -745,22 +814,91 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+	respAnnos := annotations.Annotations(resp.GetAnnotations())
+	etagMatch := respAnnos.Contains(&v2.ETagMatch{})
+
+	var grantsRet []*v2.Grant
+
+	// We have a previous etag, and the connector has indicated an etag match
+	if etagMatch {
+		if prevEtag == nil {
+			return errors.New("connector returned an etag match - but there is no previous sync generation to use")
+		}
```
btipling, line 829: Should this err if next page token is also set? should it set next page token to empty string? What does it mean if there's an etag match and a next page token? Code below starts with empty page string and iterates through all previous request pages. I see the okta connector code returns an empty page string when etagmatch is being returned.

btipling (commented):
this mechanism makes sense, one question

pkg/sync/syncer.go
```diff
@@ -734,7 +790,20 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+
+	resourceAnnos := annotations.Annotations(resource.GetAnnotations())
 	pageToken := s.state.PageToken(ctx)
+	// Empty page token, so we're on the first page of grants for this resource
+	if pageToken == "" {
+		prevSyncID, prevEtag, err = s.fetchResourceForPreviousSync(ctx, resourceID)
+		if err != nil {
+			return err
+		}
+		resourceAnnos.Update(prevEtag)
```
btipling, line 804: Oh so any grants call updates the entire resources etag?

pkg/sync/syncer.go
```diff
@@ -745,22 +814,91 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+		if err != nil {
+			return err
+		}
+		if ok {
+			updatedETag = newETag
+		}
+	}
+
+	if updatedETag != nil {
+		resourceAnnos.Update(updatedETag)
```
btipling, line 896: Oh so any grants call updates the resources etag, so they could be overwritten by a next page.

jirwin: We only reach this portion of the code if we're on the last page. If you look at line 871, if there is a next page token we exit early.

pkg/sync/syncer.go
jirwin: The code below is making a request against the store, and not the connector, so it is paginating through all of the grants from the previous generation.

I could go either way on if this is an error or not. As implemented it should still work if additional pages are provided, but we won't do the work to see if there is a previous generation for the resource on additional pages.

btipling (approved):

PR #66, opened 2023-06-26
Add new grant metadata annotation

jirwin:
Allows devs to set metadata on a grant when it is synced. This metadata is then included when the grant is looked up. This is specifically useful when revoking a grant.


pkg/types/grant/grant.go
```diff
@@ -55,7 +73,10 @@ func NewGrant(resource *v2.Resource, entitlementName string, principal GrantPrin
 	grant.Id = fmt.Sprintf("%s:%s:%s", entitlement.Id, resourceID.ResourceType, resourceID.Resource)
 
 	for _, grantOption := range grantOptions {
-		grantOption(grant)
+		err := grantOption(grant)
+		if err != nil {
+			panic(err)
```
jirwin, line 78: Unfortunate panic here. We'll need to update this function prototype to return an error, which is an API breaking change. I'll make a note to come back to this.

mstanbCO (approved):

PR #73, opened 2023-09-13
Grant expansion support

morgabra:
Adds the ability to annotate a grant denoting that it is expandable. The annotation takes a list of entitlements that need to be expanded for the principal. 

Example:
```
{
	Id:   "2IC0WmAPkihbFdZhEPsch5N5WNO",
	Name: "Engineers",
	Admins: []string{
		"2IC0Wo34fcTerFEgWmyffXmfrW8", // Carol
	},
	Members: []string{
		"2IC0Wn5oRQqVVn3COFl1O1zSzV6",       // Alice
		"group:2VJleiE6zMPoTHbKDjLC1zJkUir", // Expanded Engineers Group
	},
},
{
	Id:   "2VJleiE6zMPoTHbKDjLC1zJkUir",
	Name: "Expanded Engineers",
	Admins: []string{
		"2IC0Wn7DaxV1xqDpdg7jJRiPtCp", // Dan
	},
	Members: []string{
		"2IC0WoNfqUPT7mgO4FOaViIxBrR",       // Bob
	},
},
```

In the connector we would need to set the `GrantExpandable` annotation on the member ship entitlement grant between `Engineers` and `Expanded Engineers`. On the entitlement grant we would need to specify the entitlementIds we wish to expand from the `Expanded Engineers` group into the `Engineers` entitlement. This would look something like below.

```
annos.Append(&v2.GrantExpandable{
   EntitlementIds: []string{member.Id, admin.Id},
})
```
We then do the work to expand those entitlements from the principal group into the destination group's entitlement. 

TODO:
- [ ] Fold cycles into a single node, and don't hard fail the sync.


pkg/cli/cli.go
```diff
@@ -262,6 +266,9 @@ func NewCmd[T any, PtrT *T](
 	cmd.PersistentFlags().String("log-level", defaultLogLevel, "The log level: debug, info, warn, error ($BATON_LOG_LEVEL)")
 	cmd.PersistentFlags().String("log-format", defaultLogFormat, "The output format for logs: json, console ($BATON_LOG_FORMAT)")
 
+	// Flag for syncing grant expansion
+	cmd.PersistentFlags().BoolP("expand-grants", "e", false, "Enables expanding grants while syncing ($BATON_EXPAND_GRANTS)")
```
jirwin, line 270: What is the long term thought about this flag? Do we intend to keep it? Any reason it shouldn't be part of the default behavior?

mstanbCO: I think the main thing for now is that we can ship it without any behavior being changed. We do some work that we don't need to do unless we are expanding grants. Also we could possibly put on the capabilities object whether or not certain resources are expandable from a top level. We could then skip this step if the connector doesn't support any expandable resources. Main thing is it makes the PR safer in my eyes.

pkg/sync/syncer.go
```diff
@@ -928,6 +1038,191 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+
+		// If we have no grants for the principal in the descendant entitlement, make one.
+		directGrant := true
+		if len(descendantGrants) == 0 {
+			directGrant = false
+			// TODO(morgabra): This is kinda gnarly, grant ID won't have any special meaning.
+			// FIXME(morgabra): We should probably conflict check with grant id?
+			descendantGrant := grant.NewGrant(descendantEntitlement.GetEntitlement().GetResource(), descendantEntitlement.GetEntitlement().DisplayName, sourceGrant.GetPrincipal().GetId())
+			descendantGrants = append(descendantGrants, descendantGrant)
+			l.Info(
```
jirwin, line 1118: Probably want this at debug

pkg/sync/syncer.go
```diff
@@ -928,6 +1038,191 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
 	return nil
 }
 
+func (s *syncer) runGrantExpandActions(ctx context.Context) (bool, error) {
+	l := ctxzap.Extract(ctx)
+
+	graph := s.state.EntitlementGraph(ctx)
+	l = l.With(zap.Int("depth", graph.Depth))
+	l.Info("runGrantExpandActions: start", zap.Any("graph", graph))
```
jirwin, line 1046: Log at debug

pkg/sync/syncer.go
```diff
@@ -928,6 +1038,191 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+			}
+
+			if directGrant && len(sources.Sources) == 0 {
+				// If we are already granted this entitlement, make sure to add ourselves as a source.
+				sourcesMap[descendantGrant.GetEntitlement().GetId()] = &v2.GrantSources_GrantSource{}
+			}
+			// Include the source grant as a source.
+			sourcesMap[sourceGrant.GetEntitlement().GetId()] = &v2.GrantSources_GrantSource{}
+
+			l.Info(
```
jirwin, line 1144: Log at debug

jirwin:
Noting that we'll need to update `baton` to have the new proto for json output to function. We should also probably deal with that more gracefully while we fix that. I get this error when trying to output grants in json with baton:
`proto: google.protobuf.Any: unable to resolve "type.googleapis.com/c1.connector.v2.GrantExpandable": not found`

pkg/sync/syncer.go
```diff
@@ -928,6 +1038,191 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+			}
+		}
+
+		// If we have no grants for the principal in the descendant entitlement, make one.
+		directGrant := true
+		if len(descendantGrants) == 0 {
+			directGrant = false
+			// TODO(morgabra): This is kinda gnarly, grant ID won't have any special meaning.
+			// FIXME(morgabra): We should probably conflict check with grant id?
+			descendantGrant := grant.NewGrant(descendantEntitlement.GetEntitlement().GetResource(), descendantEntitlement.GetEntitlement().DisplayName, sourceGrant.GetPrincipal().GetId())
```
jirwin, line 1116: Is `DisplayName` here a typo?

jirwin: I understand what is going on now. I think in this case, I'm not sure we want to use the `NewGrant` helper here, as it essentially requires that the connector dev is using the helpers to create their resources.

mstanbCO (approved):

PR #74, opened 2023-09-26
Create resources for grant principals that don't exist

mstanbCO:
This PR adds support for creating resources when syncing grants where we find a principal that doesn't already exist from syncing resources. This PR also adds support for marking resourceTypes as `sourcedExternally`. If a resourceType is `sourcedExteranlly=true` then we know we need to look elsewhere for the resource during uplift.


pkg/sync/syncer.go
```diff
@@ -1032,3 +1040,34 @@ func NewSyncer(ctx context.Context, c types.ConnectorClient, opts ...SyncOpt) (S
...
+		ResourceId: &v2.ResourceId{
+			ResourceType: principal.Id.ResourceType,
+			Resource:     principal.Id.Resource,
+		},
+	})
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			// If the principal does not have a display name, we should set it to the resource ID
+			if principal.DisplayName == "" {
+				principal.DisplayName = principal.Id.Resource
```
jirwin, line 1059: I wonder if we should include the resource type here as well

pkg/sync/syncer.go
```diff
@@ -1032,3 +1040,34 @@ func NewSyncer(ctx context.Context, c types.ConnectorClient, opts ...SyncOpt) (S
...
+func (s *syncer) ensureResourceExistence(ctx context.Context, principal *v2.Resource) error {
+	l := ctxzap.Extract(ctx)
+	// Check if the principal of the grant already exists as a resource. If it does not, we should jit the resource.
+	_, err := s.store.GetResource(ctx, &reader_v2.ResourcesReaderServiceGetResourceRequest{
+		ResourceId: &v2.ResourceId{
+			ResourceType: principal.Id.ResourceType,
+			Resource:     principal.Id.Resource,
+		},
+	})
+	if err != nil {
```
jirwin, line 1055: I think I'd write this as `err != nil && !errors.Is(err, sql.ErrNoRows)` to save an indent level

pkg/sync/syncer.go
```diff
@@ -1032,3 +1040,34 @@ func NewSyncer(ctx context.Context, c types.ConnectorClient, opts ...SyncOpt) (S
 
 	return s, nil
 }
+
+// This method checks if the principal resource already exists. If it does not, we should put the resource.
+// The return value indicates if the resource was put or not by the boolean value.
+func (s *syncer) ensureResourceExistence(ctx context.Context, principal *v2.Resource) error {
```
jirwin, line 1046: Probably better to call this `ensurePrincipalExists` or something like that, since we're not using it for all resources(e.g. we're not JITing the resource on the entitlement)

mstanbCO: Fixed in latest commit

pkg/sync/syncer.go
mstanbCO: Fixed in latest commit

jirwin (approved):

PR #76, opened 2023-10-10
Add user login support to SDK

loganintech:


pkg/types/resource/user_trait.go
```diff
@@ -34,6 +34,22 @@ func WithEmail(email string, primary bool) UserTraitOption {
 		return nil
 	}
 }
+func WithUsername(username string, primary bool) UserTraitOption {
```
jirwin, line 37: I'm not sure `Username` is the right name for this field. If it is, this option should probably be renamed to `WithUserUsername()` since it is an option specific to user traits. 

maybe `external_identifier` ?

proto/c1/connector/v2/annotation_trait.proto
```diff
@@ -20,6 +20,18 @@ message UserTrait {
...
+    string username = 1 [
+      (validate.rules).string = {
+        min_bytes: 1,
+        max_bytes: 1024,
+      }
+    ];
+    // Indicates if this is the user's primary username. Only one entry can be marked as primary.
+    bool is_primary = 2;
+  }
+  repeated Username usernames = 6;
```
jirwin, line 33: nit: can this go to the bottom so it is easier for the next person to know the next field number?

jirwin (approved):

PR #77, opened 2023-10-14
Allow the tempdir used for working with c1zs to be configurable.

jirwin:
The default OS tempdir might not be in the place that we'd prefer it to be. Instead of asking users to change the default path systemwide, allow users to configure it while using the SDK.

- Adds a new `c1z-temp-dir` cli flag to any SDK connectors
- Flows that through to c1z managers and the syncer.


pkg/dotc1z/file.go
```diff
@@ -12,7 +13,8 @@ import (
 )
 
 func loadC1z(filePath string) (string, error) {
-	workingDir, err := os.MkdirTemp("", "c1z")
+	tmpDir := path.Dir(filePath)
+	workingDir, err := os.MkdirTemp(tmpDir, "c1z")
```
jirwin, line 17: there isn't a great way to flow this through without breaking some APIs. Most of the usages of this are already in a tempdir using the correct defined path, so we use the same directory and create a new tempdir there to use. This dir is removed on cleanup.

pkg/dotc1z/c1file.go
```diff
@@ -21,10 +21,19 @@ type C1File struct {
...
+
+func WithC1FTmpDir(tempDir string) C1FOption {
+	return func(o *C1File) {
+		o.tempDir = tempDir
+	}
 }
 
 // Returns a C1File instance for the given db filepath.
-func NewC1File(ctx context.Context, dbFilePath string) (*C1File, error) {
+func NewC1File(ctx context.Context, dbFilePath string, opts ...C1FOption) (*C1File, error) {
```
pquerna, line 36: style nit: not sure why this can't just be `New()` instead of ` NewC1File`?

pquerna (approved):

pkg/dotc1z/c1file.go
jirwin: This package also contains `NewC1ZFile()`, so I think it was to disambiguate. We could probably move the details of that function into an option for this function. Will consider that in a separate PR since it'd be API breaking.

PR #79, opened 2023-10-19
Add new method for vacuuming a c1z db.

jirwin:
Vacuum after removing old generations of data.

Without this vacuuming, the c1z database will continue to grow in size even though we are removing old data. Vacuuming allows us to reclaim the space used by previous generations, keeping the decompressed file size as small as expected.

### DIFF
--- a/pkg/dotc1z/sync_runs.go
+++ b/pkg/dotc1z/sync_runs.go
@@ -424,6 +424,13 @@ func (c *C1File) Cleanup(ctx context.Context) error {
 		l.Info("Removed old sync data.", zap.String("sync_date", ret[i].EndedAt.Format(time.RFC3339)), zap.String("sync_id", ret[i].ID))
 	}
 
+	err = c.Vacuum(ctx)
+	if err != nil {
+		return err
+	}
+
+	c.dbUpdated = true
+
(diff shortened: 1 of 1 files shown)


pkg/dotc1z/sync_runs.go
```diff
@@ -424,6 +424,13 @@ func (c *C1File) Cleanup(ctx context.Context) error {
 		l.Info("Removed old sync data.", zap.String("sync_date", ret[i].EndedAt.Format(time.RFC3339)), zap.String("sync_id", ret[i].ID))
 	}
 
+	err = c.Vacuum(ctx)
+	if err != nil {
+		return err
+	}
+
+	c.dbUpdated = true
```
mstanbCO, line 432: Do you need to set this to true here, when it is already set from the vacuum call?

jirwin: In practice no, but i think it is probably good to set it in any method that changes the database. This flag being set to true is what tells the c1z manager that it should write the c1z back to the origin. There isn't really a way to run this method without invoking other things that change the database, but for completeness sake I'm setting it.

mstanbCO (approved):

PR #88, opened 2023-11-29
Add grant as a return value in the grant response

loganintech:


proto/c1/connector/v2/grant.proto
```diff
@@ -98,6 +98,7 @@ message GrantManagerServiceGrantRequest {
 
 message GrantManagerServiceGrantResponse {
   repeated google.protobuf.Any annotations = 1;
+  Grant grant = 2;
```
jirwin, line 101: I think this should be `repeated Grant grant = 2;`. It is possible that multiple grants needed to be made in order to satisfy the requested grant.

loganintech: Good call

jirwin (approved):

PR #89, opened 2023-11-29
Add grants to the connector builder grant response

loganintech:


pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -53,12 +62,25 @@ func NewConnector(ctx context.Context, in interface{}) (types.ConnectorServer, e
...
+				if _, ok := rb.(ResourceProvisionerV2); ok {
+					return nil, fmt.Errorf("error: resource type %s implements both ResourceProvisioner and ResourceProvisionerV2", rType.Id)
+				}
 				if _, ok := ret.resourceProvisioners[rType.Id]; ok {
 					return nil, fmt.Errorf("error: duplicate resource type found %s", rType.Id)
 				}
 				ret.resourceProvisioners[rType.Id] = provisioner
 			}
+			if provisioner, ok := rb.(ResourceProvisionerV2); ok {
+				if _, ok := rb.(ResourceProvisioner); ok {
```
ggreer, line 76: This conditional is never true and can be removed, right? We'd return an error on line 68 if both are true.

ggreer (approved):
I think I see some dead code but it's definitely better than before.

pkg/connectorbuilder/connectorbuilder.go
loganintech: No, this can be true if they implement both kinds of grant, we want to explicitly forbid this, which is why the `if ok` returns an error

ggreer: Right, but line 67 checks that, just in a different order.

loganintech: Oh, I see what you mean

PR #91, opened 2024-01-03
(Work in progress) Start of new stuff for provisioning accounts with randomly generated passwords.

ggreer:


proto/c1/connector/v2/resource.proto
```diff
@@ -88,6 +89,123 @@ service ResourcesService {
...
+  repeated EncryptionConfig encryption_configs = 3;
+}
+
+message RotateCredentialResponse {
+  repeated EncryptedData encrypted_data = 1;
+  ResourceId resource_id = 2;
+  repeated google.protobuf.Any annotations = 3;
+}
+
+message AccountInfo {
```
jirwin, line 134: How is this different from a user trait?

proto/c1/connector/v2/resource.proto
```diff
@@ -88,6 +89,123 @@ service ResourcesService {
...
+
+message DeleteResourceRequest {
+  ResourceId resource_id = 1;
+}
+
+message DeleteResourceResponse {
+  repeated google.protobuf.Any annotations = 1;
+}
+
+message RotateCredentialRequest {
```
jirwin, line 122: Are credentials always tied to a single resource? Which resource do I include here?

jirwin: I'm curious if credentials should be able to be tied to grants?

proto/c1/connector/v2/resource.proto
```diff
@@ -88,6 +89,123 @@ service ResourcesService {
...
+  string name = 3;
+  string description = 4;
+  string schema = 5; // optional
+  bytes encrypted_bytes = 6; // This should decrypt to json so we can store username & password separately
+}
+
+message EncryptionConfig {
+  Resource prinicpal = 1;
+
+  message PublicKeyConfig {
```
jirwin, line 198: Should this be 'typed'? Or at least include a field for the algorithm being used?

pkg/cli/cli.go
```diff
@@ -290,7 +320,18 @@ func NewCmd[T any, PtrT *T](
 	cmd.PersistentFlags().String("grant-principal-type", "", "The resource type of the principal to grant the entitlement to ($BATON_GRANT_PRINCIPAL_TYPE)")
 	cmd.MarkFlagsRequiredTogether("grant-entitlement", "grant-principal", "grant-principal-type")
 	cmd.PersistentFlags().String("revoke-grant", "", "The grant to revoke ($BATON_REVOKE_GRANT)")
-	cmd.MarkFlagsMutuallyExclusive("grant-entitlement", "revoke-grant")
+
```
jirwin, line 323: I wonder if we should consider moving these provisioning things to a sub command

proto/c1/connector/v2/resource.proto
```diff
@@ -88,6 +89,123 @@ service ResourcesService {
...
+  string provider = 1;
+  string key_id = 2;
+  string name = 3;
+  string description = 4;
+  string schema = 5; // optional
+  bytes encrypted_bytes = 6; // This should decrypt to json so we can store username & password separately
+}
+
+message EncryptionConfig {
+  Resource prinicpal = 1;
```
jirwin, line 196: spelling err

proto/c1/connector/v2/resource.proto
```diff
@@ -88,6 +89,123 @@ service ResourcesService {
...
+
+  message PublicKeyConfig {
+    string provider = 1;
```
jirwin, line 199: Move `provider` and `key_id` up to `EncryptionConfig`

proto/c1/connector/v2/resource.proto
jirwin: Rename this `JWKPublicKeyConfig`

proto/c1/connector/v2/resource.proto
ggreer: Yes, credentials are tied to a single resource like a user. Technically you can include any resource and it's up to the connector to handle it or return an error. Postgres is kind of weird in that users are actually roles, and any role can be given credentials (which basically makes it a user). So you can technically send a rotatecredentials request with a role id, and postgres will `ALTER ROLE rolename WITH PASSWORD ...`. This also lets us do things like rotate API keys on some non-user resource.

After delving into the ticket processing code in c1, I think it's simpler to tie credentials to resources (accounts) instead of grants. A grant can expire or be revoked, but the credential for the relevant account could still be valid. Also credentials on resources is how most applications model them. If a user loses their password they can always do a rotate credential request.

pkg/cli/cli.go
ggreer: I agree, but that would break backwards compatibility and I don't want to do that in this pull request. Ideally I think the command line params should look something like this:

```
baton-connector --provision-action=grant --entitlement=entitlement_id --resource=resource_id --resource-type=resource_type
baton-connector --provision-action=revoke --grant=grant_id
baton-connector --provision-action=delete --resource-id=resource_id --resource-type=resource_type
baton-connector --provision-action=create-account --login=login --email=email
baton-connector --provision-action=rotate-credentials --resource-id=resource_id --resource-type=resource_type
```

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -274,11 +324,125 @@ func (b *builderImpl) Revoke(ctx context.Context, request *v2.GrantManagerServic
...
+	switch r := result.(type) {
+	case *v2.CreateAccountResponse_SuccessResult:
+		rv.Result = &v2.CreateAccountResponse_Success{Success: r}
+	case *v2.CreateAccountResponse_ActionRequiredResult:
+		rv.Result = &v2.CreateAccountResponse_ActionRequired{ActionRequired: r}
+	default:
+		return nil, status.Error(codes.Unimplemented, fmt.Sprintf("unknown result type: %T", result))
+	}
+
+	spew.Dump(rv)
```
jirwin, line 446: nit: leftover spew

pkg/provisioner/provisioner.go
```diff
@@ -156,6 +206,94 @@ func (p *Provisioner) revoke(ctx context.Context) error {
...
+			Login:  p.createAccountLogin,
+		},
+		CredentialOptions: opts,
+		EncryptionConfigs: config,
+	})
+	if err != nil {
+		return err
+	}
+
+	jwe, err := jose.ParseEncrypted(string(result.EncryptedData[0].EncryptedBytes))
```
jirwin, line 236: Do we need to len check `result.EncryptedData`?

pkg/provisioner/provisioner.go
```diff
@@ -156,6 +206,94 @@ func (p *Provisioner) revoke(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	plaintext, err := jwe.Decrypt(privKey)
+	if err != nil {
+		return err
+	}
+	// TODO FIXME: do better
+	l.Info("account created", zap.String("login", p.createAccountLogin), zap.String("email", p.createAccountEmail), zap.String("password", string(plaintext)))
```
jirwin, line 245: Probably want to adjust/drop this log

pkg/provisioner/provisioner.go
```diff
@@ -156,6 +206,94 @@ func (p *Provisioner) revoke(ctx context.Context) error {
...
+	jwe, err := jose.ParseEncrypted(string(result.EncryptedData[0].EncryptedBytes))
+	if err != nil {
+		return err
+	}
+	plaintext, err := jwe.Decrypt(privKey)
+	if err != nil {
+		return err
+	}
+	// TODO FIXME: do better
+	l.Info("credentials rotated", zap.String("resource", p.rotateCredentialsId), zap.String("resource type", p.rotateCredentialsType), zap.String("password", string(plaintext)))
```
jirwin, line 292: another log to change/drop

pkg/tasks/local/accounter.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (m *localAccountManager) Process(ctx context.Context, task *v1.Task, cc types.ConnectorClient) error {
+	accountManager := provisioner.NewCreateAccountManager(cc, m.dbPath, m.login, m.email)
+
+	err := accountManager.Run(ctx)
+	if err != nil {
+		return err
+	}
+
+	err = accountManager.Close(ctx)
```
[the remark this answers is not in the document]
ggreer: Probably an oversight. We don't defer in syncer, granter, or revoker.

PR #101, opened 2024-02-06
Add helpers for auth credentials and http client creation

martinalbert:
Adds new helpers for preparing and applying auth credentials with help of `oauth2` and `http` standard library.

Here are two example uses:
- https://github.com/ConductorOne/baton-bitbucket/pull/14
- https://github.com/ConductorOne/baton-googletagmanager/pull/2


pkg/helpers/helpers.go
vlnevyhosteny, line 1: How about creating a new package `httpclient` and putting those changes there?

I would like to proceed with this PR and add some functions for the http client.

ggreer: Yeah, or it could be moved to another file like pkg/helpers/http.go

ggreer (approved):
Looks good to me. It'd be nice to move these functions into their own file or maybe even their own package, but I'll leave that decision up to you.

PR #103, opened 2024-02-16
Sqlite: Disable sync().

ggreer:
See https://www.sqlite.org/pragma.html#pragma_synchronous for what this changes. It speeds up temporal-sync significantly.

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -13,6 +13,11 @@ import (
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
 )
 
+type pragma struct {
+	name  string
+	value string
+}
+
 type C1File struct {
 	rawDb          *sql.DB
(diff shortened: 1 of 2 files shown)


pkg/dotc1z/c1file.go
```diff
@@ -39,6 +39,10 @@ func NewC1File(ctx context.Context, dbFilePath string, opts ...C1FOption) (*C1Fi
 		return nil, err
 	}
 	db := goqu.New("sqlite3", rawDB)
+	_, err = db.Exec("PRAGMA synchronous = OFF")
```
jirwin, line 42: Should this be a `C1FOption` so we can opt-in to it?

jirwin (approved):

pkg/dotc1z/c1file.go
ggreer: I added the option in commit 354a31445213ae855d147ddf55bc704748727ea4

PR #105, opened 2024-02-21
Export httpClient property in wrapper

martinalbert:


pkg/uhttp/wrapper_test.go
```diff
@@ -137,7 +137,7 @@ func TestWrapper_NewRequest(t *testing.T) {
 			}
 
 			client := BaseHttpClient{
```
ggreer, line 139: It'd be nice if the test used the init function here.

ggreer (approved):

pkg/uhttp/wrapper_test.go
martinalbert: Will do

PR #110, opened 2024-03-04
Add new helpers and options

martinalbert:
This PR adds two new helpers and two new options in wrapper.
- helper for parsing rate limit details
- helper for identifying json content type in response
- option for parsing parsing error responses
- option for parsing rate limit details

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -53,35 +53,30 @@ type ErrorResponse interface {
 }
 
 func WithErrorResponse(resource ErrorResponse) DoOption {
-	return func(resp *http.Response) error {
-		if resp.StatusCode >= 300 {
-			if !helpers.IsJSONContentType(resp.Header.Get("Content-Type")) {
-				r, err := io.ReadAll(resp.Body)
-				if err != nil {
-					return fmt.Errorf("error reading response body: %w", err)
-				}
-
-				return fmt.Errorf("%v", string(r))
-			}
-
-			// Decode the JSON response body into the ErrorResponse
-			if err := json.NewDecoder(resp.Body).Decode(&resource); err != nil {
-				return status.Error(codes.Unknown, "Request failed with unknown error")
-			}
-
-			// Construct a more detailed error message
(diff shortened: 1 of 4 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -37,6 +42,52 @@ func WithJSONResponse(response interface{}) DoOption {
 	}
 }
 
+type ErrorResponse interface {
+	Message() string
+}
+
+func WithErrorResponse(resource ErrorResponse) DoOption {
+	return func(resp *http.Response) error {
+		if resp.StatusCode >= 300 {
```
ggreer, line 51: Should we consider 301/302 redirects to be errors? I think >= 400 is more accurate.

Also this code could early return if the conditional were reversed. eg:

```
if resp.StatusCode < 400 {
  return nil
}
```

ggreer:
Looks good except for one thing. Also I pushed a fix for the failing test.

pkg/uhttp/wrapper.go
```diff
@@ -37,6 +42,52 @@ func WithJSONResponse(response interface{}) DoOption {
 	}
 }
 
+type ErrorResponse interface {
+	Message() string
+}
+
+func WithErrorResponse(resource ErrorResponse) DoOption {
```
vlnevyhosteny, line 49: How this works in parallel with https://github.com/ConductorOne/baton-sdk/pull/110/files#diff-d010464cc4377ff4ba5b4b1c2eb245e880f72188f73564993b55c6546bc56de9R104-R106 ?

martinalbert: It should return an error with details before that part of code. I was thinking of the code you sent as a default smallest way for error handling and use this custom option in case you would want to handle error more.

ggreer:
Now that https://github.com/ConductorOne/baton-sdk/pull/108 has been merged, do you want me to rebase this branch and fix the new option functions to accept a struct instead of an http response?

ggreer:
I rebased and fixed this branch. Hopefully I didn't step on your toes.

ggreer (approved):

PR #111, opened 2024-03-12
feat: expose NewClient options to credentials wrappers

vlnevyhosteny:
I need to apply changes to `Transport` layer so I would like to expose options of `NewClient` to `GetClient`.


martinalbert (approved):
Looks good. Made me think if we even want to have a default logger option or left it upon developer consideration.

PR #117, opened 2024-03-28
Fix types so that BearerAuth implements AuthCredentials interface.

ggreer:


pkg/uhttp/authcredentials.go
```diff
@@ -13,13 +13,13 @@ import (
 )
 
 type AuthCredentials interface {
-	GetClient(ctx context.Context) (*http.Client, error)
+	GetClient(ctx context.Context, options ...Option) (*http.Client, error)
```
pquerna, line 16: should the impl structs have a type assertion on them?
```
var _ AuthCredentials  = (*NoAuth)(nil)
```

mstanbCO (approved):

PR #118, opened 2024-04-02
HTTP helper improvements

ggreer:
- Check content type of JSON and XML responses before trying to parse the body.
- Add HTTP status 429 handling to helpers.ExtractRateLimitData().

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -45,6 +45,9 @@ func NewBaseHttpClient(httpClient *http.Client) *BaseHttpClient {
 
 func WithJSONResponse(response interface{}) DoOption {
 	return func(resp *WrapperResponse) error {
+		if !helpers.IsJSONContentType(resp.Header.Get("Content-Type")) {
+			return fmt.Errorf("unexpected content type for json response: %s", resp.Header.Get("Content-Type"))
+		}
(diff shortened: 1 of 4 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -45,6 +45,9 @@ func NewBaseHttpClient(httpClient *http.Client) *BaseHttpClient {
 
 func WithJSONResponse(response interface{}) DoOption {
 	return func(resp *WrapperResponse) error {
+		if !helpers.IsJSONContentType(resp.Header.Get("Content-Type")) {
```
jirwin, line 48: Wonder if we should use a constant for `Content-type` at this point.

pkg/uhttp/wrapper.go
```diff
@@ -45,6 +45,9 @@ func NewBaseHttpClient(httpClient *http.Client) *BaseHttpClient {
 
 func WithJSONResponse(response interface{}) DoOption {
```
jirwin, line 46: Wonder if we should have a generic `WithResponse()` method that does the same as these, but switches on the content type.

jirwin (approved):
Looks good. Left a note about using a constant for the content-type header, but pretty minor.

ggreer:
I added a WithResponse() and a test for it.

PR #119, opened 2024-04-02
General retry logic in syncing

ggreer:
This changes the http client and the syncer so that individual connectors don't have to handle every recoverable http error. If the error status is something we can retry or recover from, we do so. For example, 429 (too many requests) caused connectors to error out. This PR fixes that.

TODO:
- [ ] Handle "skippable errors", such as a 404 for a specific resource. We need to advance the syncer state in that case.
- [ ] Backoff behavior (right now we always wait exactly one second)
- [x] Use something smarter than time.sleep() to wait.

If this works, we can simplify connector code and avoid an entire class of bugs.

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 
 	c1zpb "github.com/conductorone/baton-sdk/pb/c1/c1z/v1"
@@ -70,6 +72,20 @@ func (s *syncer) handleProgress(ctx context.Context, a *Action, c int) {
 	if s.progressHandler != nil {
 		s.progressHandler(NewProgress(a, uint32(c)))
 	}
+}
+
+func shouldWaitAndRetry(ctx context.Context, err error) bool {
+	if status.Code(err) != codes.Unavailable {
+		return false
+	}
+
+	l := ctxzap.Extract(ctx)
+	l.Error("retrying operation", zap.Error(err))
+
+	// TODO: this should back off based on error counts
+	time.Sleep(1 * time.Second)
+
+	return true
 }
 
 // Sync starts the syncing process. The sync process is driven by the action stack that is part of the state object.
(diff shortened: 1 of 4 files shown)


ggreer:
I think the TODOs should be done in another PR, as this code prevents some crashes that are affecting connectors right now.

pkg/sync/syncer.go
```diff
@@ -72,6 +74,20 @@ func (s *syncer) handleProgress(ctx context.Context, a *Action, c int) {
...
+func shouldWaitAndRetry(ctx context.Context, err error) bool {
+	if status.Code(err) != codes.Unavailable {
+		return false
+	}
+
+	l := ctxzap.Extract(ctx)
+	l.Error("retrying operation", zap.Error(err))
+
+	// TODO: this should back off based on error counts
+	time.Sleep(1 * time.Second)
```
pquerna, line 86: at a minimum this should use a context.Done() / time.After()  select

pkg/uhttp/wrapper.go
```diff
@@ -148,8 +149,21 @@ func (c *BaseHttpClient) Do(req *http.Request, options ...DoOption) (*http.Respo
...
+	case http.StatusUnauthorized:
+		return resp, status.Error(codes.Unauthenticated, resp.Status)
+	case http.StatusForbidden:
+		return resp, status.Error(codes.PermissionDenied, resp.Status)
+	case http.StatusNotImplemented:
+		return resp, status.Error(codes.Unimplemented, resp.Status)
+	}
+
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return resp, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
```
pquerna, line 152: ideally this would... use a baton-uthtp.Error type right? so you can add interface methods to get more info.  eg, the like the AWS Error type in their SDK?

ggreer: If we did that, wouldn't we have to check for baton uhttp errors in syncer.go's shouldWaitAndRetry()? And if we had some non-http error that was retryable, we'd have to check that in the same place. If, alternatively, we made general baton error types, they'd probably end up looking a lot like grpc's statuses, and I figured that grpc's statuses are better than whatever we'd make on our own. They correspond to most http statuses and have docs about what each error means: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

pquerna: The big issue with gprc.Status objects is out of the box you loose the ability to keep more HTTP status around.... so its reallty that some APIs are weird.

pquerna: Alternatively, maybe the way a connector that needs "weird" handling is it could inject a HookError callback, and let it map it in a non-default way

pquerna (approved):

PR #121, opened 2024-04-17
Handle cycles in entitlement graph.

ggreer:
This fixes grant expansion for circular entra groups.

### DIFF
--- a/pkg/sync/expand.go
+++ b/pkg/sync/expand.go
@@ -3,6 +3,7 @@ package sync
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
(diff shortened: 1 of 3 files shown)


pkg/sync/expand.go
```diff
@@ -200,10 +200,13 @@ func (d *EntitlementGraph) AddEdge(srcEntitlementID string, dstEntitlementID str
...
-		Expanded:        false,
-		Shallow:         shallow,
-		ResourceTypeIDs: resourceTypeIDs,
+	_, ok = d.Edges[srcEntitlementID][dstEntitlementID]
+	if !ok {
+		d.Edges[srcEntitlementID][dstEntitlementID] = &edgeInfo{
+			Expanded:        false,
+			Shallow:         shallow,
+			ResourceTypeIDs: resourceTypeIDs,
+		}
```
ggreer, line 209: When testing, I did add an else with a panic(), but we never hit it. I'm not sure if I should keep that in the code or not. I think if we try to add an existing edge, something has gone horribly wrong.

pkg/sync/expand.go
```diff
@@ -20,81 +21,158 @@ type EntitlementGraphAction struct {
...
 }
 
-// IsExpanded returns true if all entitlements in the graph have been expanded.
-func (d *EntitlementGraph) IsExpanded() bool {
-	for entilementID := range d.Entitlements {
-		if !d.IsEntitlementExpanded(entilementID) {
+func (g *EntitlementGraph) Str() string {
+	str := "\n"
+	for id, node := range g.Nodes {
+		str += fmt.Sprintf("node %v entitlement IDs: %v\n", id, node.EntitlementIDs)
```
jirwin, line 58: should we be using a `strings.Builder` for this? Not sure how many iterations happen here

jirwin: Also just realized this is probably mostly used for debugging, so maybe doesn't matter.

ggreer: Yeah I only used it for debugging. It's not called anywhere now.

pkg/sync/expand.go
```diff
@@ -112,19 +190,19 @@ func isInCycle(newCycle []string, cycles [][]string) bool {
...
 	}
 	return arr
 }
 
-func (d *EntitlementGraph) getCycle(visits []string) ([]string, bool) {
-	entitlementID := visits[len(visits)-1]
-	for descendantID := range d.GetDescendants(entitlementID) {
-		tempVisits := make([]string, len(visits))
+func (g *EntitlementGraph) getCycle(visits []int) ([]int, bool) {
+	nodeId := visits[len(visits)-1]
```
jirwin, line 201: do we need to assert that `len(visits) != 0` here?

ggreer: I can add it, but getCycle() is only called in one place, and it always has one element in that array argument.

pkg/sync/expand.go
```diff
@@ -136,74 +214,188 @@ func (d *EntitlementGraph) getCycle(visits []string) ([]string, bool) {
...
+	if node != nil {
+		return
+	}
+
+	g.Nodes[g.NodeCount] = Node{
+		Id:             g.NodeCount,
+		EntitlementIDs: []string{entitlement.Id},
+	}
+	g.NodeCount++
+	// if _, ok := d.Entitlements[entitlement.Id]; !ok {
```
jirwin, line 271: nit: unused

pkg/sync/expand_test.go
```diff
@@ -8,29 +8,118 @@ import (
...
+	require.NoError(t, err)
 	cycles, isCycle := graph.GetCycles()
 	require.True(t, isCycle)
-	require.Equal(t, [][]string{{"2", "3", "4"}}, cycles)
+	require.Equal(t, [][]int{{2, 3, 4}}, cycles)
+}
+
+func TestHandleCycle(t *testing.T) {
+	graph, err := cyclicGraph(t)
+	require.NoError(t, err)
```
jirwin, line 90: I see we do it in the `GetCycles` test, but mayve work asserting that there are cycles in the graph here before fixing them.

jirwin (approved):
LGTM

PR #123, opened 2024-04-22
feat: Add helper for refresh token oauth flow

martinalbert:
This PR adds new implementation of `AuthCredentials` interface - for supporting refresh token oauth flow.

### DIFF
--- a/pkg/uhttp/authcredentials_test.go
+++ b/pkg/uhttp/authcredentials_test.go
@@ -98,3 +98,35 @@ func TestHelpers_OAuth2_JWT_GetClient(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, client)
 }
+
+func TestHelpers_OAuth2_RefreshToken_GetClient(t *testing.T) {
+	rt := &OAuth2RefreshToken{
+		cfg: &oauth2.Config{
+			ClientID:     "test-client-id",
+			ClientSecret: "test-client-secret",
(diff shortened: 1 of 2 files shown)


loganintech (approved):

pkg/uhttp/authcredentials_test.go
```diff
@@ -98,3 +98,35 @@ func TestHelpers_OAuth2_JWT_GetClient(t *testing.T) {
...
+		cfg: &oauth2.Config{
+			ClientID:     "test-client-id",
+			ClientSecret: "test-client-secret",
+			Endpoint: oauth2.Endpoint{
+				TokenURL: "https://test-token-url",
+			},
+		},
+		accessToken:  "test-access-token",
+		refreshToken: "test-refresh-token",
+	}
```
loganintech, line 113: I wonder if there's a service we could use to _actually_ test this but I guess it's pretty much a builtin for Go so maybe it's fine.

martinalbert: Yeah, I don't think so, at least not easily accessible over long period of time, it might be worth to add mock server to test all of these flows.

loganintech: That would be great

martinalbert: Just told @vlnevyhosteny about this and he would love to look into it. I'm going to merge this PR, so I can use it in ArubaCentral. 🙌🏻

PR #124, opened 2024-04-24
feat: Support uncommon profile structures

martinalbert:
## Summary by CodeRabbit

- **New Features**
	- Added `ToProtoValue` function to convert various input types to Protocol Buffer values
	- Introduced new methods for retrieving boolean, list, and struct values from profile structures
	- Enhanced error handling for profile value conversions

- **Improvements**
	- Refined `WithRoleProfile` function to provide more robust value conversion
	- Improved type checking and value extraction from structured profiles


ggreer:
Any reason why this is still a draft? It looks like these helper functions would cut down on boilerplate.

kans:
What would uplift do with a string profile value?  If we still want this, reopen.

PR #125, opened 2024-04-29
External Ticketing

jirwin:
Beginning to draft out what support for external ticketing would look like.


pkg/types/ticket/custom_fields.go
```diff
@@ -0,0 +1,586 @@
+package ticket
```
jirwin, line 1: So much of this could be generated. Maybe worth writing a generator if this is the direction we decide to go.

PR #129, opened 2024-05-07
Add ticket request message

ButternCream:
- Add the ticket request message proto object so we can import and reuse


proto/c1/connector/v2/ticket.proto
```diff
@@ -163,6 +163,10 @@ message TicketsServiceCreateTicketRequest {
   repeated google.protobuf.Any annotations = 8;
 }
 
+message TicketsServiceCreateTicketRequest {
+  TicketRequest request = 1;
```
notanthony, line 167: annotations missing here

ButternCream: fixed https://github.com/ConductorOne/baton-sdk/pull/129/commits/0f673b22637bb9d4ec7c32f8f36ad1670b1bf021

notanthony (approved):

PR #132, opened 2024-05-09
Performance improvements

ggreer:
- Remove getDescendants() and just get node ids directly. This should drastically speed up GetCycles().
- Give up fixing cycles after 10 tries.
- Don't call GetCycles() after fixing cycles.

### DIFF
--- a/pkg/sync/expand.go
+++ b/pkg/sync/expand.go
@@ -233,7 +233,7 @@ func (g *EntitlementGraph) getDescendants(nodeID int) []Node {
 	if !ok {
 		return nil
 	}
-	nodes := make([]Node, len(g.Edges[nodeID]))
+	nodes := make([]Node, 0, len(g.Edges[nodeID]))
 	for dstNodeId := range g.Edges[nodeID] {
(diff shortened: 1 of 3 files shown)


pkg/sync/expand.go
```diff
@@ -386,16 +386,19 @@ func (g *EntitlementGraph) mergeNodes(node1ID int, node2ID int) {
...
-			return
+			return nil
 		}
 
 		// Merge all the nodes in a cycle.
 		for i := 1; i < len(cycles[0]); i++ {
 			g.mergeNodes(cycles[0][0], cycles[0][i])
 		}
 	}
+	return fmt.Errorf("could not fix cycles after %v tries", maxTries)
```
pquerna, line 403: feels like we might want slightly more information here on an error -- though i'm not quite sure what.

ggreer: We should have logged the cycles over in syncer.go before calling this function, so I also don't know what to put here.

jirwin (approved):

PR #134, opened 2024-05-10
Add support for connector-wide capabilities

jirwin:
Currently we emit connector capabilities at the resource type level as a way to describe which operations we support per resource type. With the addition of event feeds and ticketing, we now have capabilities we want to expose that aren't directly tied to a resource type.

As we collect capabilities, track which capabilities are mentioned. If any resource type supports provisioning, it will be included as a connector-wide capability.

We also check to see if the implementation has a event manager set, if it does it means the connector implements an event feed.

example output:
```json
{
  "@type": "type.googleapis.com/c1.connector.v2.ConnectorCapabilities",
  "resourceTypeCapabilities": [
    {
      "resourceType": {
        "id": "group",
        "displayName": "Group",
        "traits": [
          "TRAIT_GROUP"
        ],
        "annotations": [
          {
            "@type": "type.googleapis.com/c1.connector.v2.V1Identifier",
            "id": "group"
          }
        ]
      },
      "capabilities": [
        "CAPABILITY_SYNC",
        "CAPABILITY_PROVISION"
      ]
    },
    {
      "resourceType": {
        "id": "role",
        "displayName": "role",
        "traits": [
          "TRAIT_ROLE"
        ],
        "annotations": [
          {
            "@type": "type.googleapis.com/c1.connector.v2.V1Identifier",
            "id": "role"
          }
        ]
      },
      "capabilities": [
        "CAPABILITY_SYNC",
        "CAPABILITY_PROVISION"
      ]
    },
    {
      "resourceType": {
        "id": "user",
        "displayName": "User",
        "traits": [
          "TRAIT_USER"
        ],
        "annotations": [
          {
            "@type": "type.googleapis.com/c1.connector.v2.V1Identifier",
            "id": "user"
          }
        ]
      },
      "capabilities": [
        "CAPABILITY_SYNC"
      ]
    }
  ],
  "connectorCapabilities": [
    "CAPABILITY_SYNC",
    "CAPABILITY_PROVISION",
    "CAPABILITY_EVENT_FEED"
  ]
}
```

### DIFF
--- a/pkg/connectorbuilder/connectorbuilder.go
+++ b/pkg/connectorbuilder/connectorbuilder.go
@@ -275,6 +275,9 @@ func getCapabilities(ctx context.Context, b *builderImpl) *v2.ConnectorCapabilit
 		if _, ok := rb.(ResourceProvisioner); ok {
 			resourceTypeCapability.Capabilities = append(resourceTypeCapability.Capabilities, v2.Capability_CAPABILITY_PROVISION)
 			connectorCaps[v2.Capability_CAPABILITY_PROVISION] = struct{}{}
+		} else if _, ok = rb.(ResourceProvisionerV2); ok {
+			resourceTypeCapability.Capabilities = append(resourceTypeCapability.Capabilities, v2.Capability_CAPABILITY_PROVISION)
+			connectorCaps[v2.Capability_CAPABILITY_PROVISION] = struct{}{}
 		}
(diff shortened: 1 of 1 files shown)


jirwin:
This is failing proto breaking changes because I pulled the capability enum out of the resource type capability message for reuse. This should be safe for us to merge, as we mostly use this in its serialized JSON form.

ggreer (approved):

PR #135, opened 2024-05-13
Add ListSchemas and GetTicketing tasks

ButternCream:


pkg/cli/ticketing.go
```diff
@@ -2,18 +2,10 @@ package cli
```
ButternCream, line 2: Do we need this file still?

laurenleach (approved):

jirwin (approved):

PR #137, opened 2024-05-16
Add helpers to fetch discrete values from custom fields

jirwin:
Also added more custom field tests.

### DIFF
--- a/pkg/types/ticket/custom_fields.go
+++ b/pkg/types/ticket/custom_fields.go
@@ -15,6 +15,7 @@ import (
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
 )
 
+// CustomFieldForSchemaField returns a typed custom field for a given schema field.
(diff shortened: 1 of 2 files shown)


pkg/types/ticket/custom_fields.go
```diff
@@ -79,14 +80,19 @@ func CustomFieldForSchemaField(id string, schema *v2.TicketSchema, value interfa
...
+		rawValue, err := json.Marshal(value)
+		if err != nil {
+			return nil, err
 		}
 
+		var vals []interface{}
+		err = json.Unmarshal(rawValue, &vals)
+		if err != nil {
+			return nil, err
+		}
```
laurenleach, line 92: why this change?

jirwin: While I was writing tests, i realized that i could make it more generic so you could pass json or the proto and have it work. This is really only used for parsing the template.json when using the dev --create-ticket command.

PR #138, opened 2024-05-16
CreateTicketRequest uses ticket schema instead of a ticket schema ID

jirwin:
Let the caller define the schema to use when creating a ticket instead of looking up the schema.


pkg/tasks/local/ticket.go
```diff
@@ -99,6 +98,7 @@ func (m *localCreateTicket) Process(ctx context.Context, task *v1.Task, cc types
 	ticketRequestBody.CustomFields = cfs
 	ticketReq := &v2.TicketsServiceCreateTicketRequest{
 		Request: ticketRequestBody,
+		Schema:  schema.GetSchema(),
```
laurenleach, line 101: we have to add this to `HandleTask` in `pkg/tasks/c1api/create_ticket.go` also right?

jirwin: Good call. Updated in 3696614

laurenleach (approved):
LGTM

PR #142, opened 2024-05-23
Instrument connector server

ennyjfrick:


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,34 @@
+package metrics
+
+import (
+	"time"
+
+	"github.com/conductorone/baton-sdk/pkg/tasks"
+)
+
+type Handler interface {
```
pquerna, line 9: should we just use the otel span interface here?

ennyjfrick: @pquerna like...maybe? The original purpose of this is to be able to emit SLI metrics when we run in hosted mode  but @ggreer and I had discussed writing non-hosted connector metrics to the c1z file and parse them in an activity somewhere...so I could see keeping a more abstract interface or just having whatever sqlite implementation wrangle the Otel interface

ennyjfrick: @pquerna lmk what you think of the latest commit

ggreer (approved):

PR #144, opened 2024-05-29
Baton IDs

ggreer:
This adds some helper functions for turning any baton resource/entitlement/grant into a unique ID and parsing them back. This preserves parent resources in the ID so that we can grant/revoke in cases where the parent resource is required.

~TODO: escape control characters and parse escaped characters correctly.~


pkg/bid/bid.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+const (
+	BidPrefix            = "bid:"
+	ResourceBidPrefix    = BidPrefix + "r:"
+	EntitlementBidPrefix = BidPrefix + "e:"
+	GrantBidPrefix       = BidPrefix + "g:"
+)
+
+func ToBid(b interface{}) string {
```
morgabra, line 17: I feel like there's a way to be more strict with the typing here.

At the very least these could/should accept a 'proto.Message'.

pkg/bid/bid.go
```diff
@@ -0,0 +1,149 @@
...
+	switch bType := b.(type) {
+	case *v2.Resource:
+		return ResourceToBid(bType)
+	case *v2.Entitlement:
+		return EntitlementToBid(bType)
+	case *v2.Grant:
+		return GrantToBid(bType)
+	}
+	return "☠️" // TODO: panic? this should never happen
```
morgabra, line 26: The pattern I've seen is 2 exported forms of this function:

'MakeBID' which returns an error, and 'MustMakeBID' which is a convenience wrapper that panics.

pkg/bid/bid.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func ParseBid(bidStr string) (interface{}, error) {
```
morgabra, line 29: I wonder if we should make our own 'BID' type/proto that has a oneof for all the possible types?

If it was a oneof callers call 'ParseBID' and then get a concrete type back, then you'd call 'GetResource()' on it (or whatever) to get your parsed type out.

Or you can type switch on the oneof of you don't know what the id is, which is still a little cleaner than a bare interface because we know what types are allowed in the oneof.

pkg/bid/bid.go
```diff
@@ -0,0 +1,149 @@
...
+	case GrantBidPrefix:
+		return ParseGrantBid(bidStr)
+	}
+	return nil, fmt.Errorf("invalid baton id type: %s", bidType)
+}
+
+func parseResourcePart(resourcePart string) (*v2.Resource, error) {
+	resourceId := &v2.ResourceId{}
+	parentResourceId := &v2.ResourceId{}
+	parts := strings.Split(resourcePart, "/")
```
morgabra, line 53: The format here is pretty simple so this is probably overkill/code golf-y, but:

I'm generally a fan of the "lexical scanning" pattern for parsing things like this. Keep some state and scan character-by-character.

It lets you return really detailed errors, really simplifies things like handling escapes, and makes it easy to implement things like arbitrary depth here in this thing. (Are we *sure* we want to limit to exactly 1 depth of hierarchy?)

pkg/bid/bid.go
```diff
@@ -0,0 +1,149 @@
...
+		Principal:   principal,
+	}, nil
+}
+
+// TODO: escape colon and slash
+func EscapeParts(str string) string {
+	return ""
+}
+
+func resourcePartToStr(r *v2.Resource) string {
```
morgabra, line 124: tiny/nit: we'll probably be making a lot of these, it might be worth using a string builder/buffer to pass here instead of copying strings around.

pkg/bid/bid.go
```diff
@@ -0,0 +1,113 @@
...
+
+func resourcePartToStr(r *v2.Resource) (string, error) {
+	rid := r.GetId()
+	resourceType := escapeParts(rid.GetResourceType())
+	resource := escapeParts(rid.GetResource())
+	if resourceType == "" || resource == "" {
+		return "", fmt.Errorf("resource type or id is empty")
+	}
+	if r.ParentResourceId == nil {
+		return fmt.Sprintf("%s/%s", resourceType, resource), nil
```
pquerna, line 61: minor: string.Join as a pattern vs sprintf as a pattern -- most of these are something a stirng.Join can do

pkg/bid/bid.go
```diff
@@ -0,0 +1,113 @@
...
+	GrantBidPrefix       = "g"
+)
+
+type BID interface {
+	proto.Message
+	// TODO: be smarter about this interface.
+	GetAnnotations() []*anypb.Any
+}
+
+func MakeBid(b BID) (string, error) {
```
pquerna, line 25: style nit: can just be called `Make` or `New()`?  bid is duplicative of the package

pkg/bid/bid.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	BidPrefix            = "bid"
+	ResourceBidPrefix    = "r"
+	EntitlementBidPrefix = "e"
+	GrantBidPrefix       = "g"
+)
+
+type BID interface {
```
pquerna, line 19: should we inline a bunch of the docs from the wiki proposal here detailing the format?  either here or as a package level doc string?

pkg/bid/bid.go
pquerna: Looking this is really like an Encoder/Marhsler?

so kinda like `big.MarshalText`?

pkg/bid/scanner.go
```diff
@@ -0,0 +1,90 @@
...
+type tokenType int
+
+const (
+	colon   tokenType = iota
+	slash   tokenType = iota
+	literal tokenType = iota
+	eof     tokenType = iota
+)
+
+type BidScanner struct {
```
pquerna, line 17: can this be a private struct?

pkg/bid/parser.go
```diff
@@ -0,0 +1,245 @@
+package bid
+
+import (
+	"fmt"
+
+	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
+)
+
+func ParseBid(bidStr string) (BID, error) {
```
pquerna, line 9: style: just UnmarshalText or... Parse()

pkg/bid/parser.go
```diff
@@ -0,0 +1,245 @@
...
+	rs := &BidScanner{str: bidStr}
+
+	tType, val, err := rs.NextToken()
+	if err != nil {
+		return nil, err
+	}
+	if tType != literal || val != BidPrefix {
+		return nil, fmt.Errorf("invalid baton id prefix: %s", val)
```
morgabra, line 17: I think we should make an error to return from this package:

The easiest thing to do is like
```
var ErrInvalidBID = errors.New("bid: invalid")
```

Then everywhere wrap it:
```
return nil, fmt.Errorf("%w: missing baton id prefix", ErrInvalidBID)
```

Then callers can actually test for it with `errors.Is` and whatnot.

We could alternatively make a custom error that could include the index of the parse error and stuff, but v0v.

pkg/bid/parser.go
```diff
@@ -0,0 +1,245 @@
...
+		tType, token, err := rs.PeekToken()
+		if err != nil {
+			return nil, err
+		}
+		if tType == eof || tType == colon {
+			// Colon is in the case that this resource is part of an entitlement
+			break
+		}
+		if tType != literal {
+			return nil, fmt.Errorf("invalid baton id resource part: %s %v %v", token, rs.str, rs.index)
```
morgabra, line 88: Yeah, custom error that knows how to print the nice message + index is maybe the play instead 🤔

pkg/bid/scanner.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+type tokenType int
+
+const (
+	colon   tokenType = iota
```
mgaeta, line 11: I've just been reading about `iota` and it seems like the idiomatic way of writing these consts is the simplified: 
```go
const (
	colon tokenType = iota
	slash
	literal
	eof
)
```
In my local testing it does the same thing with the types and values.

pkg/bid/bid.go
ggreer: I added a docs to the top of this file. I think it's the correct format for go docs.

pkg/bid/bid.go
ggreer: Any recommendations for a shorter name? I called the other one bid.Parse(). Maybe bid.String()?

PR #154, opened 2024-06-11
Add immutable annotations to grants and entitlements.

ggreer:
Source ID is the baton ID of the resource that, if changed, would update the immutable grant or entitlement. I'm not sure I want to extend this to resources yet, as those are usually at least partially mutable.


proto/c1/connector/v2/annotation_entitlement.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package c1.connector.v2;
+
+import "google/protobuf/struct.proto";
+
+option go_package = "github.com/conductorone/baton-sdk/pb/c1/connector/v2";
+
+message EntitlementImmutable {
+  optional string source_id = 1;
```
pquerna, line 10: minor: i wasn't sure we want to use the optional flag in proto3 -- i think tooling adoption is messy?

ggreer: Fixed in 2c96485ca651ed96d835ceb70ddf53c17d2955da

pquerna (approved):

PR #155, opened 2024-06-11
ref: `EntitlementGraph` Tests

mgaeta:
- Refactoring the unit tests for `EntitlementGraph`.
  - Made the tests a little clearer with [table driven tests](https://go.dev/wiki/TableDrivenTests).
  - Made a little "DSL" for describing graphs that's sort of human-readable.
- Added [mapset](https://pkg.go.dev/github.com/deckarep/golang-set/v2). 
  - It's only being used in tests, do if we don't want this dependency I can do the same thing with a `map`.
- Note: `TestHandleCycle` fails one a one-node cycle. 
  - I've disabled the test and will have a fix in the follow-up PR.


pkg/sync/expand_test.go
```diff
@@ -63,81 +124,46 @@ func TestRemoveNode(t *testing.T) {
...
-	cycles, isCycle = graph.GetCycles()
-	require.False(t, isCycle)
-	require.Empty(t, cycles)
+	testCases := []struct {
+		expression     string
+		expectedCycles string
+		message        string
+	}{
+		{"1>2>3>4 4>2", "2,3,4", "example"},
+		// {"1>1 2", "1", "simplest"}, // TODO(marcos): Fix this test!
```
ggreer, line 144: 😆

ggreer (approved):

PR #160, opened 2024-06-14
[Performance]: Basic Benchmark & Multi-Put objects

pquerna:
- Adds a basic full Sync() benchmark.
- Based on early profiling of this, add a Multi-Write / Bulk `PutGrants()` function as an interface upgrade to Writer.  Just this change has a ~67.33% reduction in sync time for the basic benchmark.
- Also fix an issue with `goqu` Dialect;  We were trying to use `goqu.Dialect("sqlite3")`.  But if this Dialect isn't loaded, goqu makes a duplicate of the default... which is a huge memory allocation.  So... by using an underscore import of the dialect module, we cut memory allocations by ~16%.

```
goos: linux
goarch: amd64
pkg: github.com/conductorone/baton-sdk/pkg/sync
cpu: Intel(R) Xeon(R) E-2274G CPU @ 4.00GHz
               │   old.text   │              new.text               │
               │    sec/op    │   sec/op     vs base                │
ExpandCircle-8   1644.0m ± 1%   537.1m ± 2%  -67.33% (p=0.000 n=50)
```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -138,6 +138,7 @@ github.com/magiconair/properties v1.8.7 h1:IeQXZAiQcpL9mgcAe1Nu6cX9LLw6ExEHKjN0V
 github.com/magiconair/properties v1.8.7/go.mod h1:Dhd985XPs7jluiymwWYZ0G4Z61jb3vdS329zhj2hYo0=
 github.com/mattn/go-isatty v0.0.20 h1:xfD0iDuEKnDkl03q4limB+vH+GxLEtL/jb4xVJSWWEY=
 github.com/mattn/go-isatty v0.0.20/go.mod h1:W+V8PltTTMOvKvAeJH7IuucS94S2C6jfK/D7dTCTo3Y=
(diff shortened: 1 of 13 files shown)


go.sum
```diff
@@ -137,6 +137,7 @@ github.com/magiconair/properties v1.8.7 h1:IeQXZAiQcpL9mgcAe1Nu6cX9LLw6ExEHKjN0V
 github.com/magiconair/properties v1.8.7/go.mod h1:Dhd985XPs7jluiymwWYZ0G4Z61jb3vdS329zhj2hYo0=
 github.com/mattn/go-isatty v0.0.20 h1:xfD0iDuEKnDkl03q4limB+vH+GxLEtL/jb4xVJSWWEY=
 github.com/mattn/go-isatty v0.0.20/go.mod h1:W+V8PltTTMOvKvAeJH7IuucS94S2C6jfK/D7dTCTo3Y=
+github.com/mattn/go-sqlite3 v1.14.7 h1:fxWBnXkxfM6sRiuH3bqJ4CfzZojMOLVc0UTsTglEghA=
```
jirwin, line 140: Any ideas why this was added? Couldn't find a reason in the diff. This is the c-based driver yeah?

PR #162, opened 2024-06-26
Don't require ticket status and type

laurenleach:
- We don't require status but if the schema didn't include status we were still erroring 
- Change ticket type to not be required since not all ticket systems require type

### DIFF
--- a/pkg/types/ticket/custom_fields.go
+++ b/pkg/types/ticket/custom_fields.go
@@ -244,40 +244,36 @@ func GetCustomFieldValue(field *v2.TicketCustomField) (interface{}, error) {
 func ValidateTicket(ctx context.Context, schema *v2.TicketSchema, ticket *v2.Ticket) (bool, error) {
 	l := ctxzap.Extract(ctx)
 
-	// Look for a matching status
-	foundMatch := ticket.Status == nil
-	for _, status := range schema.GetStatuses() {
-		// Status is not required
-		if ticket.Status == nil {
-			foundMatch = true
-			break
-		}
-
-		if ticket.Status.GetId() == status.GetId() {
-			foundMatch = true
-			break
+	// Validate the ticket status is one defined in the schema
+	// Ticket status is not required so if a ticket doesn't have a status
+	// we don't need to validate, skip the loop in this case
+	validTicketStatus := ticket.Status == nil
+	if !validTicketStatus {
+		for _, status := range schema.GetStatuses() {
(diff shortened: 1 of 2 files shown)


pkg/types/ticket/custom_fields.go
```diff
@@ -265,10 +265,11 @@ func ValidateTicket(ctx context.Context, schema *v2.TicketSchema, ticket *v2.Tic
 	}
 
 	// Look for a matching ticket type
-	foundMatch = false
+	foundMatch = ticket.Type == nil
 	for _, tType := range schema.GetTypes() {
 		if ticket.Type == nil {
```
mgaeta, line 270: The code block
```go
if ticket.Type == nil {
	foundMatch = true
	break
}
```
can be removed from the loop:
```go
// Look for a matching status
// Status is not required
foundMatch := ticket.Status == nil ||
if !foundMatch {
  for _, status := range schema.GetStatuses() {
    if ticket.Status.GetId() == status.GetId() {
	    foundMatch = true
	    break
    }
  }
}
```

pkg/types/ticket/custom_fields.go
```diff
@@ -265,10 +265,11 @@ func ValidateTicket(ctx context.Context, schema *v2.TicketSchema, ticket *v2.Tic
 	}
 
 	// Look for a matching ticket type
-	foundMatch = false
+	foundMatch = ticket.Type == nil
```
btipling, line 268: I might add a more specific comment here and above to explain why we do this, if a ticket doesn't have a status or a type we don't need to validate, skip the loop in this case

pkg/types/ticket/custom_fields.go
```diff
@@ -265,10 +265,11 @@ func ValidateTicket(ctx context.Context, schema *v2.TicketSchema, ticket *v2.Tic
 	}
 
 	// Look for a matching ticket type
-	foundMatch = false
```
btipling, line 268: `hasValidTicketType`

pkg/types/ticket/custom_fields.go
```diff
@@ -245,7 +245,7 @@ func ValidateTicket(ctx context.Context, schema *v2.TicketSchema, ticket *v2.Tic
 	l := ctxzap.Extract(ctx)
 
 	// Look for a matching status
-	foundMatch := false
```
btipling, line 248: `validTicketStatus`

btipling (approved):

PR #163, opened 2024-06-26
feat: Allow `WithJSONResponse()` and `WithXMLResponse()` to work with empty response bodies

mgaeta:
A common pattern in connectors is to create a helper function like `makeRequest()` that calls `client.wrapper.Do()`. When dealing with a JSON API, most of the client methods expect a response body. However, some APIs respond with an empty body when they `204`. To handle this we have to do something a little janky: 
```go
doOptions := []uhttp.DoOption{
  WithConfluenceRatelimitData(&ratelimitData),
}
if target != nil {
  doOptions = append(doOptions, uhttp.WithJSONResponse(target))
}
response, err := c.wrapper.Do(request, doOptions...)
```

I want to change `WithJSONResponse()` to take `nil` to mean that we expect the body to be empty and that there is no JSON.

I think it's also totally reasonable to not do this and just make sure not to call `WithJSONResponse()` when we're expecting a `204`.

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -107,6 +107,9 @@ func WithXMLResponse(response interface{}) DoOption {
 		if !helpers.IsXMLContentType(resp.Header.Get(ContentType)) {
 			return fmt.Errorf("unexpected content type for xml response: %s", resp.Header.Get(ContentType))
 		}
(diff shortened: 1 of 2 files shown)


ggreer (approved):

ggreer:
Just need to fix the linter error and this is good to go.

PR #165, opened 2024-06-27
Rate limit support

ggreer:
The current behavior is to retry recoverable errors every second and linearly backoff. This adds backoff that respects rate limit annotations.

## Summary by CodeRabbit

- **New Features**
	- Enhanced response handling across resource and entitlement listing methods, ensuring structured responses are returned even in error scenarios.
	- Improved synchronization error handling by integrating annotations into retry logic, providing better context during operations.
	- Added granular error reporting for server errors in the HTTP client, improving clarity in error responses.
	- Introduced detailed error reporting for rate limiting and service unavailability in the HTTP client, enhancing user awareness of retry conditions.

- **Bug Fixes**
	- Adjusted error handling to return additional context alongside errors, facilitating better debugging and operational insights.

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"time"
 
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
 	"github.com/conductorone/baton-sdk/pkg/ratelimit"
@@ -19,6 +20,7 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -213,6 +215,20 @@ func WithResponse(response interface{}) DoOption {
 	}
 }
 
+func retryValueToTime(value string, now time.Time) (time.Time, error) {
```
ggreer, line 218: We should probably use ExtractRateLimitData in `pkg/ratelimit/http.go`, since it handles lots of different rate limit headers. If that doesn't do what we want, it should be modified to improve its behavior.

pkg/uhttp/wrapper.go
```diff
@@ -279,6 +295,22 @@ func (c *BaseHttpClient) Do(req *http.Request, options ...DoOption) (*http.Respo
 	case http.StatusRequestTimeout:
 		return resp, status.Error(codes.DeadlineExceeded, resp.Status)
 	case http.StatusTooManyRequests:
+		retryAfter := resp.Header.Get("Retry-After")
```
ggreer, line 298: Ideally we'd make this code as general as possible, so that any 4xx/5xx status code with rate limit headers gets the rate limit description set on the error. Then the function signatures in syncer.go and connectorbuilder.go wouldn't need to change.

shackra: I made that code into a function in recent commits, however, the only status errors that carry a `retry-after` header are `Too Many Requests`, `Service Unavailable`, and `Moved Permanently`; AFAIK.

PR #166, opened 2024-07-02
Offer configuration schema to connectors

shackra:


shackra:
tests currently fail because `go` fetches the `baton-sdk` from the `main` branch:

```
/tmp/baton-sdk-schema-87803883 via 🐹 v1.22.3 
➜  go build -buildmode=plugin -o example plugin.go
go: finding module for package github.com/conductorone/baton-sdk/pkg/configschema
go: downloading github.com/conductorone/baton-sdk v0.1.47
plugin.go:4:2: cannot find module providing package github.com/conductorone/baton-sdk/pkg/configschema
```

IDK if I should think of a workaround for this just for the unit tests 🤷🏻‍♂️

shackra:
I'm unable to bypass the error:

```
failed to open plugin:
plugin.Open("/tmp/baton-alternative-location-3073697725/plugin"): plugin was built with a different version of package github.com/conductorone/baton-sdk/pkg/configschema
```

In tests.

This is a problem during development because the package `pkg/configschema` can differ between the contents of `main` and the contents of whatever branch someone is working on.

I did attempt the following workarounds without success:

1. adding a `go.mod` with redirect to the project in disk
2. doing the above step but generating a `go.sum`
3. doing the above step but adding a temporal `plugin.go` file with 1 import (that import was removed from `go.mod` because it wasn't actually used lol)
4. copying `go.sum` from the project (didn't work either, but IIRC `go mod tidy` that was ran afterwards may have deleted the import, forcing `go build` to fetch it from the repo on GitHub. My bad)

Ideally, during development and testing, the same `pkg/configschema` should be used when compiling the schema definition of the configuration as a plugin, otherwise the `plugin` module will refuse to open the plugin.

shackra:
According to `git bisect`, things went south at commit 3a5a39cf8ae98ba3d37909ccb4366aa25e8d916a.
Commit 4b9a5361348252542a237c19b23e29000062f5da was the good commit, as the module `pkg/configschema` compiles the plugin and pass tests successfully.

pkg/field/fields.go
```diff
@@ -0,0 +1,114 @@
...
+		DefaultValue: "",
+	}
+
+	for _, o := range optional {
+		field = o(field)
+	}
+
+	return field
+}
+
```
ggreer, line 97: It's not a blocker for this PR, but we will want a StringList field, which parses a comma separated list of strings. I know a couple of connectors use that.

ggreer (commented):
I think we just need to change one package name and make sure the old config format still works with existing connectors, then this is ready to merge.

pkg/field/fields.go
shackra: should we (you?) fill that one as a ticket on Jira?

pkg/cli/service_unix.go
```diff
@@ -13,18 +14,14 @@ func isService() bool {
...
-	return nil
-}
-
-func runService(ctx context.Context, name string) (context.Context, error) {
+func runService(ctx context.Context, _ string) (context.Context, error) {
 	return ctx, nil
 }
 
-func initLogger(ctx context.Context, name string, opts ...logging.Option) (context.Context, error) {
+func initLogger(ctx context.Context, _ string, opts ...logging.Option) (context.Context, error) {
```
ggreer, line 21: Weird that we always ignored the name in this function. 🤷

ggreer (approved):
🎉

PR #174, opened 2024-07-10
Add struct to carry the schema of fields and constrains

shackra:


pkg/field/struct.go
```diff
@@ -0,0 +1,6 @@
+package field
+
+type Configuration struct {
+	Fields     []SchemaField
+	Constrains []SchemaFieldRelationship
```
ggreer, line 5: I'd call this "Constraints"

ggreer (commented):
I think just change the name and it's good to go.

ggreer (approved):

PR #175, opened 2024-07-11
Add configuration validation after it is loaded

shackra:


pkg/field/validation.go
```diff
@@ -0,0 +1,108 @@
...
+func Validate(c Configuration, v *viper.Viper) error {
+	present := make(map[string]int)
+	missingFieldsError := &ErrConfigurationMissingFields{}
+
+	// check if required fields are present
+	for _, f := range c.Fields {
+		isNonZero := false
+		switch f.FieldType {
+		case reflect.Bool:
+			isNonZero = v.GetBool(f.FieldName)
```
ggreer, line 45: I made `--insecure-skip-verify` required in baton-ldap to test this out. I have to set it to true or else it errors with `field insecure-skip-verify of type bool is marked as required but it has a zero-value`

Also we should probably print out whether fields are required or not in the help text.

pkg/field/validation.go
```diff
@@ -0,0 +1,108 @@
...
+
+func makeNeededTogetherError(fields map[string]int, relation SchemaFieldRelationship) error {
+	var found []string
+	for _, f := range relation.Fields {
+		if fields[f.FieldName] == 0 {
+			found = append(found, f.FieldName)
+		}
+	}
+
+	return fmt.Errorf("fields marked as needed together are missing: %s", strings.Join(found, ", "))
```
ggreer, line 107: This has a false positive if you omit all fields that are need together. For example, baton-ldap fails with `fields marked as needed together are missing: grant-entitlement, grant-principal` even though I don't specify either of those flags.

pkg/field/validation.go
shackra: > I made `--insecure-skip-verify` required in baton-ldap to test this out. I have to set it to true or else it errors with `field insecure-skip-verify of type bool is marked as required but it has a zero-value`

The problem with this is that Viper does not return pointer values, like `*bool`, giving us a third option in such case. We only have the zero-value of each type to know if the flag/envvar was set or not.

shackra: > Also we should probably print out whether fields are required or not in the help text.

Implemented

ggreer: Considering these restrictions, I think we just just not allow bools to be required. Otherwise "required" just means "it must be true", in which case... why even have the bool argument?

shackra: indeed

I'll add that restriction

pkg/field/fields.go
```diff
@@ -78,6 +85,10 @@ func BoolField(name string, optional ...fieldOption) SchemaField {
 		field = o(field)
 	}
 
+	if field.Required {
+		panic(fmt.Sprintf("requiring %s of type %s does not make sense", field.FieldName, field.FieldType))
+	}
```
shackra, line 90: @ggreer thoughts?

ggreer: Seems fine.

shackra: cool, let me mark the flags as required in Cobra and I'll let you know when this is ready to go

shackra:
@ggreer this is ready to go now :)!

ggreer (approved):

PR #177, opened 2024-07-17
Add missing relationship constraint

shackra:
- add constrain check to require at least one field/environment variable from a set
- add missing check for zero-value of string slice (missed in #176)


shackra:
new constrain tested with baton-ldap and I can confirm it works!

PR #180, opened 2024-07-18
Add unit tests

shackra:
Ensures behavior of CLI and usage of envvars with unit tests!

### DIFF
--- a/internal/tests/passingvalues_test.go
+++ b/internal/tests/passingvalues_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 func TestPassingValuesToFieldsViaCLI(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*20)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*10)
 	defer cancel()
 
 	stringfield := field.StringField("string-field", field.WithRequired(true))
@@ -32,7 +32,7 @@ func TestPassingValuesToFieldsViaCLI(t *testing.T) {
 }
 
 func TestPassingValuesToFieldsViaENVVARS(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*20)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*10)
 	defer cancel()
 
 	stringfield := field.StringField("string-field", field.WithRequired(true))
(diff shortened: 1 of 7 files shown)


internal/tests/passingvalues_test.go
```diff
@@ -0,0 +1,33 @@
...
+	carrier := field.NewConfiguration([]field.SchemaField{stringfield, intfield, boolfield, stringslicefield})
+
+	os.Args = []string{"cmd",
+		"--string-field", "foo",
+		"--int-field", "100",
+		"--bool-field",
+		`--string-slice-field="foo,bar,foobarr"`,
+	}
+
+	v, err := entrypoint(carrier)
```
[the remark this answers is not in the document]
shackra: turns out we don't support string slice in the old code https://github.com/ConductorOne/baton-sdk/blob/5aba0b64b0804100e2e1156004a67b4b3af6a922/pkg/cli/config.go#L113

pkg/field/fields.go
```diff
@@ -51,8 +51,8 @@ func (s SchemaField) String() (string, error) {
 	return value, nil
 }
 
-// StringArray retuns the default value as a string array.
-func (s SchemaField) StringArray() ([]string, error) {
+// StringSlice retuns the default value as a string array.
+func (s SchemaField) StringSlice() ([]string, error) {
```
ggreer, line 55: Do existing connectors use StringArray or StringArrayField? I know the new name is more accurate, but I'd like to avoid breaking connectors when upgrading their baton-sdk dependency.

internal/tests/passingvalues_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/conductorone/baton-sdk/pkg/field"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPassingValuesToFieldsViaCLI(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*10)
```
ggreer, line 13: I see `time.Millisecond*10` in quite a few places. It should probably be a const so any updates to the timeout don't require a big diff.

Also why 10 milliseconds and not 1 or 100? I tried different values and anything above  20ms seems to spit out errors like `panic: unexpected call to os.Exit(0) during test`.

pkg/field/fields.go
shackra: This field actually doesn't work at all btw

internal/tests/passingvalues_test.go
shackra: > Also why 10 milliseconds and not 1 or 100?

it was on a whim I chose that number lol

ggreer (approved):

PR #182, opened 2024-07-19
Set flags if their envvar counterparts are set

shackra:
Prevents conflicts with Cobra for not using any required flag when invoking a connector

- https://github.com/spf13/viper/issues/397

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -13,6 +13,7 @@ import (
 	"github.com/conductorone/baton-sdk/pkg/connectorrunner"
 	"github.com/conductorone/baton-sdk/pkg/field"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


shackra:
tested using baton-ldap, having `BATON_DOMAIN` set:

```
➜  ./dist/linux_amd64/baton-ldap
{"level":"error","ts":1721430672.5190167,"caller":"baton-ldap/main.go:58","msg":"error creating connector","error":"LDAP Result Code 200 \"Network Error\": dial tcp [::1]:389: connect: connection refused"}
LDAP Result Code 200 "Network Error": dial tcp [::1]:389: connect: connection refused
```

(the error happens because I'm not running the LDAP service)

Previously we would get an error that you need to either use the flags domain or url.

ggreer (approved):

PR #183, opened 2024-07-22
Remove client id/file mutually exclusive relationship

laurenleach:
Remove client id/file mutually exclusive relationship

Because of the default value set for file, setting client-id gives the error `fields marked as mutually exclusive were set: file, client-id`


ggreer:
We can save a c1z and run in service mode? I didn't know that was a thing.

shackra:
> We can save a c1z and run in service mode? I didn't know that was a thing.

me neither, I "translated" what I saw in the original code

ggreer (approved):

PR #186, opened 2024-07-23
Add dependent schema relationship type

laurenleach:
Add dependent schema relationship type

### DIFF
--- a/pkg/field/relationships.go
+++ b/pkg/field/relationships.go
@@ -6,11 +6,75 @@ const (
 	RequiredTogether Relationship = iota + 1
 	MutuallyExclusive
 	AtLeastOne
+	DependentOn
 )
+
+type SchemaFieldRelationshipI interface {
+	ValidateConstraint(fieldsPresent map[string]int) error
+	GetKind() Relationship
+}
+
+type DependentSchemaFieldRelationship struct {
+	RequiredFields  []SchemaField
+	DependentFields []SchemaField
+}
 
 type SchemaFieldRelationship struct {
 	Kind   Relationship
 	Fields []SchemaField
+}
+
+func (s DependentSchemaFieldRelationship) ValidateConstraint(fieldsPresent map[string]int) error {
+	var dependentPresentCount int
+	var requiredPresentCount int
+	dependentPresent := make([]string, 0)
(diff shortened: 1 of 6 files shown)


ggreer (approved):

pkg/field/relationships.go
```diff
@@ -8,11 +8,65 @@ const (
...
 )
 
+type SchemaFieldRelationshipI interface {
+	ValidateConstraint(fieldsPresent map[string]int) error
+}
+
+type DependentSchemaFieldRelationship struct {
+	RequiredFields  []SchemaField
+	DependentFields []SchemaField
+}
```
shackra, line 18: @laurenleach I don't follow why a new struct for tracking required fields and dependent fields (of those required fields?), why add a new `const` below `AtLeastOne` and extend `SchemaFieldRelationship` with a new field?

The problem with array/slice of an interface is that `[]DependentSchemaFieldRelationship` or `[]SchemaFieldRelationship` doesn't implement a `[]SchemaFieldRelationshipI`

ggreer: I like this idea.

shackra: my objection comes from this btw https://stackoverflow.com/questions/44319906/why-golang-struct-array-cannot-be-assigned-to-an-interface-array

laurenleach: why do we need `[]DependentSchemaFieldRelationship` or `[]SchemaFieldRelationship`? we can use `[]SchemaFieldRelationshipI`

shackra: `SchemaFieldRelationshipI` is needed in so far as changes to the previous API were introduced, my objection is twofold:

1. We don't know how users outside C1 will try to use the SDK
2. We can still extend the concrete type `SchemaField` to achieve the new relationship constraint

Regarding nº 1 some users can find [this pitfall](https://stackoverflow.com/questions/44319906/why-golang-struct-array-cannot-be-assigned-to-an-interface-array) when developing their own connectors relaying on the introduced interface `SchemaFieldRelationshipI`.

With nº 2 I've opened a PR showing that extending what is already there can help https://github.com/ConductorOne/baton-sdk/pull/187

PR #187, opened 2024-07-23
Add relationship constrain "depend on" keeping the previous API

shackra:


pkg/field/relationships.go
```diff
@@ -88,9 +36,10 @@ func FieldsAtLeastOneUsed(fields ...SchemaField) SchemaFieldRelationship {
 	}
 }
 
-func FieldsDependentOn(requiredFields []SchemaField, dependentFields []SchemaField) DependentSchemaFieldRelationship {
-	return DependentSchemaFieldRelationship{
-		RequiredFields:  requiredFields,
-		DependentFields: dependentFields,
+func FieldsDependOn(dependent []SchemaField, expected []SchemaField) SchemaFieldRelationship {
```
ggreer, line 39: Can we keep the same function name as before?

ggreer (approved):

PR #190, opened 2024-07-24
Fix release version action.

ggreer:
Looks like echo doesn't have a -e flag on some platforms.


shackra (approved):
we probably should use `printf` but let's try this solution first

PR #192, opened 2024-07-25
Updating uhttp package

mchavez:
- Updating uhttp package

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -233,27 +233,15 @@ func WithAcceptXMLHeader() RequestOption {
 }
 
 func WithContentTypeFormHeader() RequestOption {
-	return func() (io.ReadWriter, map[string]string, error) {
-		return nil, map[string]string{
-			"Content-Type": "application/x-www-form-urlencoded",
-		}, nil
-	}
+	return WithHeader("Content-Type", "application/x-www-form-urlencoded")
 }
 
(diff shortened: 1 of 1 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -215,6 +232,30 @@ func WithAcceptXMLHeader() RequestOption {
 	return WithHeader("Accept", "application/xml")
 }
 
+func WithContentTypeFormHeader() RequestOption {
```
mgaeta, line 235: It looks like there are already some helper functions so you could do
```go
func WithContentTypeFormHeader() RequestOption {
	return WithHeader("Content-Type", "application/x-www-form-urlencoded")
}
...
func WithAcceptVndJSONHeader() RequestOption {
	return WithHeader("Accept", "application/vnd.api+json")
}
```
or add new helpers and do something like:
```go
func WithContentType(value string) RequestOption {
	return WithHeader("Content-Type", value)
}
func WithContentTypeFormHeader() RequestOption {
	return WithContentType("application/x-www-form-urlencoded")
}
...
```

mchavez: Yes. Refactoring is always valid. Thanks..!!

mchavez: Fixed

ggreer:
Closing because https://github.com/ConductorOne/baton-sdk/pull/193 addresses this.

PR #193, opened 2024-07-25
Refactoring uhttp package code

mchavez:
- Refactoring uhttp package code

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -233,27 +233,15 @@ func WithAcceptXMLHeader() RequestOption {
 }
 
 func WithContentTypeFormHeader() RequestOption {
-	return func() (io.ReadWriter, map[string]string, error) {
-		return nil, map[string]string{
-			"Content-Type": "application/x-www-form-urlencoded",
-		}, nil
-	}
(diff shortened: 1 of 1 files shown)


ggreer (approved):

ggreer (commented):
Oh. I think the name should be WithFormBody or something. Many forms of http requests can contain bodies.

pkg/uhttp/wrapper.go
```diff
@@ -203,16 +203,53 @@ func WithJSONBody(body interface{}) RequestOption {
...
+func WithContentTypeFormHeader() RequestOption {
+	return WithContentType("application/x-www-form-urlencoded")
+}
+
+func WithContentTypeVndHeader() RequestOption {
+	return WithContentType("application/vnd.api+json")
+}
+
+func WithAcceptVndJSONHeader() RequestOption {
+	return WithAccept("application/vnd.api+json")
```
ggreer, line 244: Since all these strings ("application/json", "application/xml", etc) are duplicated in several places, maybe it makes sense to make them constants.

mchavez: Fixed

PR #194, opened 2024-07-26
Adding cache to uhttp package

mchavez:
- Adding cache to uhttp package.

## Summary by CodeRabbit

- **New Features**
	- Introduced a caching mechanism for HTTP responses, enhancing performance by minimizing repeated network calls.
	- Added a `CacheConfig` to allow dynamic configuration of caching parameters like TTL and size.
	- Enhanced `BaseHttpClient` with caching capabilities for GET requests.
	- New request option functions for better flexibility in setting HTTP headers.

- **Bug Fixes**
	- Improved error handling during HTTP client initialization and response caching.

- **Tests**
	- Updated test suite to validate the new context-based configuration for `BaseHttpClient` and its behavior under various scenarios.

### DIFF
--- a/pkg/uhttp/gocache.go
+++ b/pkg/uhttp/gocache.go
@@ -19,14 +19,14 @@ type GoCache struct {
 	rootLibrary *bigcache.BigCache
 }
 
-func NewGoCache(ctx context.Context, ttl int32, cacheMaxSize int, isLogLevelActive bool) (GoCache, error) {
+func NewGoCache(ctx context.Context, ttl int32, cacheMaxSize int, isLogLevel bool) (GoCache, error) {
 	l := ctxzap.Extract(ctx)
 	config := bigcache.Config{
 		// number of shards (must be a power of 2)
 		Shards: 1024,
 
 		// time after which entry can be evicted
-		LifeWindow: time.Duration(ttl) * time.Second, // BATON_CACHE_TTL
+		LifeWindow: time.Duration(ttl) * time.Second,
 
 		// Interval between removing expired entries (clean up).
 		// If set to <= 0 then no action is performed.
@@ -40,7 +40,7 @@ func NewGoCache(ctx context.Context, ttl int32, cacheMaxSize int, isLogLevelActi
 		MaxEntrySize: 500,
 
 		// prints information about additional memory allocation
-		Verbose: isLogLevelActive, // BATON_CACHE_DISABLE
+		Verbose: isLogLevel,
 
 		// cache will not allocate more memory than this limit, value in MB
 		// if value is reached then the oldest entries can be overridden for the new ones
--- a/pkg/uhttp/gocache.go
+++ b/pkg/uhttp/gocache.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"log"
 	"net/http"
 	"net/http/httputil"
 	"time"
@@ -18,15 +19,53 @@ type GoCache struct {
 }
 
 func NewGoCache(ctx context.Context, ttl int32) (GoCache, error) {
-	eviction := time.Duration(ttl) * time.Minute
-	c, err := bigcache.New(ctx, bigcache.DefaultConfig(eviction))
-	if err != nil {
-		return GoCache{}, err
+	config := bigcache.Config{
+		// number of shards (must be a power of 2)
+		Shards: 1024,
+
+		// time after which entry can be evicted
+		LifeWindow: 10 * time.Minute,
+
+		// Interval between removing expired entries (clean up).
+		// If set to <= 0 then no action is performed.
+		// Setting to < 1 second is counterproductive — bigcache has a one second resolution.
+		CleanWindow: 5 * time.Minute,
+
+		// rps * lifeWindow, used only in initial memory allocation
+		MaxEntriesInWindow: 1000 * 10 * 60,
+
+		// max entry size in bytes, used only in initial memory allocation
+		MaxEntrySize: 500,
+
+		// prints information about additional memory allocation
+		Verbose: true,
+
+		// cache will not allocate more memory than this limit, value in MB
+		// if value is reached then the oldest entries can be overridden for the new ones
+		// 0 value means no size limit
+		// Default value 1GB eq 1024MB
+		HardMaxCacheSize: 1024,
+
+		// callback fired when the oldest entry is removed because of its expiration time or no space left
+		// for the new entry, or because delete was called. A bitmask representing the reason will be returned.
+		// Default value is nil which means no callback and it prevents from unwrapping the oldest entry.
+		OnRemove: nil,
+
+		// OnRemoveWithReason is a callback fired when the oldest entry is removed because of its expiration time or no space left
+		// for the new entry, or because delete was called. A constant representing the reason will be passed through.
+		// Default value is nil which means no callback and it prevents from unwrapping the oldest entry.
+		// Ignored if OnRemove is specified.
+		OnRemoveWithReason: nil,
+	}
+	cache, initErr := bigcache.New(ctx, config)
+	if initErr != nil {
+		log.Fatal(initErr)
+		return GoCache{}, initErr
 	}
 
 	gc := GoCache{
-		ttl:         eviction,
-		rootLibrary: c,
+		ttl:         10 * time.Minute,
+		rootLibrary: cache,
 	}
 
 	return gc, nil
--- a/pkg/uhttp/gocache.go
+++ b/pkg/uhttp/gocache.go
@@ -53,8 +53,8 @@ func isOk(e error) bool {
 	return e == nil
 }
 
-func (c *GoCache) Get(key string) (*http.Response, error) {
-	entry, found := c.rootLibrary.Get(key)
+func (g *GoCache) Get(key string) (*http.Response, error) {
+	entry, found := g.rootLibrary.Get(key)
 	if isOk(found) {
 		r := bufio.NewReader(bytes.NewReader(entry))
 		resp, err := http.ReadResponse(r, nil)
@@ -68,9 +68,9 @@ func (c *GoCache) Get(key string) (*http.Response, error) {
 	return nil, nil
 }
 
-func (c *GoCache) Set(key string, value *http.Response) error {
+func (g *GoCache) Set(key string, value *http.Response) error {
 	cacheableResponse, _ := httputil.DumpResponse(value, true)
-	err := c.rootLibrary.Set(key, cacheableResponse)
+	err := g.rootLibrary.Set(key, cacheableResponse)
 	if err != nil {
 		return err
 	}
@@ -78,8 +78,8 @@ func (c *GoCache) Set(key string, value *http.Response) error {
 	return nil
 }
 
-func (c *GoCache) GetString(key string) string {
-	entry, found := c.rootLibrary.Get(key)
+func (g *GoCache) GetString(key string) string {
+	entry, found := g.rootLibrary.Get(key)
 	if found == nil {
 		return string(entry)
 	}
@@ -96,8 +96,8 @@ func (c *GoCache) SetString(key string, value string) error {
 	return nil
 }
 
-func (c *GoCache) SetBytes(key string, value []byte) error {
-	err := c.rootLibrary.Set(key, value)
+func (g *GoCache) SetBytes(key string, value []byte) error {
+	err := g.rootLibrary.Set(key, value)
 	if err != nil {
 		return err
 	}
@@ -105,8 +105,8 @@ func (c *GoCache) SetBytes(key string, value []byte) error {
 	return nil
 }
 
-func (c *GoCache) Delete(key string) error {
-	err := c.rootLibrary.Delete(key)
+func (g *GoCache) Delete(key string) error {
+	err := g.rootLibrary.Delete(key)
 	if err != nil {
 		return err
 	}
@@ -114,8 +114,8 @@ func (c *GoCache) Delete(key string) error {
 	return nil
 }
 
-func (c *GoCache) Clear() error {
-	err := c.rootLibrary.Reset()
+func (g *GoCache) Clear() error {
+	err := g.rootLibrary.Reset()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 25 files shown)


pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,129 @@
...
+	gc := GoCache{
+		ttl:         eviction,
+		rootLibrary: c,
+	}
+
+	return gc, nil
+}
+
+func GetCacheKey(req *http.Request) string {
+	return req.URL.String()
```
ggreer, line 36: I think we should include some http headers in the cache key. For example, hitting the same URL with `Accept: application/json` and `Accept: application/xml` will return very different responses.

The headers I can think of off the top of my head:

- [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept)
- [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) (because using different credentials probably means we'll get different results)
- [Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie)
- [Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range)

pkg/uhttp/wrapper.go
```diff
@@ -128,19 +147,32 @@ func WithResponse(response interface{}) DoOption {
 }
 
 func (c *BaseHttpClient) Do(req *http.Request, options ...DoOption) (*http.Response, error) {
+	var cacheKey string = GetCacheKey(req)
```
ggreer, line 150: We should only generate the cache key if the request method is GET.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,129 @@
...
+	"net/http"
+	"net/http/httputil"
+	"time"
+
+	bigcache "github.com/allegro/bigcache/v3"
+)
+
+type GoCache struct {
+	ttl         time.Duration
+	rootLibrary *bigcache.BigCache
```
ggreer, line 17: It'd be nice to have some stats in this struct so we could keep track of cache hit/miss rates.

pkg/uhttp/wrapper.go
```diff
@@ -32,16 +41,26 @@ type (
...
 	RequestOption func() (io.ReadWriter, map[string]string, error)
 )
 
 func NewBaseHttpClient(httpClient *http.Client) *BaseHttpClient {
+	var ctx = context.TODO()
+	l := ctxzap.Extract(ctx)
+	cache, err := NewGoCache(ctx, int32(10))
+	if err != nil {
+		l.Error("cache error", zap.Any("err", err))
+		return nil
```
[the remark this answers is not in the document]
mchavez: Yes, I did not add it because it will brake any existing implementation of the uhttp pkg.
However, I'm going to add it :)

pkg/uhttp/wrapper.go
```diff
@@ -32,16 +41,26 @@ type (
...
+		HttpClient    *http.Client
+		baseHttpCache GoCache
 	}
 
 	DoOption      func(resp *WrapperResponse) error
 	RequestOption func() (io.ReadWriter, map[string]string, error)
 )
 
 func NewBaseHttpClient(httpClient *http.Client) *BaseHttpClient {
+	var ctx = context.TODO()
```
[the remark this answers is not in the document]
mchavez: ok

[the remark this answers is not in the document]
mchavez: Fixed

pkg/uhttp/wrapper.go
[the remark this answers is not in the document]
mchavez: Fixed

pkg/uhttp/wrapper_test.go
```diff
@@ -385,7 +387,7 @@ func TestWrapper_NewRequest(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			client := NewBaseHttpClient(http.DefaultClient)
+			client, _ := NewBaseHttpClient(ctx, http.DefaultClient)
```
ggreer, line 390: The tests could require.NoError() when creating base clients.

pkg/uhttp/gocache.go
ggreer: Actually it looks like bigcache has its own stats function that we can call: https://pkg.go.dev/github.com/allegro/bigcache#BigCache.Stats

pkg/uhttp/wrapper.go
mchavez: Fixed

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func NewGoCache(ctx context.Context, ttl int32, cacheMaxSize int, isLogLevel bool) (GoCache, error) {
+	l := ctxzap.Extract(ctx)
+	config := bigcache.Config{
+		// number of shards (must be a power of 2)
+		Shards: 1024,
```
ggreer, line 26: I don't think we need that many shards. 4 or 8 is probably enough. Higher shard counts are only needed for high concurrency, and we won't have that many outstanding http requests.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return gc, nil
+}
+
+func GetCacheKey(req *http.Request) string {
+	return req.URL.String()
+}
+
+func CopyResponse(resp *http.Response) *http.Response {
```
ggreer, line 80: Is the plan to use this? I don't see it called anywhere.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+	return &c
+}
+
+func isOk(e error) bool {
+	return e == nil
+}
+
+func (g *GoCache) Get(key string) (*http.Response, error) {
+	entry, found := g.rootLibrary.Get(key)
+	if isOk(found) {
```
ggreer, line 99: Seems like you could just check if the error is not nil and return early here. No need for an isOk function.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+
+		return resp, nil
+	}
+
+	return nil, nil
+}
+
+func (g *GoCache) Set(key string, value *http.Response) error {
+	cacheableResponse, _ := httputil.DumpResponse(value, true)
```
ggreer, line 113: Need to check the error here. We probably want to crash if we can't serialize the response.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+	err := g.rootLibrary.Set(key, cacheableResponse)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (g *GoCache) GetString(key string) string {
```
ggreer, line 122: What do we need GetString, SetString, and SetBytes for? It seems like this cache should only be used for http responses.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.uber.org/zap"
+)
+
+type GoCache struct {
+	ttl         time.Duration
+	rootLibrary *bigcache.BigCache
+}
+
+func NewGoCache(ctx context.Context, ttl int32, cacheMaxSize int, isLogLevel bool) (GoCache, error) {
```
ggreer, line 22: This should probably take a config struct instead of a bunch of args. That way we don't have to change the function definition if we add more configuration options in the future.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+	l := ctxzap.Extract(ctx)
+	config := bigcache.Config{
```
ggreer, line 24: I think it's less code to initialize config to `bigcache.DefaultConfig(time.Duration(ttl) * time.Second)`, then override the config fields we want to override.

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,170 @@
...
+	"net/http/httputil"
+	"time"
+
+	bigcache "github.com/allegro/bigcache/v3"
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.uber.org/zap"
+)
+
+type GoCache struct {
+	ttl         time.Duration
```
ggreer, line 18: This value is never read, so I think we can delete it.

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,129 @@
...
+	rootLibrary *bigcache.BigCache
+}
+
+func NewGoCache(ctx context.Context, ttl int32) (GoCache, error) {
+	eviction := time.Duration(ttl) * time.Minute
+	c, err := bigcache.New(ctx, bigcache.DefaultConfig(eviction))
```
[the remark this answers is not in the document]
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/wrapper_test.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/gocache.go
mchavez: Fixed

pkg/uhttp/wrapper.go
```diff
@@ -32,16 +41,26 @@ type (
...
+	l := ctxzap.Extract(ctx)
+	cache, err := NewGoCache(ctx, int32(10))
```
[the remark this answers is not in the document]
mchavez: Fixed

jirwin:
@coderabbitai full review

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (g *GoCache) Clear() error {
+	err := g.rootLibrary.Reset()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
ggreer: Note to humans: We log these errors later, so there's no point in logging them here. It'd just spam the same error message multiple times in the logs.

pkg/uhttp/wrapper_test.go
```diff
@@ -385,7 +388,97 @@ func TestWrapper_NewRequest(t *testing.T) {
...
+	for _, tc := range test {
+		t.Run(tc.name, func(t *testing.T) {
+			u, err := url.Parse(tc.url)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			ctx = context.WithValue(ctx, ContextKey{}, tc.cc)
+			client, err := NewBaseHttpClient(ctx, http.DefaultClient)
+			assert.Nil(t, err)
```
ggreer, line 481: You can use require.NoError(t, err) and it'll print out a nicer message if there is an error.

mchavez: Fixed

pkg/uhttp/wrapper_test.go
```diff
@@ -21,9 +22,11 @@ type example struct {
 	Age  int    `json:"age"`
 }
 
+var ctx = context.Background()
+
 func TestWrapper_NewBaseHttpClient(t *testing.T) {
 	httpClient := http.DefaultClient
-	client := NewBaseHttpClient(httpClient)
+	client, _ := NewBaseHttpClient(ctx, httpClient)
```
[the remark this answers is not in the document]
mchavez: Fixed

pkg/uhttp/wrapper.go
```diff
@@ -32,17 +41,46 @@ type (
...
+	cache, err := NewGoCache(ctx, config)
+	if err != nil {
+		l.Error("in-memory cache error", zap.Any("NewBaseHttpClient", err))
+		return nil, err
+	}
+
+	return &BaseHttpClient{
+		HttpClient:    httpClient,
+		baseHttpCache: cache,
+	}, nil
```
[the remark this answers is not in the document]
mchavez: Fixed

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,137 @@
...
+	// Hash the cache string to create a key
+	hash := sha256.New()
+	_, err := hash.Write([]byte(cacheString))
+	if err != nil {
+		return "", err
+	}
+
+	cacheKey := fmt.Sprintf("%x", hash.Sum(nil))
+	return cacheKey, nil
+}
```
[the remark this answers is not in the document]
mchavez: Fixed

pkg/uhttp/gocache.go
```diff
@@ -0,0 +1,137 @@
...
+		resp, err := http.ReadResponse(r, nil)
+		if err != nil {
+			return resp, err
+		}
+
+		return resp, nil
+	}
+
+	return nil, nil
+}
```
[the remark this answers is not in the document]
mchavez: Fixed

vendor/github.com/allegro/bigcache/v3/iterator.go
```diff
@@ -0,0 +1,146 @@
...
+var emptyEntryInfo = EntryInfo{}
+
+// EntryInfo holds informations about entry in the cache
+type EntryInfo struct {
+	timestamp uint64
+	hash      uint64
+	key       string
+	value     []byte
+	err       error
+}
```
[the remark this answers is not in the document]
ggreer: This is vendored code that we can't change. Please ignore it.

ggreer (approved):

PR #196, opened 2024-07-30
Limit ticket schemas to 100

laurenleach:
## Summary by CodeRabbit

- **New Features**
	- Implemented a limit on the number of ticket schemas returned, enhancing efficiency and performance by ensuring only the first 100 schemas are processed.
  
- **Bug Fixes**
	- Improved control flow in the task handling to prevent excessive data accumulation, resulting in better usability.

### DIFF
--- a/pkg/tasks/c1api/list_ticket_schemas.go
+++ b/pkg/tasks/c1api/list_ticket_schemas.go
@@ -15,6 +15,8 @@ import (
 	"github.com/conductorone/baton-sdk/pkg/types"
 )
 
+const maxTicketSchemas = 100
+
 type listTicketSchemasTaskHelpers interface {
 	ConnectorClient() types.ConnectorClient
 	FinishTask(ctx context.Context, resp proto.Message, annos annotations.Annotations, err error) error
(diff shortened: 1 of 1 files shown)


ButternCream (approved):

pkg/tasks/c1api/list_ticket_schemas.go
```diff
@@ -48,9 +50,16 @@ func (c *listTicketSchemasTaskHandler) HandleTask(ctx context.Context) error {
 
 		ticketSchemas = append(ticketSchemas, schemas.GetList()...)
 
+		// Only return first 100 elements
+		if len(ticketSchemas) >= maxTicketSchemas {
+			ticketSchemas = ticketSchemas[:maxTicketSchemas]
```
ggreer, line 55: It'd be nice to log if we hit this condition, and if so, how many ticket schemas we ignored.

laurenleach: ill add a log, the number of ticket schemas we ignore could be more though if theres additional pages

PR #214, opened 2024-08-20
Upload debug file and delete it afterwards

shackra:
## Summary by CodeRabbit

- **New Features**
	- Introduced a debug log upload mechanism to enhance task handling and error tracking.
- **Bug Fixes**
	- Improved error handling during log uploads to ensure issues are correctly reported.

### DIFF
--- a/pkg/tasks/c1api/full_sync.go
+++ b/pkg/tasks/c1api/full_sync.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"io"
 	"os"
+	"path/filepath"
 
 	v1 "github.com/conductorone/baton-sdk/pb/c1/connectorapi/baton/v1"
 	"github.com/conductorone/baton-sdk/pkg/annotations"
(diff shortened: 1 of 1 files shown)


pkg/tasks/c1api/full_sync.go
```diff
@@ -140,3 +146,35 @@ func newFullSyncTaskHandler(task *v1.Task, helpers fullSyncHelpers, skipFullSync
...
+		}
+		defer func() {
+			err := os.Remove(debugfilelocation)
+			if err != nil {
+				l.Error("failed to delete file with debug logs", zap.Error(err), zap.String("file", debugfilelocation))
+			}
+		}()
+	} else if errors.Is(err, os.ErrNotExist) {
+		l.Warn("debug log file does not exists", zap.Error(err))
+	}
```
mgaeta, line 177: The way this is written it seems like if `os.Stat()` errors for some reason, we just return nil.
I think it might make sense to do something like.
```go
_, err := os.Stat(debugfilelocation)
if err != nil {
  if errors.Is(err, os.ErrNotExist) {
    l.Warn("debug log file does not exists", zap.Error(err))
    return nil
  }
  return err
}
```
I think if the intention is to fail silently, it would be helpful to leave a comment.

shackra: > The way this is written it seems like if os.Stat() errors for some reason, we just return nil.

I just thought that we should not care about any other error than `os.ErrNotExist`, and given how the top level is written, maybe returning the error from `os.Stat` won't break anything (who knows 🤷🏻‍♂️)

mgaeta (approved):

PR #218, opened 2024-08-23
Exit the for-loop after 5 gRPC communication errors

shackra:
## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for communication issues, improving the reliability of task processing.
	- Implemented a mechanism to control task processing termination based on specific error conditions, allowing for graceful exits.
- **Bug Fixes**
	- Improved handling of specific gRPC error codes to ensure appropriate logging and error management.

### DIFF
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -213,6 +213,11 @@ func (c *connectorRunner) run(ctx context.Context) error {
 			l.Debug("runner: dispatched task, waiting for next task", zap.Duration("wait_duration", waitDuration))
 		}
 	}
+
+	if commErrCount > 0 {
+		return fmt.Errorf("Unable to communicate with gRPC server")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/connectorrunner/runner.go
```diff
@@ -128,8 +130,9 @@ func (c *connectorRunner) run(ctx context.Context) error {
 
 	waitDuration := time.Second * 0
 	errCount := 0
+	commErrCount := 0
 	var err error
-	for {
+	for commErrCount < 5 {
```
ggreer, line 135: Do we want to reset this back to zero if we successfully handle a task?

pkg/connectorrunner/runner.go
```diff
@@ -194,6 +197,14 @@ func (c *connectorRunner) run(ctx context.Context) error {
 				defer sem.Release(1)
 				err := c.processTask(ctx, t)
 				if err != nil {
+					switch code := status.Code(err); code {
+					case codes.Canceled:
+						commErrCount++
+					case codes.Unknown:
```
ggreer, line 203: We wrap some http errors in grpc status Unknown: https://github.com/ConductorOne/baton-sdk/blob/d8b900d/pkg/uhttp/wrapper.go#L294

Maybe it makes sense to only increment the error count if it's a grpc client connection error.

shackra:
testing is needed, I'll report back when I have something that greenlight this PR for merging.

PR #238, opened 2024-10-07
Place pagination Bag in a good state when calling NextToken

shackra:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new method, `NextToken`, for improved pagination state management.
  
- **Bug Fixes**
	- Enhanced handling of empty page states to ensure valid token retrieval without errors.

- **Tests**
	- Added a new test to validate the behavior of `NextToken` with no active page state.
	- Updated existing tests for consistency with method naming conventions.

### DIFF
--- a/pkg/pagination/pagination.go
+++ b/pkg/pagination/pagination.go
@@ -93,6 +93,14 @@ func (pb *Bag) Next(pageToken string) error {
 
 // Next pops the current token, and pushes a copy of it with an updated page token.
 func (pb *Bag) NextToken(pageToken string) (string, error) {
+	// assume that `pb` was passed an empty token
+	if pb.currentState == nil {
+		pb.currentState = &PageState{
+			Token: pageToken,
+		}
+		return pb.Marshal()
+	}
+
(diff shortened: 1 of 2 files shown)


pquerna (approved):

pquerna (approved):

pkg/pagination/pagination.go
```diff
@@ -93,6 +93,14 @@ func (pb *Bag) Next(pageToken string) error {
 
 // Next pops the current token, and pushes a copy of it with an updated page token.
 func (pb *Bag) NextToken(pageToken string) (string, error) {
+	// assume that `pb` was passed an empty token
+	if pb.currentState == nil {
```
pquerna, line 97: if `pageToken` is `""` here, should we return `"", nil`?  eg, the nextToken is empty?

shackra: if `pageToken` is empty, the same state is returned per line 109. During my tests with baton-aws I didn't fall into infinite loops

PR #240, opened 2024-10-08
Add wrapper options to base http client, and add rate limiter option

jirwin:
This change allows you to specify a new `WithRateLimiter(num, period)` option when creating a new base http client.

`num`: the number of requests to be made during `period`
`period`: the duration to enforce the rate limit over

`WithRateLimiter(10, time*Minute)`: This would enforce a rate limit of 10 http requests per minute.

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -12,13 +12,16 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"time"
 
-	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
-	"github.com/conductorone/baton-sdk/pkg/ratelimit"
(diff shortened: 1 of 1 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -254,6 +289,12 @@ func (c *BaseHttpClient) Do(req *http.Request, options ...DoOption) (*http.Respo
 		resp     *http.Response
 	)
 	l := ctxzap.Extract(req.Context())
+
+	// If a rate limiter is defined, take a token before making the request.
+	if c.rateLimiter != nil {
+		c.rateLimiter.Take()
```
mgaeta, line 295: Does taking a token just block until one is available?

mgaeta (approved):

pkg/uhttp/wrapper.go
jirwin: Yes

PR #241, opened 2024-10-09
We should consider default fields in relationship constrains

btipling:
as is required for external ticketing which depends on ticketing a default field

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration handling by ensuring the schema for CLI commands contains unique fields, improving command execution and options presentation.
  
- **Bug Fixes**
	- Resolved potential issues with duplicate field names in the configuration schema.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -43,13 +43,24 @@ func DefineConfiguration(
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	v.AutomaticEnv()
 
+	confschema := schema
+	confschema.Fields = append(field.DefaultFields, confschema.Fields...)
+	// Ensure unique fields
+	uniqueFields := make(map[string]field.SchemaField)
+	for _, f := range confschema.Fields {
+		uniqueFields[f.FieldName] = f
+	}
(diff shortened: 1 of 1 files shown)


laurenleach (approved):

pkg/config/config.go
```diff
@@ -43,13 +43,24 @@ func DefineConfiguration(
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	v.AutomaticEnv()
 
+	confschema := schema
+	confschema.Fields = append(field.DefaultFields, confschema.Fields...)
+	// Ensure unique fields
+	uniqueFields := make(map[string]field.SchemaField)
+	for _, f := range confschema.Fields {
+		uniqueFields[f.FieldName] = f
```
pquerna, line 51: should this panic or otherwie fail?

btipling: It should panic these are schema definitions we wrote

btipling: It fails later when we attempt to add a duplicate flag actually

shackra: why should this panic tho?

pquerna: duplicate flags was my worry

PR #246, opened 2024-10-23
Add bulk create/get ticket tasks 

laurenleach:
## Release Notes

- **New Features**
  - Introduced bulk ticket creation and retrieval functionalities, allowing users to create and get multiple tickets in a single operation.
  - Added new request and response types for bulk operations, enhancing the API's capabilities.
  
- **Improvements**
  - Enhanced validation framework for bulk ticket operations to ensure data integrity and proper error handling.
  
- **Bug Fixes**
  - Updated existing ticket response messages to include error fields for better error reporting.

These changes significantly improve the efficiency and usability of ticket management within the application.


pkg/field/defaults.go
```diff
@@ -3,12 +3,14 @@ package field
...
-	getTicketField         = BoolField("get-ticket", WithHidden(true), WithDescription("Get ticket"), WithPersistent(true))
-	ListTicketSchemasField = BoolField("list-ticket-schemas", WithHidden(true), WithDescription("List ticket schemas"), WithPersistent(true))
-	provisioningField      = BoolField("provisioning", WithShortHand("p"), WithDescription("This must be set in order for provisioning actions to be enabled"), WithPersistent(true))
-	TicketingField         = BoolField("ticketing", WithDescription("This must be set to enable ticketing support"), WithPersistent(true))
-	c1zTmpDirField         = StringField("c1z-temp-dir", WithHidden(true), WithDescription("The directory to store temporary files in. It must exist, "+
+	createTicketField           = BoolField("create-ticket", WithHidden(true), WithDescription("Create ticket"), WithPersistent(true))
+	bulkCreateTicketField       = BoolField("bulk-create-ticket", WithHidden(true), WithDescription("Bulk create tickets"), WithPersistent(true))
+	bulkTicketTemplatePathField = StringField("bulk-ticket-template-path", WithHidden(true), WithDescription("A JSON file describing the ticket to create"), WithPersistent(true))
```
btipling, line 8: bulkTicketTemplatePathField why do we need this?

pkg/tasks/c1api/bulk_get_tickets.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	resp, err := cc.BulkGetTickets(ctx, &v2.TicketsServiceBulkGetTicketRequest{
+		TicketRequests: ticketRequests,
+	})
+	if err != nil {
+		return c.helpers.FinishTask(ctx, nil, nil, err)
+	}
+
+	if len(resp.GetTickets()) == 0 {
```
btipling, line 52: should this be a comparison to the length of `ticketRequests` ?

pkg/tasks/local/ticket.go
```diff
@@ -18,6 +18,126 @@ import (
...
+		})
+	}
+
+	bulkTicketReq := &v2.TicketsServiceBulkCreateTicketRequest{
+		TicketRequests: ticketReqs,
+	}
+
+	resp, err := cc.BulkCreateTickets(ctx, bulkTicketReq)
+	if err != nil {
+		return err
```
btipling, line 126: Hrm, do we want to attempt all the tickets and report a list of errors or one ticket fails they all fail? This would cause all ticketing to fail for zscaler because one config went out of date, and they have many. I don't think we should kill the full batch. All tickets would get stuck this way.

btipling: This does not indicate that we do fail for individual tickets. It is making a BulkCreateTickets call here.

btipling: And this is local

pkg/tasks/c1api/bulk_get_tickets.go
laurenleach: yea that makes more sense

pkg/connectorrunner/runner.go
```diff
@@ -584,6 +599,8 @@ func NewConnectorRunner(ctx context.Context, c types.ConnectorServer, opts ...Op
 			tm = local.NewListTicketSchema(ctx)
 		case cfg.getTicketConfig != nil:
 			tm = local.NewGetTicket(ctx, cfg.getTicketConfig.ticketID)
+		case cfg.bulkCreateTicketConfig != nil:
```
btipling, line 602: I think we had talked about maybe just making it so that ticketconfig could have multiple. It's not used much. Not a big deal though.

pkg/field/defaults.go
btipling: Mentioned in another comment that. maybe we could just repurpose the existing config path arg if you think that makes sense.

btipling (approved):

PR #249, opened 2024-10-30
Show the body regardless of what's the Content Type

shackra:
Show the body of the response, even if we expected it to be JSON, for error reporting and debugging purposes.

Example output:

```
{"level":"error","ts":1730327736.9045234,"caller":"connectorrunner/runner.go:182","msg":"runner: error processing on-demand task","error":"runner: error processing task: rpc error: code = NotFound desc = error: listing resources failed: rpc error: code = NotFound desc = 404 Not Found\nunexpected content type for JSON response: custom-mime/empty. body: « \" Not Available \" »","task_id":"","task_type":"sync_full"}
{"level":"error","ts":1730327736.9045494,"caller":"cli/commands.go:178","msg":"error running connector","error":"runner: error processing task: rpc error: code = NotFound desc = error: listing resources failed: rpc error: code = NotFound desc = 404 Not Found\nunexpected content type for JSON response: custom-mime/empty. body: « \" Not Available \" »"}
runner: error processing task: rpc error: code = NotFound desc = error: listing resources failed: rpc error: code = NotFound desc = 404 Not Found
unexpected content type for JSON response: custom-mime/empty. body: « " Not Available " »
```

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for non-JSON HTTP responses, providing clearer error messages that include response body content when applicable.

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -181,10 +181,6 @@ func NewBaseHttpClientWithContext(ctx context.Context, httpClient *http.Client, 
 func WithJSONResponse(response interface{}) DoOption {
 	return func(resp *WrapperResponse) error {
 		contentHeader := resp.Header.Get(ContentType)
-		if contentHeader == "" {
-			// for some reason there was no header for content-type
-			contentHeader = "custom-mime/empty"
-		}
 
 		if !IsJSONContentType(contentHeader) {
 			if len(resp.Body) != 0 {
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -180,8 +180,18 @@ func NewBaseHttpClientWithContext(ctx context.Context, httpClient *http.Client, 
 // status code 204 No Content), then pass a `nil` to `response`.
 func WithJSONResponse(response interface{}) DoOption {
 	return func(resp *WrapperResponse) error {
-		if !IsJSONContentType(resp.Header.Get(ContentType)) {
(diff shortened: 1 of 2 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -180,8 +180,18 @@ func NewBaseHttpClientWithContext(ctx context.Context, httpClient *http.Client,
 // status code 204 No Content), then pass a `nil` to `response`.
 func WithJSONResponse(response interface{}) DoOption {
 	return func(resp *WrapperResponse) error {
-		if !IsJSONContentType(resp.Header.Get(ContentType)) {
-			return fmt.Errorf("unexpected content type for json response: %s", resp.Header.Get(ContentType))
+		contentHeader := resp.Header.Get(ContentType)
+		if contentHeader == "" {
```
ggreer, line 184: An empty string will cause IsJSONContentType() to return false, so this if statement isn't needed.

shackra: but what if it is something valid like `text/plain`?

ggreer: IsJSONContentType() only returns true if content type starts with "application" and contains "json". If it's "text/plain", then we'll error and log the body.

Though I do notice that IsJSONContentType() should probably normalize the content type string like IsXMLContentType() does.

shackra: we prefer the value to remain empty then, got it.

ggreer (approved):

PR #254, opened 2024-11-18
Add tracing support via OpenTelemetry

jirwin:
Adds tracing support. SDK instruments the root process, and propagates trace context to the child process. The connector implementation can trace functions as well. grpc calls are traced automatically.

The root process will be named after the connector, e.g. `baton-github`. The child process service name will have `-server` appended to the service name.

This also adds a open telemetry stack via docker-compose. This runs an otel collector, prometheus, loki, tempo, and grafana. Currently only tracing is implemented, so prometheus and loki are going unused, but it won't be difficult to add support for those.

## TODO
- [x] Add readme for docker-compose stack
- [ ] Figure out why batch processor causes sigkills
- [x] Test service mode

image


loganintech:
A lot of the tracers are prefixed with `baton-sdk/` how would you feel about adding a version number too?

pkg/cli/commands.go
```diff
@@ -55,6 +57,17 @@ func MakeMainCommand(
 
 		l := ctxzap.Extract(runCtx)
 
+		otelShutdown, err := uotel.InitOtel(context.Background(), v.GetString("otel-collector-endpoint"), name)
```
loganintech, line 60: Any reason this doesn't use the runCtx?

pkg/cli/commands.go
```diff
@@ -210,6 +223,22 @@ func MakeGRPCServerCommand(
 			return err
 		}
 
+		l := ctxzap.Extract(runCtx)
+
+		otelShutdown, err := uotel.InitOtel(
+			context.Background(),
```
loganintech, line 229: same question re: runCtx

loganintech (approved):
Small questions, spans look good though

pkg/cli/commands.go
jirwin: Same answer for both cases. If `runCtx` gets cancelled, we want to be sure that otel can still flush its data before the process is shutdown.

ggreer:
Just FYI I rebased this and fixed conflicts. Hopefully I didn't mess anything up.

PR #255, opened 2024-11-22
Expand grants faster.

ggreer:
Improve syncing and grant expansion performance.

- Add more expandable grants to the syncer benchmark (which aren't cycles, just group->group->user). Increase the cycle depth.
- Fix behavior of mock connector in syncer benchmark. Instead of Entitlements() returning all entitlements, we return the entitlements for that specific resource. Same for Grants().
- Never checkpoint more often than once every 10 seconds. Previously we were checkpointing after every action was completed, which hurt performance since we were serializing the syncer state and saving it.
- Use shorter JSON tags so that we're serializing less data every time we checkpoint.
- Optimize GetFirstCycle() and FixCycles() to not iterate over every edge every time. If we called FixCycles() previously and the graph hasn't been modified, then we know there are still no cycles.

After these changes, we spend almost all of our time saving grants instead of checkpointing.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `HasNoCycles` property to streamline cycle detection in the entitlement graph.
	- Enhanced checkpointing mechanism with a minimum interval for creating checkpoints.

- **Bug Fixes**
	- Improved error handling in resource listing methods.

- **Refactor**
	- Updated JSON serialization tags for better efficiency in the entitlement graph structures.
	- Enhanced test code organization and resource management in the mock connector.
	- Modified the `LoadC1Z` method to include additional parameters for improved functionality.

- **Tests**
	- Added new methods for managing groups and users in the mock connector for improved benchmarking.

### DIFF
--- a/pkg/dotc1z/manager/local/local.go
+++ b/pkg/dotc1z/manager/local/local.go
@@ -84,7 +84,7 @@ func (l *localManager) LoadC1Z(ctx context.Context) (*dotc1z.C1File, error) {
 		zap.String("temp_path", l.tmpPath),
 	)
 
-	return dotc1z.NewC1ZFile(ctx, l.tmpPath, dotc1z.WithTmpDir(l.tmpDir))
+	return dotc1z.NewC1ZFile(ctx, l.tmpPath, dotc1z.WithTmpDir(l.tmpDir), dotc1z.WithPragma("journal_mode", "WAL"))
 }
 
 // SaveC1Z saves the C1Z file to the local file system.
--- a/pkg/sync/expand/graph.go
+++ b/pkg/sync/expand/graph.go
@@ -8,27 +8,29 @@ import (
 	"go.uber.org/zap"
 )
 
+// JSON tags for actions, edges, and nodes are short to minimize size of serialized data when checkpointing
+
 type EntitlementGraphAction struct {
-	SourceEntitlementID     string   `json:"source_entitlement_id"`
-	DescendantEntitlementID string   `json:"descendant_entitlement_id"`
-	Shallow                 bool     `json:"shallow"`
-	ResourceTypeIDs         []string `json:"resource_types_ids"`
-	PageToken               string   `json:"page_token"`
+	SourceEntitlementID     string   `json:"sid"`
+	DescendantEntitlementID string   `json:"did"`
+	Shallow                 bool     `json:"s"`
+	ResourceTypeIDs         []string `json:"rtids"`
+	PageToken               string   `json:"pt"`
 }
 
 type Edge struct {
-	EdgeID          int      `json:"edge_id"`
-	SourceID        int      `json:"source_id"`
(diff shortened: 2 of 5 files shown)


pkg/dotc1z/manager/local/local.go
```diff
@@ -84,7 +84,7 @@ func (l *localManager) LoadC1Z(ctx context.Context) (*dotc1z.C1File, error) {
 		zap.String("temp_path", l.tmpPath),
 	)
 
-	return dotc1z.NewC1ZFile(ctx, l.tmpPath, dotc1z.WithTmpDir(l.tmpDir))
+	return dotc1z.NewC1ZFile(ctx, l.tmpPath, dotc1z.WithTmpDir(l.tmpDir), dotc1z.WithPragma("journal_mode", "WAL"))
```
pquerna, line 87: hmm, are there any side effects with this?  When we re-compress the sqlite data file, wouldn't we also need the WAL file(s)?

pkg/sync/expand/graph.go
```diff
@@ -8,27 +8,29 @@ import (
...
-	PageToken               string   `json:"page_token"`
+	SourceEntitlementID     string   `json:"sid"`
+	DescendantEntitlementID string   `json:"did"`
+	Shallow                 bool     `json:"s"`
+	ResourceTypeIDs         []string `json:"rtids"`
+	PageToken               string   `json:"pt"`
 }
 
 type Edge struct {
-	EdgeID          int      `json:"edge_id"`
```
pquerna, line 20: since we used these previously as a effective pagination token, does this mean ideally for in-place upgades we throw away "old" pagination state, and start over?  should "version" the pagination token?

pkg/dotc1z/manager/local/local.go
ggreer: The WAL file is deleted when the last DB connection is closed. We always close the DB before calling saveC1z(), so the WAL file doesn't exist when we compress. If I comment out the calls to os.Remove() and os.RemoveAll(), the only file in the temp dir is the DB.

```
ggreer@gmac:/var/folders/by/00zvpw316q50n2r1jlwqkqgr0000gn/T/c1z2968114110% ls -la
total 18880
drwx------    3 ggreer  staff       96 Nov 22 11:16 ./
drwx------@ 570 ggreer  staff    18240 Nov 22 11:16 ../
-rw-r--r--    1 ggreer  staff  9666560 Nov 22 11:16 db
```

pkg/sync/expand/graph.go
ggreer: I'm sure we've broken this before. In an ideal world we'd version this stuff and write migrations, but I think the moneyball solution would be to end the sync if we error unmarshaling the json. I pushed a change that does this.

pquerna: I don;'t think we need a migration, we just need to start the specified sync step over from zero

PR #262, opened 2024-12-11
Multi-Recipient encryption for client credentials

mj-palanker:
Modifies crypto and jwk encryption provider to encrypt the same secret with multiple keys in the multi-recipient setup. Allows us to only return a single copy of the encrypted data when asked to encrypt it for many keys

Open Question, for a single key encryption we store a hash of the public key under the "keyId" field of the returned data... what should we do for a multi-key encryption? should we hash all keys together somehow? store a list of hashes? How are we even using this field in c1/baton?

## Summary by CodeRabbit

- **New Features**
  - Enhanced encryption management with support for multiple configurations.
  - New functionality to create recipients for ED25519 and RSA keys.
  
- **Bug Fixes**
  - Improved error handling during encryption and recipient creation processes.

- **Tests**
  - Added tests for multi-recipient encryption and updated existing tests for multiple configurations.

- **Documentation**
  - Updated comments and documentation to reflect changes in method signatures and functionality.


pkg/crypto/providers/jwk/jwk.go
```diff
@@ -72,48 +72,71 @@ func (j *JWKEncryptionProvider) GenerateKey(ctx context.Context) (*v2.Encryption
...
+	if len(recipients) > 0 {
+		ciphertext, err := ageEncrypt(recipients, plainText.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("jwk: %w", err)
+		}
+		encCipherText := base64.StdEncoding.EncodeToString(ciphertext)
+
+		encrypted = append(encrypted, &v2.EncryptedData{
+			Provider:       EncryptionProviderJwk,
+			KeyId:          "", // MJP what does "keyID" mean for multi-recipient encryption? Multiple ids?
```
ggreer, line 132: I'm pretty sure you want to append to encrypted in the loop and set KeyId to config.KeyId. That way cone or whatever can look through the response, find the key ID that it knows, and decrypt that one to provide credentials to the user.

mj-palanker: Thanks for taking a look Geoff, yeah ecdsa keys aren't supported by the multi-recipient age library we're using (NOTE TO SELF, double check that) so we handle them like we used to. For everything else though, we can actually create just a single EncryptedData package which any key can unlock, this way we don't create (secret * keys) many encrypted blocks, but instead just one that shares keys. My question here is how I should identify said package via KeyID. 
Cone currently doesn't actually filter out based on KeyID (yet, I'm actually working on that next) so we can decide now what form a "multi-key KeyID" should take for our purposes...

A trivial answer could be something like all the KeyIDs we would have, just with some delimiter. meaning we'd have to just parse them out later.

e.g.  "d0cfc64";"79ee7b51b03813b408c";"82f0a1832df2b3ec179f3b";"b47c6b3c9071b724" or what have you.

That's probably my preferred route as opposed to some weird hash of all keys (which limits our ability to ever get a matching hash again) or making EncryptedData store a list of strings which might be more "correct" actually... 🤔

pkg/crypto/providers/jwk/jwk.go
```diff
@@ -72,48 +72,73 @@ func (j *JWKEncryptionProvider) GenerateKey(ctx context.Context) (*v2.Encryption
...
+	if len(recipients) > 0 {
+		ciphertext, err := ageEncrypt(recipients, plainText.Bytes)
+		if err != nil {
+			return nil, fmt.Errorf("jwk: %w", err)
+		}
+		encCipherText := base64.StdEncoding.EncodeToString(ciphertext)
+
+		encrypted = append(encrypted, &v2.EncryptedData{
+			Provider:       EncryptionProviderJwk,
+			KeyId:          multiRecipientThumbs,
```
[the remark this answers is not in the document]
mj-palanker: Thank you robot 🤖

PR #271, opened 2024-12-18
Adds AI agent user type

santhosh-c1:
## Summary by CodeRabbit

- **New Features**
	- Added support for a new account type, `ACCOUNT_TYPE_AI_AGENT`, in the user traits.
  
- **Bug Fixes**
	- Enhanced test coverage for the `UserTrait` functionality, ensuring correct population of fields for AI agent accounts.
  
- **Chores**
	- Updated `.gitignore` to include the `.idea/` directory and reorganized comment sections.


.gitignore
```diff
@@ -18,3 +18,6 @@ dist/
 # Dependency directories (remove the comment below to include it)
 # vendor/
 /.pre-commit-config.yaml
+
+# Others
```
GowthamiDommety, line 22: do we need changes to this file?

ggreer: We do not. IDE-specific ignores should probably go in that person's global gitignore.

proto/c1/connector/v2/annotation_trait.proto
```diff
@@ -40,6 +40,7 @@ message UserTrait {
     ACCOUNT_TYPE_HUMAN = 1;
     ACCOUNT_TYPE_SERVICE = 2;
     ACCOUNT_TYPE_SYSTEM = 3;
+    ACCOUNT_TYPE_AI_AGENT = 4;
```
ggreer, line 43: What's the difference between this and a service account, and how would we know in an API? Is it a separate call? Is it an attribute?

It seems like this would be a trait option on a user, not necessarily its own user type.

ggreer:
After talking to @GowthamiDommety, I think this should be a general resource trait with its own trait options, not a user trait option. Some of the things we want to flag as non human identities are users, but in other cases they're some other type of resource like a credential. If this is a user trait option, we'd have to shoehorn those resources into the user model.

santhosh-c1:
Closing for now. Deprioritized.

PR #273, opened 2024-12-18
Introduce key_ids as eventual replacement for key_id in EncryptedData

mj-palanker:
to make multi-recipient encryption rollout easier, lets introduce the proto change now, so consumers like cone can be made ready for the eventual move to multiple keys per Encrypted data package.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the encryption output to include multiple key identifiers for better key management.
	- Added a new field to the encrypted data structure for storing multiple key IDs.

- **Bug Fixes**
	- Improved the flexibility of the key management system by modifying the data structure without affecting existing functionalities.


loganintech (approved):

jirwin (approved):

proto/c1/connector/v2/resource.proto
```diff
@@ -175,11 +175,12 @@ message CreateAccountResponse {
 
 message EncryptedData {
   string provider = 1;
-  string key_id = 2;
+  string key_id = 2 [deprecated = true];
   string name = 3;
   string description = 4;
   string schema = 5; // optional
   bytes encrypted_bytes = 6; // if 'schema' is set, this should be JSON.
```
ggreer, line 182: CreateAccountResponse contains a repeated EncryptedData, so why not just encrypt the data multiple times, once for each key id? That we can have the same data decryptable by multiple keys is an artifact of certain encryption schemes, not something that can be generalized.

PR #276, opened 2024-12-28
Print the response's body on demand for debugging purposes

shackra:
This PR adds debugging functionality that displays the response's body as it arrives to us, not immediately, but when reading it.

Example:

```shell
# assume baton-random-connector is a connector that expects a JSON response
BATON_DEBUG_PRINT_RESPONSE_BODY=yes baton-random-connector
# * prints out the response body when read *
```

The `uhttp.WithJSONResponse(&target)` option on `.Do` would trigger a read, and we would see what was read on the console.

## Summary by CodeRabbit

- **New Features**
  - Introduced a debugging option to log HTTP response bodies.
- **Improvements**
  - Enhanced HTTP client with configurable response body logging via environment variable.
  - Provides a flexible way to enable/disable response body debugging.
- **Debugging**
  - New capability to inspect HTTP response contents during development.
- **Documentation**
  - Updated linter settings to improve error handling practices.
  - Added requirements for explanations on `nolint` directives.
- **Code Organization**
  - Simplified visibility of the `ErrSyncNotComplete` variable.
  - Updated comments for clarity in the `SyncGrantExpansion` method.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -36,6 +36,7 @@ linters-settings:
         arguments:
           - fmt.Printf
           - fmt.Println
+          - fmt.Fprint
           - fmt.Fprintf
           - fmt.Fprintln
           - os.Stderr.Sync
--- a/pkg/uhttp/body_print.go
+++ b/pkg/uhttp/body_print.go
@@ -2,9 +2,6 @@ package uhttp
 
 // Implements a debugging facility for request responses. This changes
 // the behavior of `BaseHttpClient` with an unexported flag.
-//
-// If you always wanted to see the actual body of your response, now
-// you can 👁️👄👁️undress it🫦 to uncover all its... data!
 
 import (
 	"io"
@@ -24,18 +21,18 @@ func (pr *printReader) Read(p []byte) (int, error) {
 	return n, err
 }
 
-func wrapBodyToUndress(body io.Reader) io.Reader {
+func wrapPrintBody(body io.Reader) io.Reader {
 	return &printReader{reader: body}
 }
 
-type undressOption struct {
-	undress bool
+type printBodyOption struct {
+	debugPrintBody bool
 }
 
-func (o undressOption) Apply(c *BaseHttpClient) {
-	c.debugPrintBody = o.undress
+func (o printBodyOption) Apply(c *BaseHttpClient) {
+	c.debugPrintBody = o.debugPrintBody
 }
 
-func WithUndressBody(undress bool) WrapperOption {
-	return undressOption{undress: undress}
+func WithPrintBody(shouldPrint bool) WrapperOption {
+	return printBodyOption{debugPrintBody: shouldPrint}
 }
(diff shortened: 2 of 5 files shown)


pkg/uhttp/body_print.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (pr *printReader) Read(p []byte) (int, error) {
+	n, err := pr.reader.Read(p)
+	if n > 0 {
+		log.Print(string(p[:n]))
+	}
+
+	return n, err
+}
```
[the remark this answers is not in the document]
shackra: fair assumption, but this is not meant to go into production, once you get what you want during development, you should take it out.

pkg/uhttp/body_print.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type printReader struct {
+	reader io.Reader
+}
+
+func (pr *printReader) Read(p []byte) (int, error) {
+	n, err := pr.reader.Read(p)
+	if n > 0 {
+		log.Print(string(p[:n]))
```
jirwin, line 21: We shouldn't use `log.Print` here. If you want to write to stdout, you could use `fmt.Fprintf(os.Stdout, string(p[:n]))`

pkg/uhttp/body_print.go
```diff
@@ -0,0 +1,41 @@
...
+
+type undressOption struct {
+	undress bool
+}
+
+func (o undressOption) Apply(c *BaseHttpClient) {
+	c.debugPrintBody = o.undress
+}
+
+func WithUndressBody(undress bool) WrapperOption {
```
jirwin, line 39: Instead of a `WrapperOption`, which could accidentally be left behind, what if we checked for an environment variable, or added a global debug request body config option on the SDK(it could be hidden).

pkg/uhttp/body_print.go
```diff
@@ -0,0 +1,41 @@
+package uhttp
+
+// Implements a debugging facility for request responses. This changes
+// the behavior of `BaseHttpClient` with an unexported flag.
+//
+// If you always wanted to see the actual body of your response, now
+// you can 👁️👄👁️undress it🫦 to uncover all its... data!
```
btipling, line 7: Maybe the emoji stay contained to the PR?

shackra:
it is a prank PR and I'll remove them eventually :)

--
Jorge Araya
Contacto:
Telegram: t.me/shackra · Signal: Shackra.28

Enviado desde Proton Mail para Android

-------- Mensaje original --------

pkg/uhttp/body_print.go
shackra: I changed it to read an envvar named `BATON_DEBUG_PRINT_RESPONSE_BODY`, if it is non-empty it will trigger the same functionality as it was before with the functional option

.golangci.yml
```diff
@@ -36,6 +36,7 @@ linters-settings:
         arguments:
           - fmt.Printf
           - fmt.Println
+          - fmt.Fprint
```
ggreer, line 39: If we want this change here, we probably want it in [baton-template](https://github.com/ConductorOne/baton-template).

ggreer (approved):

PR #277, opened 2024-12-30
Add generic pagination helper

MarcusGoldschmidt:
Add Generic pagination helper

Helper to avoid explicit conversion on strings in scenarios where the token or id needs to be converted to another type, also can hold any kind of data.

We can think a better name instead of `GenBag`

## Summary by CodeRabbit

- **New Features**
	- Introduced a generic pagination bag implementation in Go.
	- Added methods for managing pagination states, including push, pop, and current state retrieval.
	- Implemented marshaling and unmarshaling functionality for pagination states from tokens, with error handling for invalid input.

- **Tests**
	- Added comprehensive test coverage for pagination bag functionality.
	- Verified marshaling, unmarshaling, and state management behaviors, ensuring correct handling of various input scenarios.

### DIFF
--- a/pkg/pagination/generic_bag.go
+++ b/pkg/pagination/generic_bag.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 )
 
+var ErrNoToken = fmt.Errorf("pagination token cannot be nil")
+
 type genericSerializedPaginationBag[T any] struct {
 	States       []T `json:"states"`
 	CurrentState *T  `json:"current_state"`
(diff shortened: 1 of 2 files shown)


ggreer:
If we're going with our normal naming pattern, I guess we'd call it BagV2?

I think GenBag or GenericBag is fine too. It'd be nice to see an example of it being used before we merge it and tag a release.

ggreer (approved):

PR #278, opened 2024-12-31
Make expanded grants immutable.

ggreer:
If we expand a grant and there isn't already a direct grant for the same entitlement and principal, mark the grant as immutable.

Also add tests.

## Summary by CodeRabbit

- **New Features**
	- Enhanced grant creation with immutability support.
	- Added test coverage for grant immutability scenarios.

- **Tests**
	- Introduced new tests to validate grant immutability during group membership expansion, including cyclic group memberships.

### DIFF
--- a/pkg/sync/syncer_test.go
+++ b/pkg/sync/syncer_test.go
@@ -9,7 +9,9 @@ import (
 	"testing"
 
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
+	reader_v2 "github.com/conductorone/baton-sdk/pb/c1/reader/v2"
 	"github.com/conductorone/baton-sdk/pkg/annotations"
+	"github.com/conductorone/baton-sdk/pkg/dotc1z/manager"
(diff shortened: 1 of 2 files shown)


pkg/sync/syncer_test.go
```diff
@@ -81,6 +81,107 @@ func TestExpandGrants(t *testing.T) {
 	_ = os.Remove(c1zpath)
 }
 
+func TestExpandGrantImmutable(t *testing.T) {
```
pquerna, line 84: is it possible to add a test for a circular group expansion?  i think this current test is a fine validation, but i was wondering about the behavoir of full circles, eg group A has member B which has member group C which has member A -> and a user could be attached at any point, but only have direct in a single group?

btipling (approved):

pkg/sync/syncer_test.go
ggreer: I added a test but it looks like we'll need more changes to support immutable in circular group membership.

PR #286, opened 2025-01-15
Adds Okta rate limit headers

santhosh-c1:
- This is a non-standard header that Okta uses 
- https://developer.okta.com/docs/reference/rl-best-practices/

## Summary by CodeRabbit

- **New Features**
	- Enhanced rate limit header support with additional headers for improved API compatibility, specifically for Okta integration.

### DIFF
--- a/pkg/ratelimit/http.go
+++ b/pkg/ratelimit/http.go
@@ -13,18 +13,21 @@ var limitHeaders = []string{
 	"X-Ratelimit-Limit",
 	"Ratelimit-Limit",
 	"X-RateLimit-Requests-Limit", // Linear uses a non-standard header
+	"X-Rate-Limit-Limit",         // Okta uses a non-standard header
 }
 
 var remainingHeaders = []string{
 	"X-Ratelimit-Remaining",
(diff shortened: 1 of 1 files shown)


magizh-c1 (approved):

pkg/ratelimit/http.go
```diff
@@ -13,18 +13,21 @@ var limitHeaders = []string{
 	"X-Ratelimit-Limit",
 	"Ratelimit-Limit",
 	"X-RateLimit-Requests-Limit", // Linear uses a non-standard header
```
pquerna, line 15: its weird to me we can't handle this via option parameters to uhttp or somehting.  centralizing this is... well, gonna be hard, and future error prone.

ggreer: We'll get 'em all eventually. Just one more http header and we'll have all of them covered. For sure this time. 😅

I agree that this setup is not ideal, but so far we only have two APIs that are wrong (Linear and Okta). If we have a 3rd one, then I think it's time to tweak this.

ggreer (approved):

PR #288, opened 2025-01-15
Reduce cancel time to fix flaky test.

ggreer:
## Summary by CodeRabbit

- **Tests**
	- Updated timeout duration in a test case from milliseconds to microseconds.


shackra (approved):
oops

PR #289, opened 2025-01-15
Fix flaky test.

ggreer:
- Don't try to release a semaphore if acquire errors, as the error might be context deadline exceeded in a test.
- Squelch error closing serverStdin if the error is that it's already closed.
- Make timeout in passingvalues_test 0 so that context is always deadline exceeded, preventing flakes and preventing entrypoint() from running forever.

## Summary by CodeRabbit

- **Bug Fixes**
	- Refined error handling in connector and runner components to better manage context cancellation and input stream closure scenarios.

- **Tests**
	- Updated test configurations to modify timeout behaviors in test contexts, ensuring immediate expiration.
	- Removed specific timeout constant in test files.

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/connector/connector.go
```diff
@@ -361,7 +362,7 @@ func (cw *wrapper) Close() error {
 
 	if cw.serverStdin != nil {
 		err = cw.serverStdin.Close()
-		if err != nil {
+		if err != nil && !strings.Contains(err.Error(), "already closed") {
```
pquerna, line 365: `errors.Is(err, os.ErrClosed)`?

ggreer: Ah yes, that's a better check. I'll push a fix.

PR #306, opened 2025-02-25
Fixes on ggreer/baton-actions

santhosh-c1:
- Also moves action manager to connector level, rather than resource level


pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -381,13 +388,13 @@ func NewConnector(ctx context.Context, in interface{}, opts ...Opt) (types.Conne
...
 				ret.credentialManagers[rType.Id] = credentialManagers
 			}
-
-			if actionManager, ok := rb.(ActionManager); ok {
-				if ret.actionManager != nil {
-					return nil, fmt.Errorf("error: duplicate resource type found for action manager %s", rType.Id)
-				}
-				ret.actionManager = actionManager
-			}
+			// This should be on the connector level?
```
santhosh-c1, line 391: @ggreer

ggreer: Pretty much all the actions I was playing with were on the user resource so I shoved the action manager implementation on that to reduce duplicated code.

PR #309, opened 2025-03-04
Add expiration duration to EncryptConfig and EncryptData

bemafisher1:
Added expiration_duration fields to EncryptData and EncryptConfig so that we can set expiration duration for app user credentials when provisioning app user.

## Summary by CodeRabbit

- **New Features**
	- Introduced expiration duration fields for encrypted data and encryption configuration, allowing specification of a validity period for credentials and encrypted content.
	- Enhanced validation logic to enforce the correct expiration duration, ensuring improved reliability and security in the encryption process.


proto/c1/connector/v2/resource.proto
```diff
@@ -181,6 +182,7 @@ message EncryptedData {
   string schema = 5; // optional
   bytes encrypted_bytes = 6; // if 'schema' is set, this should be JSON.
   repeated string key_ids = 7;
+  google.protobuf.Duration expiration_duration = 8;
```
mj-palanker, line 185: Don't love the name expiration_duration. it's like.... technically true, but if anything it's a lifetime duration? idk

bemafisher1: We can do `lifetime_duration`, `retention_duration` , `time_to_live`, or `validity_duration`. Which one sounds better?

mj-palanker: On second thought I think it's fine

mj-palanker: oops just saw your comment. I prefer time_to_live honestly
\

bemafisher1: Yeah, `time_to_live` is very explanatory

bemafisher1: @mstanbCO @mj-palanker updated

mstanbCO (approved):

mj-palanker (approved):
🎉

proto/c1/connector/v2/resource.proto
```diff
@@ -202,6 +204,8 @@ message EncryptionConfig {
   oneof config {
     JWKPublicKeyConfig jwk_public_key_config = 100;
   }
+
+  google.protobuf.Duration lifetime_duration = 4; // lifetime duration for encrypted data
```
ggreer, line 208: Can different credentials have different durations? eg: A temp password versus rotating credentials for an active account?

mj-palanker: Yeah the idea is to make this requested lifetime come from the vault / cred storage we're storing it in (possibly also the source e.g. rotation/creation/other), but we should respect the connector lifetime that it returns. I have no knowledge of a plan to bring overriding that to the connector itself right now, but that could always be on the table some day.

mj-palanker: From our perspective it's just what lifetime we're going to enforce when it's returned as encrypted data.

ggreer: OK, but only the account info and credential options are passed to CreateAccount(), and only credential options and the resource ID are passed to RotateCredentials(). So connector implementations don't know how long the credentials are supposed to last for when setting them, nor can connectors return info about how long the credentials will last.

If we want to enforce this, it needs to go in CredentialOptions, not EncryptionConfig. Also we'll want an annotation that CreateAccount() and RotateCredentials() can return that specifies the credential lifetime duration.

mj-palanker: The problem w/ putting it into credential options is we then have to have the same lifetime for all copies of a credential for the same action. We want the different copies of the credential to have different lifetimes, and c1 to manage those lengths. Having the conductor itself provide it's own lifetime is not something we plan on doing yet, but if we do perhaps that should be a return from `CreateAccount` or `RotateCredentials` which the sdk can prioritize using if set?

mj-palanker: Basically we want the lifetime scoped to the copy of the credential, just like the key we use to the encrypt. While a connector sourced lifetime (if we ever do support that) would effect all copies of the credential. so we could compare for each and take the shortest lifetime for each copy?

bemafisher1: Just a side note: both CreateAccountRequest and RotateCredentialRequest do contain encryption configs unless I am looking in a wrong place https://github.com/ConductorOne/baton-sdk/blob/eebc3bd0c1a1795f778130d6a5f778784297f6f3/pkg/connectorbuilder/connectorbuilder.go#L936

ggreer: ConnectorBuilder has EncryptionConfigs, but the AccountManager interface that is implemented by connectors doesn't: https://github.com/ConductorOne/baton-sdk/blob/main/pkg/connectorbuilder/connectorbuilder.go#L60 We didn't want connector authors to have to deal with encrypting stuff, which is why encryption is handled by ConnectorBuilder.

Is this lifetime just how long C1 keeps the encrypted credentials for, not how long the credentials are good for? If connectors won't change their behavior based on this duration (eg: by setting a password expiration time), and the returned duration is always the same as what's sent in, why have it in baton-sdk?

PR #310, opened 2025-03-04
Add support for using external baton resources

laurenleach:
## Summary by CodeRabbit

- **New Features**
  - Enhanced configuration of external resources with support for custom file paths and entitlement filters, including improved error handling when a resource file is missing.
  - Introduced robust validation for resource-related inputs to ensure proper configuration.
  - Expanded synchronization capabilities to manage grants, entitlements, and user/group associations more effectively.
  - Added support for grant deletion and new standardized definitions for external resource types.
  - Introduced new methods for syncing external resources and managing their configurations.
  - Added new relationships and fields for external resource management in the system.

- **Bug Fixes**
  - Improved error handling for loading external resource files.


pkg/sync/syncer.go
```diff
@@ -1801,6 +2353,18 @@ func WithSkipFullSync() SyncOpt {
 	}
 }
 
+func WithExternalResourceC1ZPath(path string) SyncOpt {
+	return func(s *syncer) {
+		s.externalResourceC1ZPath = path
+	}
+}
+
+func WithExternalResourceEntitlementIdFilter(entitlementId string) SyncOpt {
```
kans, line 2362: What if you care about two entitlements?

pkg/sync/syncer.go
```diff
@@ -1425,6 +1445,538 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+				newGrant := &v2.Grant{
+					Entitlement: grant.Entitlement,
+					Principal:   principal,
+					Id:          batonGrant.NewGrantID(principal, grant.Entitlement),
+					Sources:     grant.Sources,
+					Annotations: grant.Annotations,
+				}
+				expandedGrants = append(expandedGrants, newGrant)
+			}
+			err = s.store.DeleteGrant(ctx, grant.Id)
```
ggreer, line 1845: We're deleting the original grant before we're adding the new ones, so if the syncer dies at just the wrong time, we won't know about the original grants when we resume. I think we could collect the grants to delete and do that after the call to PutGrants(...).

proto/c1/connector/v2/annotation_baton_id.proto
```diff
@@ -0,0 +1,23 @@
...
+option go_package = "github.com/conductorone/baton-sdk/pb/c1/connector/v2";
+
+enum ExternalResourceType {
+  EXTERNAL_RESOURCE_TYPE_UNSPECIFIED = 0;
+  EXTERNAL_RESOURCE_TYPE_USER = 1;
+  EXTERNAL_RESOURCE_TYPE_GROUP = 2;
+}
+
+message ExternalResourceMatch {
+  ExternalResourceType resource_type = 1; // TODO(lauren) Use enum? or resource_type_id
```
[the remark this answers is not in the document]
laurenleach: I was actually able to use the trait enum

pkg/sync/syncer.go
laurenleach: like to include users/groups that have grants for any of these entitlements?

pkg/sync/syncer.go
```diff
@@ -1452,6 +1472,532 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+	if err != nil {
+		return err
+	}
+
+	err = s.store.PutGrants(ctx, grantsForEnts...)
+	if err != nil {
+		return err
+	}
+
+	l.Info("Synced external resources",
```
ggreer, line 1594: It'd be nice to log the external entitlement ID here too.
```suggestion
	l.Info("Synced external resources for entitlement",
	zap.String("external_entitlement_id", entitlementId),
```

pkg/sync/syncer.go
```diff
@@ -1452,6 +1472,532 @@ func (s *syncer) syncGrantsForResource(ctx context.Context, resourceID *v2.Resou
...
+		zap.Int("resource-types", len(userAndGroupResourceTypes)),
```
ggreer, line 1698: These are counts so it'd be nice if the zap field names were "resource_type_count", "resource_count", etc. The same goes for the l.Info() in SyncExternalResources().

ggreer (approved):

PR #327, opened 2025-03-13
Log Slow Queries 

kans:
## Summary by CodeRabbit

- **New Features**
	- Enhanced performance monitoring with improved detection and logging of slow queries.
	- Slow queries exceeding a set threshold now trigger warning notifications at controlled intervals, aiding in diagnosing performance issues without overwhelming logs.
	- Introduced new fields for tracking slow queries, including thresholds and logging frequency.
	- Added a method to log warnings for slow queries at a rate of once per minute per query type, ensuring efficient logging.

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
+	"time"
 
 	"github.com/doug-martin/goqu/v9"
 	// NOTE: required to register the dialect for goqu.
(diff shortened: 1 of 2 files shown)


pkg/dotc1z/c1file.go
```diff
@@ -67,9 +75,13 @@ func NewC1File(ctx context.Context, dbFilePath string, opts ...C1FOption) (*C1Fi
...
 	c1File := &C1File{
-		rawDb:      rawDB,
-		db:         db,
-		dbFilePath: dbFilePath,
+		rawDb:                 rawDB,
+		db:                    db,
+		dbFilePath:            dbFilePath,
+		pragmas:               []pragma{},
+		slowQueryLogTimes:     make(map[string]time.Time),
+		slowQueryThreshold:    5 * time.Second,
```
kans, line 83: Thoughts on the values?

ggreer: I think 5 seconds is fine for now.

Can we limit the maximum number of items in slowQueryLogTimes? A sync does a lot of different queries and I don't want to keep a ton of them around. Also I think we should only log this once a minute maximum, not once a minute per unique query.

kans: The query that we key on is fully escaped/uninterpolated.  How many entries do you expect in that case?

ggreer (approved):

PR #331, opened 2025-03-18
Adds baton action traits

santhosh-c1:
- This is so that the consumer can identify common actions without needing to know the name

## Summary by CodeRabbit

- **New Features**
  - Enabled action definitions to support multiple traits, enhancing their expressiveness.
  - Introduced a new option for account deactivation as part of the expanded trait configuration.

### DIFF
--- a/proto/c1/connector/v2/action.proto
+++ b/proto/c1/connector/v2/action.proto
@@ -25,6 +25,11 @@ enum BatonActionStatus {
   BATON_ACTION_STATUS_FAILED = 5;
 }
 
+enum BatonActionTrait {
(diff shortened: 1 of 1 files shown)


proto/c1/connector/v2/action.proto
```diff
@@ -34,6 +39,8 @@ message BatonActionSchema {
 
   string display_name = 5;
   string description = 6;
+
+  repeated BatonActionTrait traits = 7;
```
magizh-c1, line 43: are we moving towards a trait instead of a action name to determine the action to run ? not following why.

santhosh-c1:
Closing for now, until we know more about traits.

PR #337, opened 2025-03-26
Encrypted Creds

kans:
lifted from kans/encrypted-connector-config

## Summary by CodeRabbit

- **New Features**
  - Introduced robust JWT authentication with token validation, enhanced middleware, and new configuration options.
  - AWS Lambda now supports the Ruby 3.4 runtime.
  - Updated configuration responses to use a raw byte format.

- **Chores**
  - Upgraded various third-party dependencies for improved performance and compatibility.
  - Corrected header formatting in gRPC requests.

- **Tests**
  - Added comprehensive unit tests for JWT and authentication middleware functionality.


pkg/auth/jwt.go
```diff
@@ -0,0 +1,143 @@
...
+		_, err = remoteKeySet.Get()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get keyset from JWKS URL: %w", err)
+		}
+		keySet = remoteKeySet
+	}
+
+	// Set default MaxExpiry if not specified.
+	if config.MaxExpirationFromNow == 0 {
+		config.MaxExpirationFromNow = 24 * time.Hour // TODO(morgabra): ??
```
ggreer, line 92: ??

pkg/lambda/grpc/config/config.go
```diff
@@ -99,16 +99,16 @@ func GetConnectorConfigServiceClient(ctx context.Context, clientID string, clien
 
 	dialOpts := []grpc.DialOption{
 		grpc.WithTransportCredentials(transportCreds),
-		grpc.WithUserAgent(fmt.Sprintf("%s baton-lambda/%s", clientName, "v0.0.1")),
+		grpc.WithUserAgent(fmt.Sprintf("%s baton-lambda/%s", clientID, "v0.0.1")),
```
ggreer, line 102: This version should probably be in a constant somewhere. Or maybe it should be the baton-sdk version?

kans: A const prefix would be nice at least.  I think we want the installation domain (dev instance/staging/prod) in the UA because we commingle lambdas in the same aws account for all dev instances and they will send traffic through the same proxy.

ggreer (approved):

PR #346, opened 2025-04-01
Add match id annotation

laurenleach:
## Summary by CodeRabbit

- **New Features**
	- Introduced identifier-based external resource matching, allowing resources to be matched using a unique string.
	- Enhanced validation with comprehensive error reporting for external resource match criteria.
	- Updated resource grant synchronization to support processing based on the new identifier annotation.

[BB-420](https://conductorone.atlassian.net/browse/BB-420)

[BB-420]: https://conductorone.atlassian.net/browse/BB-420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -1828,6 +1828,35 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 	expandedGrants := make([]*v2.Grant, 0)
 	for _, grant := range grants {
 		annos := annotations.Annotations(grant.Annotations)
+		matchResourceMatchIDAnno, err := GetExternalResourceMatchIDAnnotation(annos)
+		if err != nil {
+			return err
+		}
+		if matchResourceMatchIDAnno != nil {
+			for _, principal := range principals {
+				rAnnos := annotations.Annotations(principal.GetAnnotations())
+				batonID := &v2.BatonID{}
+				if !rAnnos.Contains(batonID) {
+					continue
+				}
+				if principal.GetId().GetResource() == matchResourceMatchIDAnno.Id {
+					newGrant := &v2.Grant{
+						Entitlement: grant.Entitlement,
+						Principal:   principal,
+						Id:          batonGrant.NewGrantID(principal, grant.Entitlement),
+						Sources:     grant.Sources,
+						Annotations: grant.Annotations,
+					}
+					expandedGrants = append(expandedGrants, newGrant)
+					// break out of principal list iteration since we found a match
+					break
+				}
+			}
+			// We still want to delete the grant even if there are no matches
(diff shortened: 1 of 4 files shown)


pkg/sync/syncer.go
```diff
@@ -1828,6 +1828,35 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 	expandedGrants := make([]*v2.Grant, 0)
 	for _, grant := range grants {
 		annos := annotations.Annotations(grant.Annotations)
+		matchResourceMatchIDAnno, err := GetExternalResourceMatchIDAnnotation(annos)
+		if err != nil {
+			return err
+		}
+		if matchResourceMatchIDAnno != nil {
+			for _, principal := range principals {
```
kans, line 1836: How many grants & principals do we expect?

laurenleach: I'm not sure, I'll use a map

pkg/sync/syncer.go
```diff
@@ -1828,6 +1845,26 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
...
+					Entitlement: grant.Entitlement,
+					Principal:   principal,
+					Id:          batonGrant.NewGrantID(principal, grant.Entitlement),
+					Sources:     grant.Sources,
+					Annotations: grant.Annotations,
+				}
+				expandedGrants = append(expandedGrants, newGrant)
+			}
+			// We still want to delete the grant even if there are no matches
+			grantsToDelete = append(grantsToDelete, grant.Id)
```
shackra, line 1864: Why you want to delete the grant that may have matches?? 🫨

shackra (commented):
Leaving one comment because it felt weird, just to be sure you truly intent what you intended

pkg/sync/syncer.go
laurenleach: we want to delete the grant even if there were no matches, these are for grants specifically for principals synced from another connector c1z, so if there are no principals matching our criteria, we delete the grant. 

for example, in gcp-wif if we have a policy that has ```principalSet://iam.googleapis.com/locations/global/workforcePools/pool-id/attribute.email/nonexistent@email.com``` 

and we use the annotation `ExternalResourceMatch{Key: "email", Value: "nonexistent@email.com"}` 

but we have no user with that email, there is no grant we should have here. Similar for `ExternalResourceMatchID`, if we don't have a principal with that ID, we don't have a valid grant.

laurenleach: For grants that do have matches, we generate a new grant using the principal that matches.

MarcusGoldschmidt (approved):

PR #349, opened 2025-04-07
Handle grant expansion for external resource matching

laurenleach:
## Summary by CodeRabbit

- **New Features**
  - Introduced dynamic grant expansion that automatically updates permissions based on expandable attributes and linked entitlements.
  
- **Stability Improvements**
  - Enhanced processing efficiency with improved error detection and logging during entitlement handling, ensuring more reliable updates to grants.

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -1893,6 +1893,10 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 					continue
 				}
 				resourceBID, err := bid.MakeBid(parsedEnt.Resource)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/sync/syncer.go
```diff
@@ -1967,8 +2059,6 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 		}
 	}
 
-	s.state.FinishAction(ctx)
```
ggreer, line 1970: Where do we finish this action now? Or were we not supposed to finish it?

laurenleach: the callers of this were already calling `s.state.FinishAction(ctx)`, so this was a bug that would finish the expansion action before we did it

ggreer (approved):

PR #351, opened 2025-04-09
Check any external resource match annos exist before any processing

laurenleach:
## Summary by CodeRabbit

- **Refactor**
  - Improved internal processing by filtering out items that don’t meet specific criteria during resource synchronization. This update reduces unnecessary operations, which may result in improved overall performance and a smoother user experience.

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -1856,6 +1856,10 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 
 	for _, grant := range grants {
 		annos := annotations.Annotations(grant.Annotations)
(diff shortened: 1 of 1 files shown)


pkg/sync/syncer.go
```diff
@@ -1856,6 +1856,10 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 
 	for _, grant := range grants {
 		annos := annotations.Annotations(grant.Annotations)
+		if !annos.Contains(&v2.ExternalResourceMatchAll{}) && !annos.Contains(&v2.ExternalResourceMatch{}) &&
```
ggreer, line 1859: It would be nice if we had annos.ContainsAny() which would help us avoid iterating over the annotations three times here.

ggreer (approved):

PR #355, opened 2025-04-16
HTTP wrapper: Add helper for link header pagination

ggreer:
Inspired by baton-http.

This was the trickiest of the pagination options. I'll update this PR with the simpler ones.

## Summary by CodeRabbit

- **New Features**
  - Added support for extracting next page links from HTTP headers to enable NextLink pagination in API responses.

- **Tests**
  - Introduced tests to verify correct parsing of HTTP Link headers for pagination.

### DIFF
--- a/pkg/uhttp/pagination.go
+++ b/pkg/uhttp/pagination.go
@@ -1,0 +1,99 @@
+package uhttp
+
+import (
+	"strings"
+
+	"github.com/conductorone/baton-sdk/pkg/pagination"
+)
+
+/*
+ * Uhttp pagination handling.
+ * There are three common types of pagination:
+ * 1. NextLink: http header containing a url to fetch the next page
+ * 2. Cursor: http body containing a token to fetch the next page
+ * 3. Offset: offset + limit to fetch the next page
+ *    - Subset of offset: incremental page numbers
+ *
+ * All of these helper functions take a bag and push the next page state on (if there is a next page).
+ */
+
+type NextLinkConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/uhttp/pagination.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"strings"
+
+	"github.com/conductorone/baton-sdk/pkg/pagination"
+)
+
+/*
+ * Uhttp pagination handling.
+ * There are three common types of pagination:
```
gontzess, line 11: i'm not sure how common it is (maybe its not very), but for the Rootly API which followed the JSONAPI spec, `links` are included as part of the body, separate from `data`: https://github.com/ConductorOne/baton-rootly/blob/main/pkg/connector/client/models.go#L36

thought i'd mention for awareness

pkg/uhttp/pagination.go
```diff
@@ -0,0 +1,99 @@
...
+func parseLinkHeader(header string) (map[string]string, error) {
+	if header == "" {
+		// Empty header is fine, it just means there are no more pages.
+		return nil, nil
+	}
+
+	links := make(map[string]string)
+	headerLinks := strings.Split(header, ",")
+	for _, headerLink := range headerLinks {
+		linkParts := strings.Split(headerLink, ";")
```
gontzess, line 37: maybe this has an obvious answer, but how do we know these headers will always be listed in this format with `,` and `;`? like is this http spec or a format that you've already established elsewhere?

gontzess: oh perfect, thanks for linking. i was unsuccessful at searching for this earlier

PR #356, opened 2025-04-23
Baton ID: Compare keys and values with `strings.EqualFold`

shackra:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved matching of external resource identifiers by making key and profile value comparisons case-insensitive. This ensures more consistent results regardless of letter casing in user and group traits.

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -1974,7 +1974,7 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 						l.Error("error getting user trait", zap.Any("userPrincipal", userPrincipal))
 						continue
 					}
-					if strings.EqualFold(matchExternalResource.Key, "email") {
+					if matchExternalResource.Key == "email" {
 						if userTraitContainsEmail(userTrait.Emails, matchExternalResource.Value) {
 							newGrant := newGrantForExternalPrincipal(grant, userPrincipal)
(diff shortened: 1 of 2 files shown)


pkg/sync/syncer.go
```diff
@@ -1974,7 +1974,7 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 						l.Error("error getting user trait", zap.Any("userPrincipal", userPrincipal))
 						continue
 					}
-					if matchExternalResource.Key == "email" {
+					if strings.EqualFold(matchExternalResource.Key, "email") {
```
ggreer, line 1977: We're the ones making this key, so I think it's fine to use == here. If it's not "email" (all lower case), something has gone wrong.

pkg/sync/syncer.go
```diff
@@ -1974,7 +1974,7 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
...
 						if userTraitContainsEmail(userTrait.Emails, matchExternalResource.Value) {
```
ggreer, line 1978: I think you'll need to mess with some strings in userTraitContainsEmail() to get emails to match if they differ in case.

laurenleach: I used `EqualFold` in `userTraitContainsEmail`

ggreer (approved):

PR #367, opened 2025-05-06
introduce resource changed event

mj-palanker:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new event type to the event feed for tracking resource changes.
  - Added validation for the new resource change event, ensuring accurate and reliable event data.
  - Added validation support for resource identifiers with optional parent references to enhance data integrity.


proto/c1/connector/v2/event_feed.proto
```diff
@@ -71,3 +72,8 @@ message RevokeEvent {
   c1.connector.v2.Entitlement entitlement = 1;
   c1.connector.v2.Resource principal = 2;
 }
+
+// generic light weight event indicating a resource was changed
+message ResourceChangeEvent {
+  c1.connector.v2.ResourceId resource_id = 1;
```
ggreer, line 78: We probably want an optional parent resource ID in this because in some cases, a resource can only be fetched if we know its parent resource. One way of doing this would be to make this a resource instead of a resource ID, as resource has a parent resource ID attribute. Either solution makes sense to me.

ggreer: Actually I think using a resource has the advantage of letting connectors return the updated resource, allowing us to skip a round trip of "resource X updated? ok crawl this resource". If the resource is sparsely populated (just contains an ID), then we'd queue up a targeted sync.

mj-palanker: talked w/ brad earlier about having more complex events which actually contain enough information to produce their results directly, but we're not aiming to do that yet (but boy are there exciting opportunities there). We likely want different event types themselves in those cases. Thinking about the full Resource thing though... or type annotations? there was clear ask from Paul to make these lightweight, but not sure exact motivation there.

mj-palanker: talked again w/ brad and logan, we're now making the content a oneof and a new message which is resource ID and Parent resource ID. we can later take a full resource, or some kind of resource diff if we want to.

loganintech (approved):

proto/c1/connector/v2/event_feed.proto
loganintech: love it

we can always add new types here

PR #369, opened 2025-05-06
remove nested oneof

mj-palanker:


loganintech (approved):

proto/c1/connector/v2/event_feed.proto
```diff
@@ -75,7 +75,5 @@ message RevokeEvent {
 
 // generic light weight event indicating a resource was changed
 message ResourceChangeEvent {
-  oneof resource {
-    c1.connector.v2.ResourceIdWithParentId resource_id_with_parent_id = 100;
-  }
+  c1.connector.v2.ResourceIdWithParentId resource_id_with_parent_id = 100;
```
ggreer, line 78: Nitpick: ResourceIdWithParentId isn't used by anything else. We can avoid a nil check in go code by having the attributes directly on the ResourceChangeEvent:

```suggestion
  ResourceId resource_id = 1 [(validate.rules).message = {required: true}];
  ResourceId parent_resource_id = 2;
```

ggreer (approved):

proto/c1/connector/v2/event_feed.proto
mj-palanker: done

PR #371, opened 2025-05-07
Implement a naive compactor algorithm to apply C1Zs on top of a base C1Z

loganintech:
## Summary by CodeRabbit

- **New Features**
  - Added conditional update methods for entitlements, grants, resources, and resource types, allowing updates only if incoming data is newer.
  - Introduced a compaction feature to merge multiple sync files into a single compacted file.
  - Added a naive compactor that overlays newer data on top of older data during compaction.
  - Implemented generic pagination support for protobuf-based list operations to streamline data retrieval.
  - Added support for a new sync compactor mode in the connector runner and CLI, enabling on-demand sync file compaction.
  - Introduced new configuration fields to enable and configure sync file compaction.
  - Added validation and protobuf support for the new compact syncs task type.

- **Refactor**
  - Unified and modularized bulk insertion logic for connector objects and related entities, improving code reuse and maintainability.


pkg/synccompactor/compactor.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return &Compactor{entries: compactableSyncs, destDir: destDir}, nil
+}
+
+func (c *Compactor) Compact(ctx context.Context) (*CompactableSync, error) {
+	if len(c.entries) < 2 {
+		return nil, nil
+	}
+
+	tempDir := os.TempDir()
```
ggreer, line 39: It's fine to call NewC1ZFile() without a temp dir, as it calls os.MkdirTemp() which, if the first argument is empty, uses os.TempDir().

pkg/synccompactor/compactor.go
```diff
@@ -0,0 +1,124 @@
...
+
+	baseFile, err := baseC1Z.LoadC1Z(ctx)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	latestAppliedSync, err := baseFile.GetSync(ctx, &reader_v2.SyncsReaderServiceGetSyncRequest{
+		SyncId:      info.syncID,
+		Annotations: nil,
+	})
```
ggreer, line 71: If baseFile.GetSync() errors (if the sync ID doesn't exist, for example), we don't close baseFile or baseC1Z.

pkg/synccompactor/naive/naive.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := n.processResourceTypes(ctx); err != nil {
+		return err
+	}
+	if err := n.processResources(ctx); err != nil {
+		return err
+	}
+	if err := n.processEntitlements(ctx); err != nil {
+		return err
+	}
+	if err := n.processGrants(ctx); err != nil {
```
ggreer, line 34: Partial syncs don't expand grants or external IDs, so that will need to be done once everything's compacted.

pkg/synccompactor/compactor.go
loganintech: great

loganintech: But how do I get the filename?

pkg/synccompactor/compactor.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+
+	return latestAppliedSync.Sync, baseFile, baseC1Z, func() {
+		_ = baseFile.Close()
+		_ = baseC1Z.Close(ctx)
+	}, nil
+}
```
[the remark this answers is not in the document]
loganintech: Good suggestion

pkg/synccompactor/naive/naive.go
loganintech: That's ok, this should support any kind of sync on the applied side. This just copies the data from the tables over. But yes running grant expansion seems valuable.

Maybe we can just do that on the final compacted result?

pkg/synccompactor/compactor.go
loganintech: Fixed

pkg/synccompactor/compactor.go
loganintech: Just kidding, I still pass the filename as-is I just remove the temp dir

ggreer (approved):
Probably best to merge this and then clean it up in a subsequent PR. I know our table names won't have special characters in them, but it would be nice to fix that potential sql injection spot.

loganintech:
@ggreer will merge as soon as I get it to work once :)

PR #374, opened 2025-05-10
[BB-700] Just log a warning if connector can't create debug.log

shackra:
This PR should stop connectors from failing if the file to output debugging information cannot be created. Also, some information about the system is collected (the current working directory) because the connector may be looking to create it somewhere where it should

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of debug log file creation and upload with fallback to the current directory and clearer warnings when the temporary directory is unavailable.

- **Style**
  - Enhanced code formatting for consistency, including blank lines and struct declaration style.

### DIFF
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -53,7 +53,6 @@ func (c *connectorRunner) Run(ctx context.Context) error {
 		var err error
 		tempDir := c.tasks.GetTempDir()
 		if tempDir == "" {
-			// Obtener el directorio de trabajo actual
 			wd, err := os.Getwd()
 			if err != nil {
 				l.Info("unable to get the current working directory")
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -45,20 +45,25 @@ var ErrSigTerm = errors.New("context cancelled by process shutdown")
 
 // Run starts a connector and creates a new C1Z file.
 func (c *connectorRunner) Run(ctx context.Context) error {
+	l := ctxzap.Extract(ctx)
 	ctx, cancel := context.WithCancelCause(ctx)
 	defer cancel(ErrSigTerm)
 
 	if c.tasks.ShouldDebug() && c.debugFile == nil {
 		var err error
-		c.debugFile, err = os.Create(filepath.Join(c.tasks.GetTempDir(), "debug.log"))
+		tempDir := c.tasks.GetTempDir()
+		if tempDir == "" {
+			l.Warn("no temporal folder found on this system according to our task manager")
+		}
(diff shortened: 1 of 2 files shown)


pkg/connectorrunner/runner.go
```diff
@@ -45,20 +45,35 @@ var ErrSigTerm = errors.New("context cancelled by process shutdown")
...
 func (c *connectorRunner) Run(ctx context.Context) error {
+	l := ctxzap.Extract(ctx)
 	ctx, cancel := context.WithCancelCause(ctx)
 	defer cancel(ErrSigTerm)
 
 	if c.tasks.ShouldDebug() && c.debugFile == nil {
 		var err error
-		c.debugFile, err = os.Create(filepath.Join(c.tasks.GetTempDir(), "debug.log"))
+		tempDir := c.tasks.GetTempDir()
+		if tempDir == "" {
```
btipling, line 55: how about just

```_, err := os.Stat(tempDir)
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			l.Warn("debug log file does not exists", zap.Error(err))
			return nil
		}
		return err
	} else {
	... the rest of the code
```
that's what what we do here https://github.com/ConductorOne/baton-sdk/blob/main/pkg/tasks/c1api/full_sync.go#L190

pkg/connectorrunner/runner.go
```diff
@@ -45,20 +45,35 @@ var ErrSigTerm = errors.New("context cancelled by process shutdown")
...
+			wd, err := os.Getwd()
+			if err != nil {
+				l.Info("unable to get the current working directory")
```
btipling, line 58: we can't imo just info on an err that error could be anything, we usually know which error we want to ignore

pkg/connectorrunner/runner.go
```diff
@@ -45,20 +45,35 @@ var ErrSigTerm = errors.New("context cancelled by process shutdown")
...
+					zap.String("current working directory", wd))
+			} else {
+				l.Warn("no temporal folder found on this system according to our task manager")
+			}
+		}
+		debugFile := filepath.Join(tempDir, "debug.log")
+		c.debugFile, err = os.Create(debugFile)
 		if err != nil {
-			return err
+			l.Warn("cannot create file", zap.String("full file path", debugFile))
```
btipling, line 71: can't just warn on any error whatsoever and keep going

pkg/connectorrunner/runner.go
shackra: I've have changed this and the other file (full_sync.go) too. Thoughts?

shackra: as in having consistent code across files? Well, that could work too

PR #375, opened 2025-05-12
[BB-621] baton-sdk: add retry mechanism to grant and revoke method

Bencheng21:
Currently, the Grant and Revoke methods do not have a retry mechanism.
If either of these methods fails, the entire workflow fails, requiring customers to resubmit the ticket — which is not user-friendly.

In this PR, we introduce a retry mechanism that attempts the operation up to 2 times before returning an error.
Since the Temporal activity has a timeout of 5 minutes, we avoid excessive retries to prevent C1 tickets from getting stuck.

#### Test. 
image

### DIFF
--- a/pkg/connectorbuilder/connectorbuilder.go
+++ b/pkg/connectorbuilder/connectorbuilder.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"slices"
 	"sort"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -1327,3 +1368,49 @@ func (b *builderImpl) GetActionStatus(ctx context.Context, request *v2.GetAction
 	b.m.RecordTaskSuccess(ctx, tt, b.nowFunc().Sub(start))
 	return resp, nil
 }
+
+func (b *builderImpl) shouldWaitAndRetry(ctx context.Context, err error, baseDelay time.Duration) bool {
```
ggreer, line 1372: This is copied from syncer, right? It should only live in one place.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -955,34 +976,54 @@ func (b *builderImpl) Revoke(ctx context.Context, request *v2.GrantManagerServic
...
+		baseDelay = 30 * time.Second
+		rt        = request.Grant.Entitlement.Resource.Id.ResourceType
+	)
+	for {
+		provisioner, ok := b.resourceProvisioners[rt]
+		if ok {
+			annos, err := provisioner.Revoke(ctx, request.Grant)
+			if err != nil {
+				l.Error("error: revoke failed", zap.Error(err))
+				b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
```
ggreer, line 990: This records the task as failed, then calls the code to potentially retry. I think it'll mess up our metrics.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -955,34 +976,54 @@ func (b *builderImpl) Revoke(ctx context.Context, request *v2.GrantManagerServic
...
-			l.Error("error: revoke failed", zap.Error(err))
-			b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
-			return nil, fmt.Errorf("error: revoke failed: %w", err)
+	var (
+		attempt   = 0
+		baseDelay = 30 * time.Second
+		rt        = request.Grant.Entitlement.Resource.Id.ResourceType
+	)
+	for {
+		provisioner, ok := b.resourceProvisioners[rt]
```
ggreer, line 985: The provisioner won't change. It's probably best to push this out of the for loop.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -914,36 +915,56 @@ func (b *builderImpl) Grant(ctx context.Context, request *v2.GrantManagerService
...
-	provisioner, ok := b.resourceProvisioners[rt]
-	if ok {
-		annos, err := provisioner.Grant(ctx, request.Principal, request.Entitlement)
-		if err != nil {
-			l.Error("error: grant failed", zap.Error(err))
-			b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
-			return nil, fmt.Errorf("error: grant failed: %w", err)
+	var (
+		attempt   = 0
+		baseDelay = 30 * time.Second
```
ggreer, line 920: I'd make the default a little quicker. Maybe 10 seconds? Then the wait is 10, 20, 40, right? And it's either success or failure after a 70 seconds. With a base delay of 30 seconds, it'd be 30, 60, 120, for a total of 3.5 minutes. That seems a little sluggish.

Bencheng21: Given litmos connector allows 100 requests / minute, 
`10s and 20s` looks aggresive to me. it's most likely the retry will still fail after just 10 seconds.

pkg/connectorbuilder/connectorbuilder.go
Bencheng21: Mostly copied from syncer, but it's a little bit different. 
1. syncer retrys in 1s, 2s, 3s, etc. Retrying in seconds is aggresive in provisioner. 
2. syncer activity has a timeout of 24 hours, so we could retry multiple times. However, provisioner activity has a timeout of 5 minutes, so it's not ok to retry multiple times.

ggreer (approved):

PR #380, opened 2025-05-12
Generalize retry code

ggreer:
Syncing, granting, and revoking all have retry logic, and they're all pretty similar. This unifies them into a single retryer package.

## Summary by CodeRabbit

- **New Features**
  - Introduced a standardized retry mechanism with configurable backoff for improved reliability during operations that may encounter transient errors.

- **Refactor**
  - Unified and simplified retry logic across grant, revoke, and sync operations, reducing code duplication and enhancing maintainability.

- **Tests**
  - Added tests to validate the new retry behavior and ensure correct handling of retryable and non-retryable errors.

### DIFF
--- a/pkg/connectorbuilder/connectorbuilder.go
+++ b/pkg/connectorbuilder/connectorbuilder.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"slices"
 	"sort"
 	"time"
@@ -23,6 +22,7 @@ import (
 	"github.com/conductorone/baton-sdk/pkg/crypto"
 	"github.com/conductorone/baton-sdk/pkg/metrics"
 	"github.com/conductorone/baton-sdk/pkg/pagination"
+	"github.com/conductorone/baton-sdk/pkg/retry"
 	"github.com/conductorone/baton-sdk/pkg/types"
 	"github.com/conductorone/baton-sdk/pkg/types/tasks"
 	"github.com/conductorone/baton-sdk/pkg/uhttp"
@@ -915,11 +915,13 @@ func (b *builderImpl) Grant(ctx context.Context, request *v2.GrantManagerService
 	tt := tasks.GrantType
 	l := ctxzap.Extract(ctx)
 
-	var (
-		attempt   = 0
-		baseDelay = 30 * time.Second
-		rt        = request.Entitlement.Resource.Id.ResourceType
-	)
+	rt := request.Entitlement.Resource.Id.ResourceType
+
+	retryer := retry.NewRetryer(ctx, retry.RetryConfig{
+		MaxAttempts:  3,
+		InitialDelay: 30 * time.Second,
+		MaxDelay:     120 * time.Second,
+	})
 
 	provisioner, v1ok := b.resourceProvisioners[rt]
 	provisionerV2, v2ok := b.resourceProvisionersV2[rt]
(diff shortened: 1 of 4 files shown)


btipling (approved):

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -915,54 +915,48 @@ func (b *builderImpl) Grant(ctx context.Context, request *v2.GrantManagerService
...
-		baseDelay = 30 * time.Second
-		rt        = request.Entitlement.Resource.Id.ResourceType
-	)
+	rt := request.Entitlement.Resource.Id.ResourceType
 
-	provisioner, v1ok := b.resourceProvisioners[rt]
-	provisionerV2, v2ok := b.resourceProvisionersV2[rt]
-	if !v1ok && !v2ok {
+	retryer := retry.NewRetryer(ctx, retry.RetryConfig{
+		MaxAttempts:  3,
```
Bencheng21, line 921: MaxAttempts should be 2. 
The Temporal activity has a timeout of 5 minutes, so the time from start to close shouldn't exceed that limit. c1 ticket would end up in a weird state if we retry 3 times. 

For example, 
image

Customers are not aware of the current status of the ticket.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -915,54 +915,48 @@ func (b *builderImpl) Grant(ctx context.Context, request *v2.GrantManagerService
...
 		b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
 		return nil, fmt.Errorf("error: resource type does not have provisioner configured")
 	}
 
 	for {
-		if v1ok {
-			annos, err := provisioner.Grant(ctx, request.Principal, request.Entitlement)
-			if err != nil {
-				l.Error("error: grant failed", zap.Error(err))
-				if !b.shouldWaitAndRetry(ctx, err, baseDelay) || attempt >= 2 {
```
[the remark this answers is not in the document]
ggreer: Yes, the old code tried twice before erroring. I changed it to three because that seems like a good number of attempts before giving up.

pkg/connectorbuilder/connectorbuilder.go
ggreer: Three retries would wait 30, 60, then 90 seconds, which is 3 minutes total.

pkg/retry/retry.go
```diff
@@ -0,0 +1,113 @@
...
+	l := ctxzap.Extract(ctx)
+
+	if r.maxAttempts > 0 && r.attempts > r.maxAttempts {
+		l.Warn("max attempts reached", zap.Error(err), zap.Uint("max_attempts", r.maxAttempts))
+		return false
+	}
+
+	// use linear backoff by default
+	var wait time.Duration
+	if r.attempts > math.MaxInt64 {
```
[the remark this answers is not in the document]
ggreer: It was to get the linter to shut up.

pkg/connectorbuilder/connectorbuilder.go
Bencheng21: Three minutes seems to be the ideal timing. I haven’t looked into the code to see how C1 tickets change status, but based on my testing, three attempts don’t work — the C1 ticket seems to be stuck.

Bencheng21: image
I am not sure if it's because of dev env. I retry twice, which is 30 + 60 seconds, but the c1 ticket still takes around 4 minutes to respond. 
so I don't think it would work if the last retry takes 90 seconds.

PR #384, opened 2025-05-14
Better usage of the filesystem in the compactor

loganintech:
## Summary by CodeRabbit

- **Refactor**
  - Introduced a managed temporary working directory for intermediate files during compaction.
  - Replaced OS-level file renames with explicit file copying for final output.
  - Removed intermediate file deletion from the compaction process.
  - Simplified file path handling with absolute path resolution.

- **Bug Fixes**
  - Added automatic cleanup of temporary working directories after compaction.

- **Tests**
  - Refactored tests to support flexible compactor initialization and cleanup.
  - Ensured tests verify correct compaction results with temporary directory usage.

- **Chores**
  - Improved logging to show temporary directory locations during compaction.

### DIFF
--- a/pkg/synccompactor/compactor.go
+++ b/pkg/synccompactor/compactor.go
@@ -56,20 +56,13 @@ func NewCompactor(ctx context.Context, outputDir string, compactableSyncs []*Com
 
 	// If no tmpDir is provided, use the tmpDir
 	if c.tmpDir == "" {
-		c.tmpDir = path.Join(os.TempDir(), "baton_sync_compactor")
+		c.tmpDir = os.TempDir()
 	}
-
-	createdDir := false
-	if _, err := os.Stat(c.tmpDir); err != nil {
-		if os.IsNotExist(err) {
-			if err := os.MkdirAll(c.tmpDir, 0755); err != nil {
-				return nil, nil, err
-			}
-			createdDir = true
-		} else {
-			return nil, nil, err
-		}
+	tmpDir, err := os.MkdirTemp(c.tmpDir, "baton-sync-compactor-")
+	if err != nil {
+		return nil, nil, err
 	}
+	c.tmpDir = tmpDir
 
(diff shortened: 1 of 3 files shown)


pkg/synccompactor/compactor.go
```diff
@@ -81,55 +86,35 @@ func (c *Compactor) Compact(ctx context.Context) (*CompactableSync, error) {
...
-	return base, nil
+	return &CompactableSync{FilePath: finalPath, SyncID: base.SyncID}, nil
 }
 
 func mvFile(sourcePath string, destPath string) error {
 	source, err := os.Open(sourcePath)
 	if err != nil {
 		return fmt.Errorf("failed to open source file: %w", err)
 	}
+	defer source.Close()
```
ggreer, line 117: This now copies, not moves, right?

loganintech: Correct, that's why moving the close from defer to here is ok, because we copy it and leave the original for the caller of the compactor to cleanup later.

pkg/synccompactor/compactor_test.go
```diff
@@ -145,8 +183,9 @@
 
 	// Create compactor
 	compactableSyncs := []*CompactableSync{firstCompactableSync, secondCompactableSync}
-	compactor, err := NewCompactor(ctx, outputDir, compactableSyncs, WithTmpDir(tmpDir))
+	compactor, cleanup, err := createCompactor(compactableSyncs)
 	require.NoError(t, err)
+	defer cleanup()
```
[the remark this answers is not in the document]
loganintech: I'm ok with the test failing if cleanup fails. Cleanup is just best-effort

pkg/synccompactor/compactor.go
loganintech: Oh yes good point

ggreer (approved):

PR #385, opened 2025-05-14
EventProviderV2 for multiple event streams

mj-palanker:
Co authored w/ @johnallers 
New Interface "EventProviderV2" which allows a connector to implement multiple event feeds. These feeds self describe via the EventFeedMetadata object, and implement the same ListEvents as the previous eventProvider. The Baton "ListEvents" request now takes an optional feed-id which indicates that you want to hit a specific v2 event feed. No id defaults you back to the old interface. 

## Summary by CodeRabbit

- **New Features**
  - Added support for multiple event feeds with event feed IDs and configurable start times for event consumption.
  - Introduced event feed metadata, new event types, and start position options for enhanced event stream control.
  - Added CLI options and default fields to specify event feed ID and start time.
  - Added a new task type for listing event feeds.
  - Added an API method to list available event feeds.

- **Enhancements**
  - Improved validation and error reporting for event feed requests and metadata.
  - Updated system capabilities to indicate multi-feed event provider support.


pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -135,13 +135,35 @@ type CredentialManager interface {
...
+	EventLister
+}
+
+// NewEventProviderV2 is a new interface that allows connectors to provide multiple event feeds.
+//
+// This is the recommended interface for implementing event feed support in new connectors.
+// TODO MJP does this get auto made when an EventFeed is registered? What does that look like? How do I make this easiest?
+type EventProviderV2 interface {
+	ConnectorBuilder
+	EventFeeds(ctx context.Context) []EventFeed
```
johnallers, line 158: @mj-palanker We may want to add `error` as a return parameter for `EventFeeds()`. I'm imagining having to do some feed specific wireup that might fail here.

For Google Workspace, I started on the path of creating a specific API service for each feed (admin and usage) to support `ListEvents()`. I ended up backing that out, because there was already a mechanism to handle that on the connector-level, but maybe that's not always the case?

mj-palanker: we talked on zoom and we think it's fine to do it w/o an error possibility as that's how resource syncers do it.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -430,6 +454,19 @@ func NewConnector(ctx context.Context, in interface{}, opts ...Opt) (types.Conne
...
+				if _, ok := ret.eventFeeds[feedData.Id]; ok {
+					return nil, fmt.Errorf("error: duplicate event feed id found: %s", feedData.Id)
+				}
+				ret.eventFeeds[feedData.Id] = ef
+			}
+		}
+
 		if b, ok := c.(EventProvider); ok {
 			ret.eventFeed = b
 		}
```
mj-palanker, line 472: If we have a v1 feed, should it get registered as a v2 feed as well?

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -1032,12 +1073,29 @@ func (b *builderImpl) ListEvents(ctx context.Context, request *v2.ListEventsRequ
...
+
+	feedId := request.EventFeedId
+	if feedId == "" {
+		// We're requesting the old event feed
+		if b.eventFeed == nil {
+			b.m.RecordTaskFailure(ctx, tasks.ListEventsType, b.nowFunc().Sub(start))
+			return nil, fmt.Errorf("error: event feed not implemented")
+		}
+		return b.listEventsHandleFeed(ctx, request, b.eventFeed, start)
 	}
```
mj-palanker, line 1085: Similar to above, we have internally called these feeds by a static constant name (since there's only one of them) should we allow that as a pass-through to an old feed if you don't have one by that name? Migration would look like moving the feed to the new version, but keeping the "well known constant ID"... that sounds right to me.

mj-palanker:
Wait I definitely need grpcs to get this new information out, right?

proto/c1/connector/v2/event_feed.proto
```diff
@@ -13,6 +13,7 @@ option go_package = "github.com/conductorone/baton-sdk/pb/c1/connector/v2";
 
 service EventService {
   rpc ListEvents(ListEventsRequest) returns (ListEventsResponse);
+  rpc ListEventFeeds(ListEventFeedsRequest) returns (ListEventFeedsResponse);
```
mj-palanker, line 16: Is it a problem to have added this directly to the existing event service? or is that fine?

proto/c1/connector/v2/event_feed.proto
```diff
@@ -41,6 +49,29 @@ message ListEventsResponse {
...
+  string page_token = 2 [(validate.rules).string = {
+    ignore_empty: true
+    min_bytes: 1
+    max_bytes: 1048576 // 1MB
+  }];
+  repeated google.protobuf.Any annotations = 3;
+}
+
+message ListEventFeedsResponse {
+  repeated EventFeedMetadata list = 1;
```
ggreer, line 66: Connectorbuilder's ListEventFeeds doesn't obey the page token or page size right now, so I think we can kill the page size/token and just always list all event feeds. Although we have no limit on the number of feeds, if a connector has more than a few then we are doing something wrong.

proto/c1/connector/v2/event_feed.proto
```diff
@@ -78,3 +109,35 @@ message ResourceChangeEvent {
...
+    min_bytes: 1
+    max_bytes: 1024
+  }];
+  repeated EventType supported_event_types = 2 [(validate.rules).repeated = {
+    unique: true
+    items: {
+      enum: {defined_only: true}
+    }
+  }];
+  EventFeedStartAt start_at = 3;
```
ggreer, line 141: Is EventFeedStartAt to tell us how the feed behaves? I think some event feeds let you sort in either direction, so maybe this should be a set (like supported_event_types) instead of one enum.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -135,13 +135,54 @@ type CredentialManager interface {
...
+
+type oldEventFeedWrapper struct {
+	feed EventLister
+}
+
+func (e *oldEventFeedWrapper) EventFeedMetadata(ctx context.Context) *v2.EventFeedMetadata {
+	return &v2.EventFeedMetadata{
+		Id:                  "baton_feed_event",
+		SupportedEventTypes: []v2.EventType{v2.EventType_EVENT_TYPE_UNSPECIFIED},
+		StartAt:             v2.EventFeedStartAt_EVENT_FEED_START_AT_TAIL,
```
ggreer, line 176: Do all existing event feeds start at the tail?

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -1027,17 +1097,58 @@ func (b *builderImpl) GetAsset(request *v2.AssetServiceGetAssetRequest, server v
...
 	defer span.End()
 
 	start := b.nowFunc()
-	tt := tasks.ListEventsType
-	if b.eventFeed == nil {
-		b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
-		return nil, fmt.Errorf("error: event feed not implemented")
+
+	feedId := request.EventFeedId
+	if feedId == "" {
```
ggreer, line 1131: Is it simpler to default feedId to baton_feed_event?

Ideally we'd have exactly one way of addressing event feeds. If we can remove b.eventFeed entirely and only use b.eventFeeds, I think the code would be easier to understand.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -1027,17 +1097,58 @@ func (b *builderImpl) GetAsset(request *v2.AssetServiceGetAssetRequest, server v
...
+func (b *builderImpl) ListEventFeeds(ctx context.Context, request *v2.ListEventFeedsRequest) (*v2.ListEventFeedsResponse, error) {
+	ctx, span := tracer.Start(ctx, "builderImpl.ListEventFeeds")
+	defer span.End()
+
+	start := b.nowFunc()
+	tt := tasks.ListEventFeedsType
+
+	if len(b.eventFeeds) == 0 {
+		b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
+		return nil, fmt.Errorf("error: no event feeds found")
```
ggreer, line 1109: Why not return an empty list instead of erroring? I'm ambivalent about what to do here. I think either behavior is acceptable. I'm not sure which one is less surprising.

If we do return an error, it'd be nice if it had the appropriate grpc status code.

proto/c1/connector/v2/event_feed.proto
mj-palanker: Sorry, `EventFeedStartAt` tells c1 if we care about historic events or only very recent events. currently we always get 90 days of history backwards, but that's totally unneeded for things which only care about live changes. Usage events for instance we do need to process all of, while live change feed can start at the time it's initiated.

pkg/connectorbuilder/connectorbuilder.go
mj-palanker: It's always safe to process more (though not always needed) as opposed to dropping anything.

pkg/connectorbuilder/connectorbuilder.go
mj-palanker: I like that lemme try it.

pkg/connectorbuilder/connectorbuilder.go
mj-palanker: went with just returning empty. We had been relying on weird stuff in c1 like "call and see if it errors" but this is in the capabilities properly, so it should be fine.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -430,8 +476,31 @@ func NewConnector(ctx context.Context, in interface{}, opts ...Opt) (types.Conne
...
 		}
 
+		if b, ok := c.(EventProviderV2); ok {
+			for _, ef := range b.EventFeeds(ctx) {
+				feedData := ef.EventFeedMetadata(ctx)
+				if feedData == nil {
+					return nil, fmt.Errorf("error: event feed metadata is nil")
+				}
+				if err := feedData.Validate(); err != nil {
+					return nil, fmt.Errorf("error: event feed metadata is invalid: %w", err)
```
ggreer, line 486: I just noticed, but it'd be nice to try and log the ID or something that narrows down which event feed is invalid.

proto/c1/connector/v2/event_feed.proto
ggreer: OK, so wouldn't this be a set similar to supported_event_types? (eg: For APIs that support both fetching historical data and tailing the latest stuff.) Or is the idea to expose the same underlying event API as multiple event feeds which differ only by their ID and start_at?

I thought ListEventsRequest's start_at would tell whether to start with historical or latest data. Maybe the names should be different since one is an enum and one is a timestamp.

ggreer (approved):

PR #387, opened 2025-05-22
Add helper functions for generic http response handling and making new entitlements.

ggreer:
## Summary by CodeRabbit

- **New Features**
  - Added a flexible entitlement creation method that allows dynamic assignment of entitlement purposes based on a string input.
  - Introduced a generic HTTP response handler that supports both JSON and XML formats, with improved handling for list and map responses.

### DIFF
--- a/pkg/uhttp/wrapper.go
+++ b/pkg/uhttp/wrapper.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"reflect"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/uhttp/wrapper.go
```diff
@@ -249,6 +250,50 @@ func WithResponse(response interface{}) DoOption {
...
+			err = WithJSONResponse(&v)(resp)
+			if err != nil {
+				return err
+			}
+			if list, ok := v.([]any); ok {
+				(*response)["items"] = list
+			} else if vMap, ok := v.(map[string]any); ok {
+				*response = vMap
+			} else {
+				return status.Errorf(codes.Internal, "unsupported content type: %s", reflect.TypeOf(v))
```
pquerna, line 273: this can just be `%T` i think

PR #394, opened 2025-06-09
Avoid overriding specific Content-Type and Accept headers

elzorrorebelde:
For some APIs, like Share Point REST API, providing additional information on the Content-Type can determine if the request fails or success (given that the actual content is what the API expects). This change prevents the SDK from blindly overwriting values for those two headers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of Content-Type and Accept headers when composing requests with JSON bodies: custom header values that include parameters (e.g., charset or vendor params) are preserved appropriately while default JSON headers are applied as needed.

- **Tests**
  - Added tests confirming requests with custom Content-Type or Accept headers plus JSON body options produce the expected method, URL, headers, and JSON-encoded body.


shackra:
El martes 26 de agosto del 2025, cursor[bot] escribió:

> @cursor[bot] commented on this pull request.
> The special handling for Content-Type and Accept headers has two 
> issues. The logic's case-sensitive map access
> can prevent existing headers from being found, leading to 
> duplicate logical headers.

you mean, like having `Content-Type` there and then setting
`content-type` on the map? well, that could be a problem, yes.

> Also, it only updates
> these headers if the new value contains a semicolon, silently 
> ignoring explicit overrides with simpler values.

that's the point, my binary friend, as that was the previous 
behavior (always overwriting
the values)

-- 
Jorge Araya

Contacto:
 Telegram: t.me/shackra · Signal: Shackra.28

PR #400, opened 2025-06-25
Append logs to lambda invoke function error message

laurenleach:
This includes specific information about the error 

image

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error messages for failed Lambda function invocations by including a filtered summary of relevant logs, making troubleshooting easier for users.


pkg/lambda/grpc/client.go
```diff
@@ -127,3 +145,31 @@ func NewClientConn(transport ClientTransport) grpc.ClientConnInterface {
...
+func extractMeaningfulLogLines(raw string) string {
+	lines := strings.Split(raw, "\n")
+	var filtered []string
+
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+
+		// Skip noisy AWS system log lines
+		if line == "" ||
+			strings.HasPrefix(line, "START RequestId:") ||
```
ggreer, line 158: Maybe this should be a slice of strings and just use `slices.ContainsFunc(...)`. It's technically less efficient, but it would be easier to read. I don't have a strong opinion on this. Just something I thought of while looking at the diff.

ggreer (approved):

PR #408, opened 2025-07-22
Add retries to ListTicketSchemas

mbburch:
This PR adds support for handling rate limits occurring when attempting to get ticket schemas. Making this change in the sdk supports [this ticket](https://conductorone.atlassian.net/browse/BB-1129) for Linear.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of ticket schema listing by adding automatic retry logic for failed attempts.

### DIFF
--- a/pkg/connectorbuilder/connectorbuilder.go
+++ b/pkg/connectorbuilder/connectorbuilder.go
@@ -325,25 +325,36 @@ func (b *builderImpl) ListTicketSchemas(ctx context.Context, request *v2.Tickets
 		return nil, fmt.Errorf("error: ticket manager not implemented")
 	}
 
-	out, nextPageToken, annos, err := b.ticketManager.ListTicketSchemas(ctx, &pagination.Token{
-		Size:  int(request.PageSize),
-		Token: request.PageToken,
+	retryer := retry.NewRetryer(ctx, retry.RetryConfig{
+		MaxAttempts:  3,
+		InitialDelay: 15 * time.Second,
+		MaxDelay:     60 * time.Second,
 	})
-	if err != nil {
+
+	for {
+		out, nextPageToken, annos, err := b.ticketManager.ListTicketSchemas(ctx, &pagination.Token{
+			Size:  int(request.PageSize),
+			Token: request.PageToken,
+		})
+		if err == nil {
+			if request.PageToken != "" && request.PageToken == nextPageToken {
+				b.m.RecordTaskFailure(ctx, tt, b.nowFunc().Sub(start))
+				return nil, fmt.Errorf("error: listing ticket schemas failed: next page token is the same as the current page token. this is most likely a connector bug")
+			}
+
+			b.m.RecordTaskSuccess(ctx, tt, b.nowFunc().Sub(start))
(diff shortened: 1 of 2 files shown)


pkg/retry/retry.go
```diff
@@ -52,7 +52,23 @@ func (r *Retryer) ShouldWaitAndRetry(ctx context.Context, err error) bool {
...
+				hasRateLimitData = true
+				break
+			}
+		}
+	}
+
+	// Retry on Unavailable, DeadlineExceeded, ResourceExhausted, or Unknown with rate limit data
+	if status.Code(err) != codes.Unavailable && status.Code(err) != codes.DeadlineExceeded &&
+		status.Code(err) != codes.ResourceExhausted &&
+		!(status.Code(err) == codes.Unknown && hasRateLimitData) {
```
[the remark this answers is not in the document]
laurenleach: I don't think we should change this here, we should add the logic in the connector to return one of these status codes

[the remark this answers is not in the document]
mbburch: Cool, that makes sense. Its in the connector, so good point. Was trying to cover too many bases.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -325,25 +325,36 @@ func (b *builderImpl) ListTicketSchemas(ctx context.Context, request *v2.Tickets
...
+		MaxDelay:     60 * time.Second,
 	})
-	if err != nil {
+
+	for {
+		out, nextPageToken, annos, err := b.ticketManager.ListTicketSchemas(ctx, &pagination.Token{
+			Size:  int(request.PageSize),
+			Token: request.PageToken,
+		})
+		if err == nil {
```
Bencheng21, line 339: Could we do the opposite? 
In general,it's `if err != nil`

mbburch: Leaving as is, since that's the pattern with retries in Validate and Grant.

pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -325,25 +325,36 @@ func (b *builderImpl) ListTicketSchemas(ctx context.Context, request *v2.Tickets
 		return nil, fmt.Errorf("error: ticket manager not implemented")
 	}
 
-	out, nextPageToken, annos, err := b.ticketManager.ListTicketSchemas(ctx, &pagination.Token{
-		Size:  int(request.PageSize),
-		Token: request.PageToken,
+	retryer := retry.NewRetryer(ctx, retry.RetryConfig{
+		MaxAttempts:  0,
+		InitialDelay: 15 * time.Second,
+		MaxDelay:     0,
```
mbburch, line 331: @laurenleach I know there was some concern about having `MaxAttempts` and `MaxDelay` without a limit. I had it this way to test it out. Do we want to stick with it here for now?

laurenleach: I was more worried about Validate, @ggreer do you have any concerns?

PR #420, opened 2025-08-06
Performant existence check for resources

mj-palanker:
adds a new `ResourcesReaderService` function `CheckResourcesExist` which does a single query with many resource Ids to find which  do / don't exist within the c1z. This is entirely motivated by performance issues we are seeing when callers are checking over and over which resources do / don't exist within a c1z.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new API method to check which resource IDs exist, supporting up to 1000 resource IDs per request.
  * Added request and response validation for the new resource existence check.
  * The response now returns only the resource IDs that are confirmed to exist.


mj-palanker:
We don't actually need this for our performance right now as we found a much better approach.
Still might be cool? but closing for now.

PR #433, opened 2025-08-20
Treat 504 gateway timeout as unavailable error.

ggreer:
This causes us to retry requests that return http status 504.


johnallers (approved):
lgtm

PR #436, opened 2025-08-21
[inc-sync-better-resource-events] Added proto for create grant event type

VVill-ga:
## Summary by CodeRabbit

* **New Features**
  * Added a Create Grant event to event feeds so newly created grants appear alongside existing events; annotations are supported on these events.

* **Validation**
  * Create Grant events now require entitlement and principal (resource) fields.
  * Validation returns richer, structured error information and aggregates multiple issues to make troubleshooting clearer.


GowthamiDommety (approved):

proto/c1/connector/v2/event_feed.proto
```diff
@@ -85,6 +86,14 @@ message GrantEvent {
   c1.connector.v2.Grant grant = 1;
 }
 
+message CreateGrantEvent {
+  string entitlement_id = 1;
+  ResourceId resource_id = 2 [(validate.rules).message = {required: true}];
+  string principal_id = 3;
```
ggreer, line 92: Why not have a principal that is a ResourceId?

proto/c1/connector/v2/event_feed.proto
```diff
@@ -85,6 +86,14 @@ message GrantEvent {
   c1.connector.v2.Grant grant = 1;
 }
 
+message CreateGrantEvent {
+  string entitlement_id = 1;
```
ggreer, line 90: Is this the slug?

proto/c1/connector/v2/event_feed.proto
```diff
@@ -85,6 +86,14 @@ message GrantEvent {
...
+  ResourceId resource_id = 2 [(validate.rules).message = {required: true}];
+  string principal_id = 3;
+  ResourceId parent_resource_id = 4;
+  string parent_principal_id = 5;
+}
```
ggreer, line 95: Actually, why not use a Grant message for all this? It has everything we need.

proto/c1/connector/v2/event_feed.proto
owardlaw: in this case the principal is the user_id

proto/c1/connector/v2/event_feed.proto
owardlaw: It has a lot more fields than we want, but the main reason is `Grant` is semi deprecated here and not in other places , we were recommended to not use it.

proto/c1/connector/v2/event_feed.proto
owardlaw: This is a generic field to point us to an entitlement at some point; so what is added to it will be specific to the connector @mj-palanker correct me if i am wrong.

ggreer (approved):

PR #448, opened 2025-08-29
Create ResourceDeleterV2

MarcusGoldschmidt:
- ResourceDeleterV2 enables use the `v2.Resource` to delete, this provide more info and prevents adding context into resource id
- Ran protofmt

## Summary by CodeRabbit

- New Features
  - Delete requests can include an optional parent resource; deletion prefers V2 handlers with fallback to legacy handlers.
- Validation
  - Nested validation added so embedded parent resource IDs are validated and errors are surfaced consistently.
- Tests
  - New unit tests covering V2 delete paths and connector wiring.
- Style
  - Minor proto formatting cleanups (whitespace/newline only).

### DIFF
--- a/pkg/connectorbuilder/connectorbuilder.go
+++ b/pkg/connectorbuilder/connectorbuilder.go
@@ -1225,9 +1225,9 @@ func (b *builderImpl) CreateResource(ctx context.Context, request *v2.CreateReso
 		Create(ctx context.Context, resource *v2.Resource) (*v2.Resource, annotations.Annotations, error)
 	}
 
-	manager, ok := b.resourceManagers[rt]
+	manager, ok := b.resourceManagersV2[rt]
 	if !ok {
-		manager, ok = b.resourceManagersV2[rt]
+		manager, ok = b.resourceManagers[rt]
 	}
 
 	if ok {


pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -96,6 +105,15 @@ type ResourceDeleter interface {
 	Delete(ctx context.Context, resourceId *v2.ResourceId) (annotations.Annotations, error)
 }
 
+// ResourceDeleterV2 extends ResourceSyncer to add capabilities for deleting resources.
+//
+// This is the recommended interface for implementing resource deletion operations in new connectors.
+// It differs from ResourceDeleter by having the resource, not just the id.
+type ResourceDeleterV2 interface {
+	ResourceSyncer
+	Delete(ctx context.Context, resourceId *v2.ResourceId, parentResourceID *v2.ResourceId) (annotations.Annotations, error)
```
Bencheng21, line 114: hm, why not pass in the `v2.Resource` instead of passing in two parameters?

Bencheng21: where is this `parentResourceID` from? 
isn't parentResourceID from `resource.ParentResource`?

MarcusGoldschmidt: Right, but when using `v2.Resource` we need to fill all the data, annotations and profile of it

proto/c1/connector/v2/resource.proto
```diff
@@ -114,6 +115,7 @@ message DeleteResourceResponse {
 
 message DeleteResourceV2Request {
```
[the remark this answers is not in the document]
MarcusGoldschmidt: Yes, we have two, V1 it's when having both create and delete, v2 just have delete

pkg/connectorbuilder/connectorbuilder.go
Bencheng21: Now I see. TIL

laurenleach (approved):

PR #456, opened 2025-09-02
Parallel syncing

jirwin:


pkg/sync/syncer.go
```diff
@@ -60,6 +61,8 @@ type ProgressCounts struct {
 	GrantsProgress       map[string]int
 	LastGrantLog         map[string]time.Time
 	LastActionLog        time.Time
+	mu                   sync.RWMutex // Protect concurrent access to maps
```
ggreer, line 64: The TODO on line 70 can be removed. :)

Also maybe it makes sense to put the progress stuff in progress.go.

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+}
+
+// NewLocalStateContext creates a new local state context for a resource
+func NewLocalStateContext(resourceID *v2.ResourceId) *LocalStateContext {
+	return &LocalStateContext{
+		resourceID:           resourceID,
+		pageToken:            "",
+		needsExpansion:       false,
+		hasExternalResources: false,
+		shouldFetchRelated:   true, // Default to true for backward compatibility
```
ggreer, line 68: I think this defaults to false in sequential syncer. It's only turned on for incremental sync. Also it only works for resource types that implement Get() for incremental sync.

If we're hitting a rate limited API, getting resources one at a time can really hurt performance, so I think this should be off except for incremental syncs.

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+	WorkerCount int
+	// Default bucket for resource types not explicitly configured
+	// If a resource type doesn't specify a sync_bucket, this default will be used
+	DefaultBucket string
+}
+
+// DefaultParallelSyncConfig returns a default configuration
+func DefaultParallelSyncConfig() *ParallelSyncConfig {
+	return &ParallelSyncConfig{
+		WorkerCount:   1,
```
ggreer, line 110: Shouldn't we default to a number higher than 1?

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+func (c *ParallelSyncConfig) WithDefaultBucket(bucket string) *ParallelSyncConfig {
+	c.DefaultBucket = bucket
+	return c
+}
+
+// task represents a unit of work for the parallel syncer
+type task struct {
+	Action       Action
+	ResourceID   string
+	Priority     int              // Higher priority tasks are processed first
```
ggreer, line 133: Maybe it makes sense to have lower numbers be higher priority. That way we don't have priority inflation up to max int.

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+	ResourceType *v2.ResourceType // The resource type for this task
```
ggreer, line 134: Do we need the whole resource type? It's just the sync bucket that we care about, right?

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+		return errors.New("task queue is closed")
+	}
+
+	// Determine the bucket for this task
+	bucket := q.getBucketForTask(t)
+
+	// Create the bucket queue if it doesn't exist
+	if _, exists := q.bucketQueues[bucket]; !exists {
+		queueSize := q.config.WorkerCount * 10
+		q.bucketQueues[bucket] = make(chan *task, queueSize)
```
kans, line 170: Is there recovery for a dropped task?  Some call sites are adding tasks in a loop over a page.  I'd expect to always hit this error and presumably fail the sync with the default worker count of 1.

pkg/sync/parallel_syncer.go
loganintech: IMO whatever codepath gets us to partial sync should require a worker count as an arg that must be >= 2

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+	// Try to get a task from each available bucket in round-robin order
+	// Use a more robust approach that handles the case where a queue becomes empty
+	for _, bucketName := range availableBuckets {
+		queue := q.bucketQueues[bucketName]
+
+		// Double-check the queue still has items before trying to read
+		if len(queue) == 0 {
+			l.Debug("bucket queue became empty", zap.String("bucket", bucketName))
+			continue
+		}
```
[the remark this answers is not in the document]
loganintech: we also just do the exact same check in the select statement down below, I think this can be removed

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+}
+
+// getMapKeys returns the keys of a map as a slice
+func getMapKeys(m map[string]chan *task) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
+}
```
loganintech, line 270: As of `golang 1.25` we have `maps.Keys()` func that does exactly this

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+	}
+	return stats
+}
+
+// Close closes the task queue
+func (q *taskQueue) Close() {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if !q.closed {
```
loganintech, line 307: if closed { return }

for ....

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+func (w *worker) Start() {
+	defer w.wg.Done()
+
+	l := ctxzap.Extract(w.ctx)
+	l.Debug("worker started", zap.Int("worker_id", w.id))
+
+	// Track which bucket this worker is currently working on
+	currentBucket := ""
+	consecutiveFailures := 0
+	maxConsecutiveFailures := 3
```
loganintech, line 351: is this configurable? maybe we make this a const?

pkg/sync/parallel_syncer.go
```diff
@@ -0,0 +1,1706 @@
...
+
+	for {
+		select {
+		case <-w.ctx.Done():
+			l.Debug("worker stopped", zap.Int("worker_id", w.id))
```
loganintech, line 356: Maybe we check `w.ctx.Err()` to see if we exited with an error

PR #463, opened 2025-09-08
New sync type: Resources only.

ggreer:
If syncing WithSkipEntitlementsAndGrants() but not targeted resource IDs, start a new sync of type resource_only. Resume full or resource_only syncs if possible.

## Summary by CodeRabbit

- New Features
  - Explicit sync modes (Full, Partial, Resources Only) and start/resume by selected mode.
  - New option to skip entitlements and grants during full syncs for faster initial resource population.

- Bug Fixes
  - Resume logic now continues the most recent unfinished sync of the same mode when appropriate.
  - Treat Resources Only as partial for cleanup to avoid unintended deletions.
  - Improved validation with clearer errors for invalid sync types and fixed silent DB iteration errors.

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -2761,7 +2761,12 @@ func WithExternalResourceEntitlementIdFilter(entitlementId string) SyncOpt {
 func WithTargetedSyncResourceIDs(resourceIDs []string) SyncOpt {
 	return func(s *syncer) {
 		s.targetedSyncResourceIDs = resourceIDs
-		s.syncType = connectorstore.SyncTypePartial
+		if len(resourceIDs) > 0 {
+			s.syncType = connectorstore.SyncTypePartial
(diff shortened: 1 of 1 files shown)


pkg/sync/syncer.go
```diff
@@ -282,16 +282,23 @@ func (s *syncer) startOrResumeSync(ctx context.Context) (string, bool, error) {
 	var newSync bool
 	var err error
 	if len(s.targetedSyncResourceIDs) == 0 {
-		syncID, newSync, err = s.store.StartSync(ctx)
-		if err != nil {
-			return "", false, err
+		if s.skipEntitlementsAndGrants {
```
kans, line 285: maybe just set the synctype in the condition, and do the rest outside of the branching logic (s.store.StartOrResumeSync...)

kans (approved):

PR #468, opened 2025-09-09
Support Unknown Annotat

kans:
untested!


pkg/lambda/grpc/transport.go
```diff
@@ -10,15 +12,79 @@ import (
...
+						var annObj map[string]json.RawMessage
+						if err := json.Unmarshal(ann, &annObj); err != nil {
+							continue
+						}
+						var typeURL string
+						if t, ok := annObj["@type"]; ok {
+							if err := json.Unmarshal(t, &typeURL); err == nil {
+								// It would be nice to log here, but we have no context.
+								if _, err := protoregistry.GlobalTypes.FindMessageByURL(typeURL); err == nil {
+									wellKnownAnnotations = append(wellKnownAnnotations, ann) // keep only known types
```
kans, line 60: A different approach could be to package the unknown annotation into a known type, so we can log or do something with it later?

pkg/lambda/grpc/transport.go
```diff
@@ -10,15 +12,79 @@ import (
...
+Our requests are small in size, dwarfed by the work of the connector,
+so the performance impact is negligible.
+*/
 func (f *Request) UnmarshalJSON(b []byte) error {
 	f.msg = &pbtransport.Request{}
-	return protojson.Unmarshal(b, f.msg)
+
+	// Parse top-level as raw and surgically filter req.annotations
+	var top map[string]json.RawMessage
+	if err := json.Unmarshal(b, &top); err != nil {
```
ggreer, line 37: Can we try to protojson.Unmarshal() first, then fall back to this code?

PR #474, opened 2025-09-12
Log Errors from Targeted Sync, but don't stop the whole sync

mj-palanker:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sync resilience: when a targeted resource is missing during a partial sync, the system logs a warning, skips that resource, and continues processing remaining actions—reducing aborted syncs and improving overall reliability.


ggreer:
This PR doesn't change behavior (isWarning() already checks for notfound) and it's been sitting around for a while. If there's something else that we want to change in partial syncs, feel free to ping me or open another PR.

PR #477, opened 2025-09-14
Remove sync type restrictions in compaction

mj-palanker:
The current sync type restrictions (in do compaction) will always fail for compaction of more that 2 (because we process them in reverse order).

This PR trivially removes them, but I'll work on another to try and do a smart thing here


mj-palanker:
Merged the better version https://github.com/ConductorOne/baton-sdk/pull/478

PR #496, opened 2025-09-30
[BB-772] Field group config

MarcusGoldschmidt:


laurenleach (approved):

pkg/field/fields.go
```diff
@@ -74,6 +74,9 @@ type SchemaField struct {
 	ConnectorConfig connectorConfig
 
 	WasReExported bool
+
+	// Groups
+	FieldGroups []SchemaFieldGroup
```
Bencheng21, line 79: looks like we need to add it to `ConnectorConfig` ?

Bencheng21 (approved):

PR #507, opened 2025-10-08
file upload string field helper

gontzess:


pkg/field/fields.go
```diff
@@ -213,6 +213,26 @@ func StringField(name string, optional ...fieldOption) SchemaField {
 	return field
 }
 
+func FileUploadField(name string, bonusStrings []string, optional ...fieldOption) SchemaField {
```
gontzess, line 216: could also be fine with a different name like `FileUploadStringField` to represent its still a string type

pkg/field/fields.go
```diff
@@ -213,6 +213,26 @@ func StringField(name string, optional ...fieldOption) SchemaField {
...
+		},
+	}
+
+	for _, o := range optional {
+		field = o(field)
+	}
+
+	return field
+}
+
```
gontzess, line 235: does this need it's own FieldRule config? technically its not a new field type..
https://github.com/ConductorOne/baton-sdk/blob/gontzess/duct-12853-file-upload-field/pkg/field/fields.go#L33-L39

PR #540, opened 2025-11-02
add new option to skip grant expansion in compaction.

mj-palanker:
## Summary by CodeRabbit

* **New Features**
  * Added a new configuration option to control grant expansion behavior in the compaction process. Users can now enable or disable grant expansion based on their needs, with grant expansion enabled by default to maintain backward compatibility.

### DIFF
--- a/pkg/synccompactor/compactor.go
+++ b/pkg/synccompactor/compactor.go
@@ -32,8 +32,9 @@ const (
 )
 
 type Compactor struct {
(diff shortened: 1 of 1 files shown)


pquerna:
this doesn't make sense to me.  you still need to expand grants guys.  ust not at every stage of compaction.  we also need to make expansion fast.  why is it slow?  the SCC part runs in a small number of seconds even for massive graphs.

PR #619, opened 2026-01-06
perf: add streaming iterator API for memory-efficient result processing

arreyder:
## Summary

- Add `rowIterator[T]` type for streaming access to query results without materializing the entire result slice
- Add `streamConnectorObjects()` function that returns the iterator
- Refactor `listConnectorObjects()` to use the streaming API internally, reducing code duplication
- Pre-allocate result slices in `ListSyncRuns()` based on pageSize

The iterator provides:
- `Next()/Value()/Err()` for standard iteration pattern
- `Collect()` to materialize remaining results when a slice is needed
- `Close()` to release database resources
- `NextPageToken()` for pagination

This is useful for:
- Memory-efficient processing of large result sets
- Early termination when a specific item is found
- Chained processing pipelines

## Context

Profile analysis of `temporal_sync` showed `go-sqlite.(*conn).columnText` allocating ~914 MB/sec. While the root cause is in the driver, application-level optimizations provide defense in depth and reduce peak memory usage.

## Test plan

- [x] Existing tests pass
- [x] Verified `listConnectorObjects` behavior unchanged (uses iterator internally)

## Summary by CodeRabbit

* **Refactor**
  * Optimized query result streaming with enhanced memory management to support efficient handling of large result sets.
  * Refined internal memory allocation strategy for improved performance in result collection operations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/pkg/dotc1z/sql_helpers.go
+++ b/pkg/dotc1z/sql_helpers.go
@@ -27,6 +27,101 @@ const bulkPutParallelThreshold = 100
 const insertChunkSize = 1000
 const maxPageSize = 10000
 
+// rowIterator provides streaming access to query results without materializing the entire result slice.
+// Use this for memory-efficient processing of large result sets.
+type rowIterator[T proto.Message] struct {
+	rows          *sql.Rows
+	factory       func() T
+	unmarshalOpts proto.UnmarshalOptions
+	err           error
+	current       T
+	count         uint32
+	pageSize      uint32
+	lastRow       int
+	nextPageToken string
+	done          bool
+}
+
+// Next advances the iterator to the next row.
+// Returns true if a row is available, false when iteration is complete or an error occurred.
+func (it *rowIterator[T]) Next() bool {
+	if it.done || it.err != nil {
+		return false
+	}
+
+	if !it.rows.Next() {
+		it.done = true
+		it.err = it.rows.Err()
+		return false
+	}
+
+	it.count++
+	if it.count > it.pageSize {
+		// We've hit the extra row that indicates more pages exist
+		it.nextPageToken = strconv.Itoa(it.lastRow + 1)
+		it.done = true
+		return false
+	}
+
+	var data sql.RawBytes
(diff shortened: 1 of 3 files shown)


arreyder:
Closing after investigation — the streaming iterator can't deliver its stated benefit given the current architecture, and the incremental refactors that *were* in this PR (cached view sync run, `resolveSyncID` extraction, `parentResourceIdReq` filter) have since landed on main independently.

## Why it doesn't land
The underlying sqlite `*sql.DB` is configured with `SetMaxOpenConns(1)` (see `c1file.go:139`) to keep idle pool connections from holding WAL read locks during checkpoint-truncate. A live `rowIterator` pins that sole connection, so any other DB call issued before `Close()` blocks until iteration finishes. This was verified by a deadlock test — a concurrent `ListResources` blocked for the full 300 ms while the iterator held the connection.

In practice that means:
- **External callers can't safely interleave iteration with other store calls** — it's a silent deadlock footgun.
- **The one in-tree candidate consumer (grant expansion) can't use streaming** — `expander.runAction` fetches descendant grants per source grant during iteration, which would deadlock.
- **No current consumer benefits**: `listConnectorObjects` still goes through `Collect()` which materializes the whole page, so there's zero measured memory saving.

Production profile (`be-temporal-worker`, 3 h aggregate): `listConnectorObjects` is a real hotspot (3.26% of alloc_space, 8.68 GB cum) dominated by `proto.Unmarshal` on `*v2.Grant`. The fix belongs on the unmarshal path, not the slice path.

## Follow-ups
1. Try pooling `*v2.Grant` / proto message reuse on the read path — same shape as #773 (zstd encoder/decoder pool).
2. Longer-term architectural fix for true streaming: dedicated read-only `*sql.DB` (no WAL-lock concern for reads) so iterators don't block writers, then expose server-streaming RPC for `ListGrantsForEntitlement` and convert `fetchGrantsForEntitlementV2` in `ductone/c1` to consume it.

PR #622, opened 2026-01-06
perf(dotc1z): pool zstd encoders and decoders to reduce allocations

arreyder:
## Summary

- Add `sync.Pool` for `zstd.Encoder` and `zstd.Decoder` instances to reduce allocation overhead
- Profiling showed `zstd.ensureHist` allocating **215 MB/min** in temporal_sync due to creating new encoders per `saveC1z` call
- Pool only used when encoder/decoder options match defaults (safe fallback to new instances otherwise)

## Changes

| File | Change |
|------|--------|
| `pkg/dotc1z/pool.go` | NEW - encoder/decoder pool with get/put functions |
| `pkg/dotc1z/pool_test.go` | NEW - unit tests and benchmarks |
| `pkg/dotc1z/file.go` | Modified `saveC1z` to use pooled encoders |
| `pkg/dotc1z/decoder.go` | Modified decoder to use pooled decoders |

## Safety Measures

- Pool only used when options match pool defaults (concurrency, maxMemory)
- Encoders always closed on error paths (not returned to pool in bad state)
- `Reset(nil)` called before returning to pool to release writer/reader references
- Decoder `Reset()` errors handled gracefully (don't return bad decoders to pool)

## Bug Fix (d7c1e91)

Fixed logic error where pool never grew because instances were only returned if they originally came FROM the pool. When pool was empty, new instances were created but never added back.

**Root cause:** `fromPool` tracked *origin*, but return-to-pool decision should be based on *settings compatibility*.

**Fix:** 
- `file.go`: Check `encoderConcurrency == pooledEncoderConcurrency` instead of `fromPool`
- `decoder.go`: Renamed `fromPool` → `poolCompatible`, set based on settings not origin
- Added regression tests `TestPoolGrowsFromSaveC1z` and `TestPoolGrowsFromDecoder`

## Benchmark Results

```
BenchmarkEncoderAllocationOnly/pooled-16         	 1766419	       682 ns/op	     112 B/op	       2 allocs/op
BenchmarkEncoderAllocationOnly/new_each_time-16  	    2296	    604193 ns/op	 2328009 B/op	      30 allocs/op
```

**20,785x reduction** in bytes allocated per encode operation.

### temporal_sync (60s profile)

| Function | Allocation | % of Total |
|----------|------------|------------|
| `zstd.(*fastBase).ensureHist` | **384 MB** | 7.15% |
| `zstd.(*Decoder).startStreamDecoder.func2` | **171 MB** | 3.18% |
| `zstd.encoderOptions.encoder` | 39.17 MB | 0.73% |
| `zstd.(*blockEnc).init` | 32.68 MB | 0.61% |
| **Total zstd** | **664.74 MB** | **12.37%** |

- Duration: 60.64s → **~665 MB/min** from zstd encoder/decoder creation
- `dotc1z.saveC1z` → 288.42 MB cumulative (encoder path)

### temporal_worker (62s profile)

| Function | Allocation | % of Total |
|----------|------------|------------|
| `zstd.(*Decoder).startStreamDecoder.func2` | **806 MB** | 4.43% |
| `zstd.(*Decoder).DecodeAll` | **457 MB** | 2.51% |
| `zstd.(*blockDec).decodeSequences` | 64.51 MB | 0.35% |
| `zstd.(*blockDec).reset` | 53.47 MB | 0.29% |
| **Total zstd** | **3,262.57 MB** | **17.95%** |

- Duration: 62.05s → **~3.2 GB/min** from zstd decoder creation
- Via `protozstd.Unmarshal` → `decompressValue` (decoder path)

### Expected Impact

| Service | Current Allocation | Expected Reduction |
|---------|-------------------|-------------------|
| temporal_sync | ~665 MB/min | ~99% (encoder+decoder pooling) |
| temporal_worker | ~3.2 GB/min | ~99% (decoder pooling) |

## Test plan

- [x] All existing `dotc1z` tests pass
- [x] New pool unit tests (concurrent access, round-trip correctness)
- [x] Regression tests verify pool grows from empty state
- [x] Benchmark demonstrates allocation reduction
- [ ] Deploy to staging and profile to verify improvement

### DIFF
--- a/pkg/dotc1z/decoder.go
+++ b/pkg/dotc1z/decoder.go
@@ -121,6 +121,7 @@ type decoder struct {
 	zd *zstd.Decoder
 
 	decodedBytes uint64
+	fromPool     bool // true if zd came from the pool
 
 	initOnce       sync.Once
 	headerCheckErr error
@@ -141,6 +142,24 @@ func (d *decoder) Read(p []byte) (int, error) {
 			maxMemSize = defaultDecoderMaxMemory
 		}
 
+		// Try to use a pooled decoder if options match the pool's defaults.
+		// Pool decoders use: concurrency=1, lowmem=true, maxMemory=defaultDecoderMaxMemory.
+		usePool := d.o.decoderConcurrency == 1 && maxMemSize == defaultDecoderMaxMemory
+		if usePool {
+			zd, ok := getDecoder()
+			if zd != nil {
+				if err := zd.Reset(d.f); err != nil {
+					// Reset failed, return decoder to pool and fall through to create new one.
+					putDecoder(zd)
+				} else {
+					d.zd = zd
+					d.fromPool = ok
+					return
+				}
+			}
+		}
+
+		// Non-default options or pool unavailable: create new decoder.
 		zstdOpts := []zstd.DOption{
 			zstd.WithDecoderLowmem(true),          // uses lower memory, trading potentially more allocations
 			zstd.WithDecoderMaxMemory(maxMemSize), // sets limit on maximum memory used when decoding stream
@@ -200,7 +219,13 @@ func (d *decoder) Read(p []byte) (int, error) {
 
 func (d *decoder) Close() error {
 	if d.zd != nil {
-		d.zd.Close()
+		if d.fromPool {
+			// Return pooled decoder for reuse.
+			putDecoder(d.zd)
+		} else {
+			d.zd.Close()
+		}
+		d.zd = nil
 	}
 	return nil
 }
--- a/pkg/dotc1z/decoder.go
+++ b/pkg/dotc1z/decoder.go
@@ -120,8 +120,8 @@ type decoder struct {
 	f  io.Reader
 	zd *zstd.Decoder
 
-	decodedBytes uint64
-	fromPool     bool // true if zd came from the pool
+	decodedBytes   uint64
+	poolCompatible bool // true if zd has pool-compatible settings and should be returned to pool
 
 	initOnce       sync.Once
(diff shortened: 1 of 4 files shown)


arreyder:
Closing — superseded, not pursuing.

btipling (approved):
Approving — implementation is solid. Reset(nil) semantics check out against the vendored zstd, error paths correctly exclude bad-state encoders from the pool, Close is idempotent, and the 215 MB/min ensureHist pressure win is meaningful.

Rebased locally on latest main (one minor conflict in decoder.go where main introduced `getMaxMemSize()` — resolved by using the helper; the pool logic is unchanged).

Two items to consider as follow-ups (non-blocking):

1. **Pool currently doesn't engage at production defaults.** `pooledEncoderConcurrency = GOMAXPROCS` (pool.go:16) but the caller default in `c1file.go:242` is `encoderConcurrency = 1`, so the gate at `file.go:158` is false on every regular sync-path call. To make this actually do work in prod, pair with a one-line change to `c1file.go:242` (`1` → `0`, which `saveC1z` already promotes to GOMAXPROCS).

2. **Global-pool test flakiness risk.** `TestPoolGrowsFromSaveC1z` / `TestPoolGrowsFromDecoder` drain the package-global pool and assert exact `fromPool` values. `sync.Pool` can evict via GC between Gets, and any future test added to this package could flip the assertions. Consider swapping the globals via `t.Cleanup` (`oldPool := encoderPool; encoderPool = sync.Pool{}; t.Cleanup(...)`) or dropping the exact-membership invariant.

Not blocking the merge — both are things to address opportunistically.

PR #662, opened 2026-02-03
Partial Uplift: Incremental Grant Expansion for Invalidated Subgraph

kans:
## Goals:
   * Bound incremental work for grant expansion to only affected subgraphs.
   * Prevent broad re-expansion from metadata artifacts.
   * Thorough correctness coverage with targeted performance-regression coverage.

## Summary
1. Implement SQL-driven incremental expansion over old/new sync diffs (attached DB), including:
   * edge delta computation (added/removed expandable edges),
   * changed grant entitlement detection,
   *  staged invalidation + targeted re-expansion.
2. Move expansion/source metadata into SQL columns (expansion, needs_expansion, sources) and make grant diffing compare stable payload fields (data + expansion) to avoid false-positive churn.
3. Fix two major over-expansion regressions:
   * direct-grant source tracking causing broad false diffs,
   * stale needs_expansion flags after full expansion causing later partial runs to re-expand too much.
4. Add chunked dirty marking + pagination-safe processing for large grant sets.
6. Expand coverage substantially:
   * end-to-end incremental correctness matrix,
   * dirty-subgraph and needs-expansion stage tests,
   * migration/backfill tests for grant metadata columns,
   * perf/regression-focused tests and diff/expand benchmarks.

### DIFF
--- a/pkg/connectorstore/connectorstore.go
+++ b/pkg/connectorstore/connectorstore.go
@@ -62,6 +62,9 @@ type InternalWriter interface {
 	ListGrantsInternal(ctx context.Context, opts GrantListOptions) (*InternalGrantListResponse, error)
 	// SetSupportsDiff marks the sync as supporting diff operations.
 	SetSupportsDiff(ctx context.Context, syncID string) error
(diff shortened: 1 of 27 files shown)


kans:
superceded 5 times over by more recent work

PR #703, opened 2026-02-24
Remove NeedsExpansion from syncer state.

ggreer:
Now that grant expansion annotations are moved into their own column, we don't need to keep track of this. It's cheap to check whether any grants need expanding, so always do it unless the syncer has been told otherwise.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal synchronization logic and state management for improved efficiency.


kans:
We did this in pebble (or don't need to anymore).

PR #708, opened 2026-02-26
Set SQLite temp_store_directory to database parent directory

arreyder:
SQLite writes temp files (sort spill, temp tables, temp indexes) to the OS temp directory by default. After the removal of temp_store=MEMORY, these files land in /tmp outside the managed c1z directory tree. If the process is killed before db.Close(), they leak.

By setting temp_store_directory to the database file's parent directory, cleanupDbDir's os.RemoveAll catches any leftover SQLite temp files.

## Summary by CodeRabbit

* **Improvements**
  * Optimized database temporary file storage location for enhanced performance and system resource management.

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -440,6 +440,19 @@ func (c *C1File) init(ctx context.Context) error {
 		return err
 	}
 
+	// Direct SQLite temp files into the same directory as the database file.
+	// Without this, SQLite writes temp files (sort spill, temp tables, temp
+	// indexes) to the OS temp directory (/tmp), outside the managed c1z
+	// directory tree. Those files are only cleaned up on db.Close(); if the
+	// process is killed they leak. By pointing temp_store_directory at the
+	// db's parent directory, cleanupDbDir's RemoveAll catches them.
+	dbDir := filepath.Dir(c.dbFilePath)
+	//nolint:gosec // dbDir is derived from our own dbFilePath, not user input.
+	_, err = c.db.ExecContext(ctx, fmt.Sprintf("PRAGMA temp_store_directory = '%s'", dbDir))
+	if err != nil {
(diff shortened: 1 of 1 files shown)


jerenkrantz (approved):

pkg/dotc1z/c1file.go
```diff
@@ -440,6 +440,18 @@ func (c *C1File) init(ctx context.Context) error {
...
 	}
 
+	// Direct SQLite temp files into the same directory as the database file.
+	// Without this, SQLite writes temp files (sort spill, temp tables, temp
+	// indexes) to the OS temp directory (/tmp), outside the managed c1z
+	// directory tree. Those files are only cleaned up on db.Close(); if the
+	// process is killed they leak. By pointing temp_store_directory at the
+	// db's parent directory, cleanupDbDir's RemoveAll catches them.
+	dbDir := filepath.Dir(c.dbFilePath)
+	_, err = c.db.ExecContext(ctx, fmt.Sprintf("PRAGMA temp_store_directory = '%s'", dbDir))
```
ggreer, line 674: SQLite docs say this pragma is deprecated: https://sqlite.org/pragma.html#pragma_temp_store_directory Are we sure we want to use this? I think our current solution of setting the SQLITE_TMPDIR env var in prod is the best option. That would also prevent us from leaking temp files when using baton-sdk elsewhere (such as service mode connectors).

arreyder:
Closing as stale. The stated premise — that temp files leak to /tmp 'after the removal of temp_store=MEMORY' — no longer matches main: there is no temp_store setting in c1file.go at all today, so SQLite uses its compiled default. The change still applies cleanly and is low-risk, but the original justification is from a Feb-2026 code state and it's sat 3+ months without review demand. Happy to reopen with a fresh rationale if actual /tmp temp-file leakage on SIGTERM is confirmed in current prod.

PR #737, opened 2026-03-15
Add periodic WAL checkpoint and WAL size logging

arreyder:
## Summary
Two changes to prevent unbounded WAL growth and add visibility:

1. **Periodic PASSIVE checkpoint** after each `executeChunkedInsert` commit
2. **WAL stats as span attributes** on checkpoint spans for APM trace analytics

## Problem
50GB+ WAL files observed in prod. SQLite's auto-checkpoint (every 1000 WAL pages) silently fails when the Go context is cancelled (Temporal activity deadline). The WAL grows unboundedly across transactions with no reclaim until `Close()`, which then faces a massive WAL to checkpoint.

## How it works

**PASSIVE checkpoint** (`CheckpointWAL`): Non-blocking — checkpoints as many frames as possible without waiting for writers. Called after each chunked insert commit to keep the WAL bounded. Errors are non-fatal (the final TRUNCATE checkpoint at Close handles the rest).

**WAL stats on spans** (not logs): Both PASSIVE and TRUNCATE checkpoint paths record attributes on their OTel spans:
- `wal.size_bytes` — actual WAL file size on disk
- `wal.checkpoint.log_pages` — total pages in WAL
- `wal.checkpoint.checkpointed_pages` — pages successfully checkpointed
- `wal.checkpoint.busy` — whether checkpoint was blocked

These are queryable in Datadog APM trace analytics without adding log volume (~45K log lines/hour avoided). The only log line retained is a `Warn` for incomplete TRUNCATE checkpoints, which are actionable and infrequent.

## Risk
Low — PASSIVE checkpoint is explicitly non-blocking. It's the lightest checkpoint mode SQLite offers. If the WAL has no uncheckpointed frames, it's a no-op. The only cost is one span + one `PRAGMA` call per chunked insert.

## Test plan
- [ ] Build passes
- [ ] Existing dotc1z tests pass
- [ ] Deploy via c1 vendor bump, verify WAL size stays bounded
- [ ] Query `wal.size_bytes` in APM spans to monitor WAL growth

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/doug-martin/goqu/v9"
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -424,7 +425,7 @@ func (c *C1File) Close(ctx context.Context) error {
 	return nil
 }
 
-// CheckpointWAL runs a non-blocking PASSIVE checkpoint and logs WAL stats.
+// CheckpointWAL runs a non-blocking PASSIVE checkpoint and records WAL stats on a span.
 // PASSIVE checkpoints as many frames as possible without blocking writers,
 // preventing unbounded WAL growth between transactions.
 func (c *C1File) CheckpointWAL(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


jerenkrantz (approved):

arreyder:
Rebased onto current main (was 37 commits behind) and re-requested review — the original approval was on `230d65e1` before @kans's feedback triggered a significant scope change (dropped the PASSIVE post-commit checkpoint; kept only the `truncateWAL` observability). Also adds relevance now that WAL is on the prod `Close()` path (see c1file.go L332-L409, which explicitly refuses to save if WAL isn't empty — the exact failure mode we're tracking in the c1z corruption project).

arreyder:
Closing as obsolete. The unbounded-WAL problem this targeted was eliminated when the default journal mode switched to TRUNCATE in #807 (merged 2026-05-07) — WAL is now opt-in, so there's no growing sidecar to bound by default. The current diff also only adds OTel span attributes to truncateWAL rather than the per-chunk PASSIVE checkpoint described in the original motivation, and WAL-size visibility is now covered by CurrentDBSizeBytes. If APM visibility for opt-in-WAL callers is still wanted, that span-attribute hunk can be cherry-picked into a fresh small PR.

PR #738, opened 2026-03-16
Checkpoint sync state on all cancellation types, not just DeadlineExceeded

arreyder:
## Summary

- Checkpoint sync state when the context is cancelled for any reason (SIGTERM, etc.), not just `DeadlineExceeded`
- Use a background context with 30s timeout for the checkpoint since the caller's context is already cancelled
- Broaden cleanup error handler to treat any context cancellation as `ErrSyncNotComplete`

## Problem

When a sync worker receives SIGTERM, the sync context is cancelled but `context.Cause()` returns a cancellation error, not `DeadlineExceeded`. The `default:` branch in both `sequentialSync` and `parallelSync` exited without checkpointing, losing all progress since the last periodic checkpoint. This forced a full re-sync of those pages on the next attempt.

## Changes

- `pkg/sync/parallel_syncer.go`: Both `sequentialSync` and `parallelSync` now checkpoint in the `default:` cancellation case using `context.WithTimeout(context.Background(), 30*time.Second)`
- `pkg/sync/syncer.go`: Cleanup error handler uses `runCtx.Err() != nil` instead of `errors.Is(err, context.DeadlineExceeded)` to catch all cancellation types

## Notes

This is a rewrite of #694 against current main. The original PR couldn't be rebased because the sync loop was refactored into `sequentialSync`/`parallelSync` methods. Supersedes #694.

Note: `pkg/sync` has a pre-existing test timeout on main (not introduced by this PR).

### DIFF
--- a/pkg/sync/parallel_syncer.go
+++ b/pkg/sync/parallel_syncer.go
@@ -60,7 +60,7 @@ func (s *syncer) sequentialSync(
 				if checkpointErr := s.Checkpoint(checkpointCtx, true); checkpointErr != nil {
 					l.Error("error checkpointing before exiting cancelled sync", zap.Error(checkpointErr))
 				}
-				return warnings, err
+				return warnings, errors.Join(err, ErrSyncNotComplete)
 			}
 		default:
 		}
@@ -291,7 +291,7 @@ func (s *syncer) parallelSync(
 				if checkpointErr := s.Checkpoint(checkpointCtx, true); checkpointErr != nil {
 					l.Error("error checkpointing before exiting cancelled sync", zap.Error(checkpointErr))
 				}
-				return warnings, err
+				return warnings, errors.Join(err, ErrSyncNotComplete)
 			}
 		default:
 		}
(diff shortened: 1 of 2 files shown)


arreyder:
Good catch — fixed both issues in ad00d761:

1. **Cleanup resumability (syncer.go:456)**: Narrowed from `runCtx.Err() != nil` to `errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)`. Only treats cleanup as resumable when the error is actually context-caused.

2. **Preservability (parallel_syncer.go)**: Both `default` cancellation cases now return `errors.Join(err, ErrSyncNotComplete)` instead of raw `err`. This makes `IsSyncPreservable` return true so the checkpointed c1z is actually kept and the sync can resume.

dstokes (approved):

ennyjfrick (approved):

arreyder:
Closing during status cleanup — not being pursued. Reopen if revisited.

PR #753, opened 2026-04-06
CXH-1277: handle weighted rate limit header format

aldevv:
RFC format: https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#name-ratelimit-limit
Zscaler returns rate limit headers in the weighted format from the RateLimit header RFC draft (e.g. `500, 500;w=1, 20000;w=60`), which caused `strconv.ParseInt` to fail and propagate an error that masked the real 500 response.

- Add `normalizeRateLimitValue` to extract the leading integer from weighted header values
- Apply normalization in `ExtractRateLimitData` for both limit and remaining headers

### DIFF
--- a/pkg/ratelimit/http.go
+++ b/pkg/ratelimit/http.go
@@ -3,6 +3,7 @@ package ratelimit
 import (
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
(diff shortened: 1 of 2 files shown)


pkg/ratelimit/http.go
```diff
@@ -63,6 +64,19 @@ func parseTime(timeStr string) (time.Time, error) {
 	return t, nil
 }
 
+// normalizeRateLimitValue extracts the leading integer from a rate limit header
+// that may use the weighted format from the RateLimit header RFC draft
+// (e.g. "500, 500;w=1, 20000;w=60" → "500", "100;w=60" → "100").
```
aldevv:
```suggestion
// (e.g. "500, 500;w=1, 20000;w=60" → "500", "100;w=60" → "100").
// docs: https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#name-ratelimit-limit
```

pkg/ratelimit/http.go
```diff
@@ -63,6 +64,20 @@ func parseTime(timeStr string) (time.Time, error) {
...
+// docs: https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#name-ratelimit-limit
+func normalizeRateLimitValue(s string) string {
+	s = strings.TrimSpace(s)
+	for i := 0; i < len(s); i++ {
+		if s[i] < '0' || s[i] > '9' {
+			return s[:i]
+		}
+	}
+	return s
```
btipling, line 78: I'd return "0" if s is the empty string

btipling (approved):

PR #760, opened 2026-04-09
Harden c1z save paths and add diagnostics

gontzess:
## Why

The c1z corruption investigation narrowed the remaining local risk to the baton-sdk persistence boundary: creating the temp c1z, copying it into place, and distinguishing whether corruption is introduced before or after the local save completes.

The failure signature was not a bad header. The saved file could pass the `C1ZF\x00` magic check and still fail later with `corrupt stream, did not find end of stream`, which points at persistence or finalize problems rather than simple header validation.

The investigation also ruled out a few earlier theories:

- not a save-order race: the sequential sync path closes the live `dotc1z.C1File` before the outer save/copy path runs
- not the parallel timeout classification bug for this incident: this connector was running with parallel worker count `0`
- timing still pointed at the close/save boundary: timeout was logged at `2026-04-08 11:42:28Z`, local c1z save at `2026-04-08 11:55:22Z`, upload at `2026-04-08 11:56:11Z`, and the next resume/load then failed with end-of-stream corruption

That gap before local save is consistent with large-file finalize/compress/save work, so this PR hardens the local save path and adds source-versus-destination evidence there.

## What this changes

- harden `saveC1z()` and `localManager.SaveC1Z()` with safer temp-file handling and atomic publish helpers
- keep the explicit sqlite db sync before compression
- add size and lightweight checksum logging around the temp, staging, and final saved c1z files
- keep Unix publish durability as `rename + parent-directory fsync`
- use Windows `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING|MOVEFILE_WRITE_THROUGH)` instead of trying to fsync a directory handle after rename
- add focused tests for successful save, failure cleanup, and preserving the previous destination on save failure

## Why this helps

After this change, the local save path should no longer leave behind a partially written destination c1z, and the added logging gives enough evidence to tell whether any remaining corruption is already present immediately after local save or appears later in the flow.

The save-path checks are also tighter now:

- same-directory staging is explicit because it is required for atomic rename
- the local copy gate uses a cheap checksum rather than a crypto hash while still catching same-length corruption
- Unix keeps the parent-directory sync needed for rename durability
- Windows now uses a native write-through publish step instead of a false post-rename failure path
- the db sync close path is a little safer for future edits

Together with the paired c1 containment PR, this closes the partial-write hypothesis and sharpens any next investigation toward upstream generation or downstream post-save handling.


pkg/dotc1z/manager/local/local.go
```diff
@@ -161,34 +164,113 @@ func (l *localManager) SaveC1Z(ctx context.Context) error {
...
-	size, err := io.Copy(dstFile, tmpFile)
+	// Stage in the destination directory so the final rename stays on the same
+	// filesystem and remains atomic.
+	destDir := filepath.Dir(l.filePath)
+	dstFile, err := os.CreateTemp(destDir, filepath.Base(l.filePath)+".tmp-*")
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create staging file: %w", err)
+	}
+	stagingPath := dstFile.Name()
```
kans: Call this dstFilePath, or call the other thing stagingFile - staging is more descriptive.

pkg/dotc1z/sync_parent_dir_unix.go
```diff
@@ -0,0 +1,24 @@
...
+package dotc1z
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+func SyncParentDir(path string) error {
+	dir := filepath.Dir(path)
```
kans, line 14: Why fsync the dir instead of the one file we care about?

edit: we fsync teh dir for the `rename`!  We have a different set of APIs to use on Windows.  This is called by the local manager, so we should make it work for Windows too.

pkg/dotc1z/c1file.go
```diff
@@ -387,10 +387,34 @@ func (c *C1File) Close(ctx context.Context) error {
 			return cleanupDbDir(c.dbFilePath, fmt.Errorf("c1z: WAL file not empty after close (size=%d) - refusing to save incomplete data", walInfo.Size()))
 		}
 
+		dbFile, err := os.OpenFile(c.dbFilePath, os.O_RDWR, 0)
```
kans, line 390: Can we defer close this for future proofing?

pkg/dotc1z/manager/local/local.go
```diff
@@ -161,34 +164,117 @@ func (l *localManager) SaveC1Z(ctx context.Context) error {
...
+			if closeErr := dstFile.Close(); closeErr != nil {
+				log.Warn("failed to close staging file", zap.Error(closeErr))
+			}
+		}
+		if !success {
+			_ = os.Remove(stagingPath)
+		}
+	}()
+
+	sourceHash := sha256.New()
```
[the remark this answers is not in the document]
gontzess: did not reduce it to size-only because that would miss same-length corruption, but did make this check more cheap

pkg/dotc1z/sync_parent_dir_windows.go
```diff
@@ -0,0 +1,38 @@
...
+package dotc1z
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"golang.org/x/sys/windows"
+)
+
+func SyncParentDir(path string) error {
```
kans: Cursor says: `I checked the Windows API docs while reviewing this: FlushFileBuffers requires a writable file handle, and this code opens a directory handle with GENERIC_READ. That means the new Windows save path is very likely to return an error every time it reaches SyncParentDir. Because the call happens after os.Rename, the function can report failure even though the destination file was already replaced. The same problem also affects pkg/dotc1z/manager/local/local.go.`.

gontzess: Windows: keep file `Sync()` -> publish with `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING|MOVEFILE_WRITE_THROUGH)`

gontzess: ci passes now too

PR #762, opened 2026-04-10
Classify transient network errors across lambda and uhttp

gontzess:
## Why

Some connector SDK clients return raw Go network errors instead of gRPC status errors. Today those transient failures can fall through Baton SDK as `codes.Unknown` instead of a retryable status.

There are two places this shows up:

- the lambda gRPC server path, where non-status application errors can be converted to `Unknown`
- the `uhttp` transport path, where SDK clients that reuse the transport directly, such as through `oauth2`, can bypass the retryable wrapping that already exists in the higher-level HTTP client wrapper

That makes transient resets and timeouts look non-retryable to downstream sync handling, which can cause long-running syncs to drop current progress instead of resuming from the latest artifact state.

## What This Changes

- classify wrapped `ECONNRESET` and `unexpected EOF` as `codes.Unavailable` before non-status lambda errors fall through to `Unknown`
- classify URL timeouts as `codes.DeadlineExceeded`
- classify temporary URL errors as `codes.Unavailable`
- add shared transient network wrapping for the `uhttp` transport path so transport-based SDK clients get the same retryable behavior as callers that use the higher-level wrapper
- add focused tests for the lambda classifier and the `uhttp` transport path

This keeps transient network handling consistent across Baton SDK call paths instead of only the wrapper-based HTTP path.

## Validation

- `GOCACHE=/tmp/gocache GOLANGCI_LINT_CACHE=/tmp/golangci-lint-cache golangci-lint run ./pkg/uhttp ./pkg/lambda/grpc`
- `GOCACHE=/tmp/gocache go test -timeout 30s ./pkg/lambda/grpc ./pkg/uhttp`

### DIFF
--- a/pkg/uhttp/errors.go
+++ b/pkg/uhttp/errors.go
@@ -39,5 +39,5 @@ func wrapTransientNetworkError(err error) error {
 		return status.Error(codes.DeadlineExceeded, "request timeout")
 	}
 
-	return nil
+	return err
 }
--- a/pkg/uhttp/errors.go
+++ b/pkg/uhttp/errors.go
@@ -1,0 +1,43 @@
+package uhttp
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/url"
+	"syscall"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// wrapTransientNetworkError mirrors Baton HTTP retry classification for callers
+// that use the transport directly, such as oauth2-backed SDK clients.
+func wrapTransientNetworkError(err error) error {
+	if err == nil {
+		return nil
+	}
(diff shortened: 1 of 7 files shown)


btipling (approved):

pkg/uhttp/errors.go
```diff
@@ -0,0 +1,43 @@
...
+		if urlErr.Temporary() {
+			return WrapErrors(codes.Unavailable, fmt.Sprintf("temporary error: %v", urlErr.URL), urlErr)
+		}
+	}
+
+	if errors.Is(err, context.DeadlineExceeded) {
+		return status.Error(codes.DeadlineExceeded, "request timeout")
+	}
+
+	return nil
```
kans: Its probably better to return the error, so that we don't put this functoinality on the caller

kans (approved):

PR #765, opened 2026-04-14
Add entitlement exclusion groups and swap provisioning

pquerna:
## Summary

- Adds `EntitlementExclusionGroup` annotation so connectors can declare mutually exclusive entitlements (at most one grant per principal per resource per group)
- Adds `SwapService` gRPC service for atomic entitlement transitions, with SDK fallback (sequential Revoke+Grant with retry) when connectors don't implement native swap
- Adds `SwapPartialFailure` structured error detail for partial failure recovery
- Adds `CAPABILITY_ATOMIC_SWAP` per-resource-type capability flag
- Adds builder helpers: `WithExclusionGroup`, `WithExclusionGroupOrder`, `WithExclusionGroupDefault`
- Validates at most one exclusion group annotation per entitlement

## Test plan

- [x] Entitlement builder helper tests (annotation round-trip)
- [x] Native swap happy path
- [x] Fallback swap happy path (sequential Revoke+Grant)
- [x] Fallback partial failure (SwapPartialFailure error detail extraction)
- [x] Resource type mismatch validation
- [x] No provisioner configured error
- [x] CAPABILITY_ATOMIC_SWAP detection (present/absent per resource type)
- [x] Exclusion group validation (zero, single, duplicate, mixed batch)
- [x] All existing tests pass


pkg/connectorbuilder/resource_provisioner.go
```diff
@@ -173,5 +191,127 @@ func (b *builder) addProvisioner(_ context.Context, typeId string, in interface{
...
+	defer span.End()
+
+	start := b.nowFunc()
+	tt := tasks.SwapType
+	l := ctxzap.Extract(ctx)
+
+	revokeRT := request.GetRevokeGrant().GetEntitlement().GetResource().GetId().GetResourceType()
+	grantRT := request.GetGrantEntitlement().GetResource().GetId().GetResourceType()
+
+	if revokeRT != grantRT {
```
kans: I think this is too opinionated for the SDK in the case the API isn't the source of exclusion.  The grant and revoke sides can be configured for direct/indirect access respectively.  We can warn instead.

pkg/connectorbuilder/resource_provisioner.go
```diff
@@ -173,5 +191,127 @@ func (b *builder) addProvisioner(_ context.Context, typeId string, in interface{
...
+
+	// Try native swap first
+	if swapper, ok := b.swapProvisioners[grantRT]; ok {
+		retryer := retry.NewRetryer(ctx, retry.RetryConfig{
+			MaxAttempts:  3,
+			InitialDelay: 15 * time.Second,
+			MaxDelay:     60 * time.Second,
+		})
+		for {
+			grants, annos, err := swapper.Swap(ctx, request.GetRevokeGrant(), request.GetGrantEntitlement())
```
kans: The connector implementation is almost certainly going to have the same fallback revoke / grant loop in it (ie, the same thing that is below this block), mixed in with a bunch of bespoke imperative logic.

kans:
@pquerna 
1. Does exclusion need to be backwards propagated post expansion, or are we handling that on the C1 side?

kans (approved):

PR #766, opened 2026-04-14
Add lambda runtime state and bundle fetch to SDK

gontzess:
## Why
Lambda-managed connectors need to refresh runtime state, config, and runtime bundle bytes without waiting for a full redeploy. The SDK is the natural seam for that because it already owns the config client, the lambda startup path, and the warm refresh callback.

## What this changes
- adds runtime-state RPC support to the SDK config client
- carries runtime bundle bytes through the lambda config snapshot
- fetches runtime bundles during both cold start and warm refresh
- preserves bootstrap config across reload so the rebuilt connector uses the same host settings
- adds focused tests around lambda config hydration and startup behavior

## Validation
This branch is part of the D2 hotload test stack. With the matching runtime-framework branch and the Jira test connector branch, we proved:
- warm refresh from a fetched runtime bundle archive
- cold start from a fetched runtime bundle archive
- fail-closed behavior when the runtime bundle is missing or invalid

## Notes
This PR is intended to land with the matching runtime-framework work. The D2 bundle insertion helper lives on the C1 side of the stack and is part of the recorded test path.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"strings"
 
+	v1 "github.com/conductorone/baton-sdk/pb/c1/connectorapi/baton/v1"
 	"github.com/conductorone/baton-sdk/pkg/connectorbuilder"
 	"github.com/conductorone/baton-sdk/pkg/field"
 	"github.com/conductorone/baton-sdk/pkg/types"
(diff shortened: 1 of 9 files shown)


pkg/cli/cli.go
```diff
@@ -16,13 +17,84 @@ import (
...
-	SelectedAuthMethod  string
-	SyncResourceTypeIDs []string
+	SessionStore          sessions.SessionStore
+	TokenSource           oauth2.TokenSource
+	SelectedAuthMethod    string
+	SyncResourceTypeIDs   []string
+	RuntimeState          *v1.RuntimeState
+	RuntimeBundle         []byte
+	RuntimePolicy         []byte
+	BootstrapConfig       map[string]any
```
kans, line 39: What is this?

gontzess: gave this a bit of a rework and annotations now

gontzess:
going with https://github.com/ConductorOne/baton-sdk/pull/785

PR #770, opened 2026-04-15
Use lambda client secret

laurenleach:


kans (approved):

kans:
@laurenleach a test for this would be nice

PR #771, opened 2026-04-16
fix: normalize protojson whitespace in capabilities output

russellhaering:
## Summary
- `protojson` intentionally injects non-deterministic whitespace via `detrand.Bool()`, causing the `capabilities` command to produce cosmetically different JSON across runs (single vs double space after colons)
- This creates noise commits in CI workflows that capture `baton_capabilities.json`
- Fix: round-trip through `json.Indent` to normalize all whitespace from the token stream, preserving protojson's key order while producing byte-identical output every time
- This is the [approach recommended by the protobuf maintainers](https://github.com/golang/protobuf/issues/1121)

## Test plan
- [ ] Run `capabilities` command multiple times and verify output is byte-identical: `for i in $(seq 20); do ./connector capabilities 2>/dev/null | md5sum; done`
- [ ] Verify JSON output is valid and parses correctly
- [ ] Verify no semantic changes vs current output (only whitespace normalization)

### DIFF
--- a/pkg/cli/commands.go
+++ b/pkg/cli/commands.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


gontzess (approved):
LGTM

PR #773, opened 2026-04-18
perf(dotc1z): pool zstd encoders and decoders to reduce allocations

arreyder:
Reviving #622 (closed, not merged). This re-opens the same branch with btipling's two non-blocking follow-ups from the prior review folded in.

## Summary

- Add `sync.Pool` for `zstd.Encoder` and `zstd.Decoder` to eliminate per-call allocation overhead on the dotc1z hot path.
- Profiling showed `zstd.ensureHist` allocating **~215 MB/min** in `temporal_sync` from encoder churn, and the decoder side allocating **~3.2 GB/min** in `temporal_worker`.
- Pool only engages when options match pool defaults; callers with custom concurrency/memory settings fall back to fresh instances.

## Changes

| File | Change |
|---|---|
| `pkg/dotc1z/pool.go` | NEW — encoder/decoder pool with `get*`/`put*` helpers |
| `pkg/dotc1z/pool_test.go` | NEW — unit tests, concurrent stress, round-trip, benchmarks |
| `pkg/dotc1z/file.go` | `saveC1z` uses pooled encoder when concurrency matches |
| `pkg/dotc1z/decoder.go` | decoder uses pooled decoder when options match |

## Follow-ups from prior review (new in this revival)

1. **Pool now engages at the production default.** In the prior PR, `pooledEncoderConcurrency = GOMAXPROCS` but the callers default to `encoderConcurrency = 1`, so the gate at `file.go:158` was always false and the pool carried zero traffic in prod. Pinning `pooledEncoderConcurrency = 1` makes the pool match the actual sync-path default. The alternative — flipping callers to `0` (→GOMAXPROCS) — was deliberately avoided because klauspost/compress's multi-threaded encoder has theoretical race concerns.

2. **Pool-growth tests are now hermetic.** `TestPoolGrowsFromSaveC1z`, `TestPoolGrowsFromDecoder`, `TestEncoderPool`, and `TestDecoderPool` now swap the package-global `sync.Pool` via `t.Cleanup`, so GC eviction and cross-test ordering can't flake the exact-membership assertions. The package globals became pointer-valued to allow the swap.

## Safety Measures

- Pool only used when options match pool defaults (concurrency, maxMemory).
- Encoders always closed on error paths (not returned to pool in bad state).
- `Reset(nil)` called before returning to pool to release writer/reader references.
- Decoder `Reset()` errors handled gracefully (don't return bad decoders to pool).

## Benchmark (apples-to-apples at concurrency=1)

```
BenchmarkEncoderAllocationOnly/pooled-16          1,202,396     1000 ns/op     116 B/op     2 allocs/op
BenchmarkEncoderAllocationOnly/new_each_time-16       4,809   338275 ns/op 1758312 B/op    19 allocs/op
```

~15,000x reduction in bytes allocated per encode.

## Production Profile Context

- `temporal_sync` (60s profile): `zstd.(*fastBase).ensureHist` ≈ 384 MB, total zstd allocs ≈ 665 MB (12.4% of total).
- `temporal_worker` (62s profile): decoder-side allocs ≈ 3.26 GB (17.9% of total) via `protozstd.Unmarshal` → `decompressValue`.

## Test plan

- [x] All existing `dotc1z` tests pass
- [x] Pool unit tests (concurrent access, round-trip correctness)
- [x] Regression tests verify pool grows from empty state
- [x] Isolated-pool tests are no longer sensitive to cross-test ordering
- [x] Benchmark demonstrates allocation reduction at production concurrency
- [ ] Deploy to staging and profile to verify improvement

Prior approval: @btipling on #622.

### DIFF
--- a/pkg/dotc1z/pool.go
+++ b/pkg/dotc1z/pool.go
@@ -1,0 +1,88 @@
+package dotc1z
+
+import (
+	"runtime"
+	"sync"
+
+	"github.com/klauspost/compress/zstd"
+)
+
+// encoderPool manages reusable zstd.Encoder instances to reduce allocation overhead.
+// All pooled encoders are configured with GOMAXPROCS concurrency.
+var encoderPool sync.Pool
+
+// pooledEncoderConcurrency is the concurrency level used for pooled encoders.
+// Set at package init to GOMAXPROCS to match the default behavior.
+var pooledEncoderConcurrency = runtime.GOMAXPROCS(0)
+
+// getEncoder retrieves a zstd encoder from the pool or creates a new one.
+// The returned encoder is NOT bound to any writer - call Reset(w) before use.
+// Returns the encoder and a boolean indicating if it came from the pool.
+func getEncoder() (*zstd.Encoder, bool) {
+	if enc, ok := encoderPool.Get().(*zstd.Encoder); ok && enc != nil {
+		return enc, true
+	}
+
+	// Create new encoder with default concurrency.
+	// This should not fail with valid options, but handle it gracefully.
+	enc, err := zstd.NewWriter(nil,
+		zstd.WithEncoderConcurrency(pooledEncoderConcurrency),
+	)
+	if err != nil {
+		// Fallback: return nil and let caller create encoder with their options
+		return nil, false
+	}
+	return enc, false
+}
+
+// putEncoder returns a zstd encoder to the pool for reuse.
+// The encoder is reset to release any reference to the previous writer.
+// Encoders should be in a clean state (Close() called) before returning.
+func putEncoder(enc *zstd.Encoder) {
+	if enc == nil {
+		return
+	}
+	// Reset to nil writer to release reference to previous output.
+	// This is safe even if the encoder was already closed.
+	enc.Reset(nil)
+	encoderPool.Put(enc)
+}
+
+// decoderPool manages reusable zstd.Decoder instances to reduce allocation overhead.
+// All pooled decoders are configured with concurrency=1 (single-threaded) and low memory mode.
+var decoderPool sync.Pool
+
+// getDecoder retrieves a zstd decoder from the pool or creates a new one.
+// The returned decoder is NOT bound to any reader - call Reset(r) before use.
+// Returns the decoder and a boolean indicating if it came from the pool.
+func getDecoder() (*zstd.Decoder, bool) {
+	if dec, ok := decoderPool.Get().(*zstd.Decoder); ok && dec != nil {
+		return dec, true
+	}
+
+	// Create new decoder with default options matching decoder.go defaults.
+	dec, err := zstd.NewReader(nil,
+		zstd.WithDecoderConcurrency(1),
+		zstd.WithDecoderLowmem(true),
+		zstd.WithDecoderMaxMemory(defaultDecoderMaxMemory),
+	)
+	if err != nil {
+		// Fallback: return nil and let caller create decoder with their options
+		return nil, false
+	}
+	return dec, false
+}
+
+// putDecoder returns a zstd decoder to the pool for reuse.
+// The decoder is reset to release any reference to the previous reader.
+func putDecoder(dec *zstd.Decoder) {
+	if dec == nil {
+		return
+	}
+	// Reset to nil reader to release reference to previous input.
+	// If Reset fails (bad state), don't return to pool.
+	if err := dec.Reset(nil); err != nil {
+		return
+	}
+	decoderPool.Put(dec)
+}
--- a/pkg/dotc1z/pool.go
+++ b/pkg/dotc1z/pool.go
@@ -15,7 +15,8 @@ var encoderPool = &sync.Pool{}
 
 // pooledEncoderConcurrency is the concurrency level used for pooled encoders.
 // Must match the c1zOptions / C1File default (1) so the pool actually engages
-// on the sync hot path.
+// on the sync hot path. Note: pkg/synccompactor/compactor.go passes
+// WithEncoderConcurrency(0) (GOMAXPROCS) and intentionally bypasses the pool.
 const pooledEncoderConcurrency = 1
 
 // getEncoder retrieves a zstd encoder from the pool or creates a new one.
--- a/pkg/dotc1z/pool.go
+++ b/pkg/dotc1z/pool.go
@@ -1,10 +1,18 @@
 package dotc1z
 
 import (
+	"os"
 	"sync"
 
 	"github.com/klauspost/compress/zstd"
 )
+
+// poolDisabled is an operational kill-switch. Set BATON_ZSTD_POOL_DISABLE=1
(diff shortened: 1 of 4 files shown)


btipling (approved):
Approving — the pool now engages at production defaults (pinned `pooledEncoderConcurrency = 1` to match `c1file.go:242`), and the test isolation via `withIsolatedPools` + pointer-valued pools resolves the flakiness from the prior round. Both critical items from the earlier review are cleanly addressed.

Two optional polish suggestions left as inline comments with applyable diffs.

## Suggestion: add an env kill-switch before merge (not blocking)

An env kill-switch would let ops flip the pool off in prod without a revert + re-release cycle:

```go
// pool.go
var poolDisabled = os.Getenv("BATON_ZSTD_POOL_DISABLE") == "1"

func getEncoder() (*zstd.Encoder, bool) {
    if poolDisabled {
        return nil, false // fall through to fresh creation
    }
    // ... existing logic
}
```

Same for `getDecoder`. Zero behavior change when unset. Five lines of cheap insurance.

## Why the suggestion

### Reset semantics check out

A direct read of klauspost's Reset code at the vendored `v1.18.0` — the foundation the whole pool depends on:

- **[`Encoder.Reset(w)`](https://github.com/klauspost/compress/blob/v1.18.0/zstd/encoder.go#L96-L132)**: drains pending goroutines via `s.wg.Wait()` and `s.wWg.Wait()`, resets the [inner block encoder](https://github.com/klauspost/compress/blob/v1.18.0/zstd/encoder.go#L122) (keeps dict, clears block state), zeros all frame flags (`headerWritten`, `eofWritten`, `fullFrameWritten`), sets `s.w = w`, zeros all counters and errors. Options (`e.o`) persist — so concurrency, compression level, dict stay stable across pool reuse.
- **`Encoder.Reset(nil)`**: explicitly supported (same function, with `w=nil`). Documented API for "release the writer reference before pooling." Next `Reset(realWriter)` reinitializes cleanly.
- **[`Decoder.Reset(r)`](https://github.com/klauspost/compress/blob/v1.18.0/zstd/decoder.go#L171-L238)**: returns `ErrDecoderClosed` if Close was called (closed decoders cannot be reused — handled by `putDecoder` skipping the pool on Reset error). If `r == nil`, sets `current.err = ErrDecoderNilInput` — supported idle state. Otherwise stashes and reinitializes with explicit `streamWg.Wait()` drain.

The pattern this PR uses — `Close()` → `putEncoder` (which calls `Reset(nil)`) → next caller → `Reset(realWriter)` — is exactly what these APIs are designed for. Not a clever workaround; it's the documented reuse pattern.

### Residual risk that motivates the kill-switch

What the PR can't fully de-risk, and why being deliberate is worth it:

- [`s.encoder.Reset(dict, false)` at `encoder.go:122`](https://github.com/klauspost/compress/blob/v1.18.0/zstd/encoder.go#L122) resets the **inner** block encoder by calling its Reset, not recreating. If klauspost ever has a quiet bug in one of the block-encoder implementations where inner state doesn't fully clear (adaptive compression state, dict cache, amd64-assembly sequence-encoder state), pooled calls would silently produce different bytes than fresh-allocation calls. Extremely unlikely — API contract says Reset fully cleans — but it's the kind of thing only real prod traffic proves.
- Future klauspost upgrades could silently change Reset semantics. Tests here would still pass; uplift would fail.
- Since c1z writes are consumed by C1's uplift pipeline (and archived immutably to S3), a silent byte-level corruption would surface as "failed to decompress c1z" in uplift rather than at sync time — a bad failure mode to chase across logs after the fact.

Blast radius: every sync on every tenant.

pkg/dotc1z/pool.go
```diff
@@ -0,0 +1,91 @@
...
+// Pooled encoders are single-threaded (concurrency=1) to match the production
+// default set in c1file.go. A concurrent encoder path (GOMAXPROCS) is intentionally
+// avoided here because klauspost/compress's multi-threaded encoder has theoretical
+// race concerns in its goroutine synchronization.
+var encoderPool = &sync.Pool{}
+
+// pooledEncoderConcurrency is the concurrency level used for pooled encoders.
+// Must match the c1zOptions / C1File default (1) so the pool actually engages
+// on the sync hot path.
+const pooledEncoderConcurrency = 1
```
btipling, line 28: Minor clarification — this comment says the pool "actually engages on the sync hot path" but `pkg/synccompactor/compactor.go:136` passes `WithEncoderConcurrency(0)` and intentionally bypasses the pool. Noting it inline so it doesn't surprise anyone later:

```suggestion
// pooledEncoderConcurrency is the concurrency level used for pooled encoders.
// Must match the c1zOptions / C1File default (1) so the pool actually engages
// on the sync hot path. Note: pkg/synccompactor/compactor.go passes
// WithEncoderConcurrency(0) (GOMAXPROCS) and intentionally bypasses the pool.
const pooledEncoderConcurrency = 1
```

pkg/dotc1z/pool_test.go
```diff
@@ -0,0 +1,544 @@
...
+func BenchmarkEncoderPoolAllocs(b *testing.B) {
+	testData := bytes.Repeat([]byte("benchmark data "), 1000)
+	tmpDir := b.TempDir()
+
+	dbFile := filepath.Join(tmpDir, "bench.db")
+	err := os.WriteFile(dbFile, testData, 0600)
+	require.NoError(b, err)
+
+	b.Run("pooled_encoder", func(b *testing.B) {
+		b.ReportAllocs()
```
btipling, line 399: Missing `b.ResetTimer()` after `b.ReportAllocs()` — sibling benchmarks `BenchmarkEncoderAllocationOnly` (line 434) and its non-pooled sub-bench (line 452) do this. Benign today because setup is above the `b.Run`, but matches convention:

```suggestion
		b.ReportAllocs()
		b.ResetTimer()
```

Same one-line addition applies to the companion benchmark blocks at lines 366, 482, and 509.

pkg/dotc1z/pool.go
arreyder: Addressed in ebfd2377 — extended the comment to call out the synccompactor bypass exactly as suggested.

pkg/dotc1z/pool_test.go
arreyder: Addressed in ebfd2377 — added b.ResetTimer() after b.ReportAllocs() in all four benchmark sub-benches (lines 355, 366, 482, 509). Matches the convention used in BenchmarkEncoderAllocationOnly.

arreyder:
Added the env kill-switch in 854cb39d as you suggested. `BATON_ZSTD_POOL_DISABLE=1` makes `getEncoder`/`getDecoder` return `(nil, false)` (callers already handle that by falling through to fresh `zstd.NewWriter`/`NewReader`) and turns `putEncoder`/`putDecoder` into no-ops. Verified with a new `TestPoolDisabled` that also flips the switch off afterwards to confirm the pool stayed empty (so put*-while-disabled are truly no-op, not just delayed).

PR #776, opened 2026-04-21
perf(dotc1z): filter grants by has_external_match column at SQL

arreyder:
## Summary

`processGrantsWithExternalPrincipals` iterates every grant in the sync but only acts on rows carrying an `ExternalResourceMatch{,All,ID}` annotation. On `be-temporal-sync`, `listGrantsWithExpansionInternal` was **20.33% flat / 23.53% cum of alloc_space** (~22 GB / 3 h), dominated by the per-row sqlite `columnBlob` copy + `proto.Unmarshal` work for rows the caller was about to skip anyway.

Push the annotation check down to SQL:

- New `has_external_match` column on `grants`, set on write from the in-memory grant's annotations via a cheap `Any.MessageIs` check (no unmarshal).
- `grantExtractFields` sets the column in **every** upsert mode — including `PreserveExpansion`, because `executeGrantChunkedUpsert` still writes `EXCLUDED.data` on conflict there; the column must track the data blob's annotation set exactly.
- Migration `ALTER`s the column in with default `0` for older c1zs, plus a partial index `(sync_id) WHERE has_external_match = 1` matching the existing `needs_expansion` index shape.
- Backfill (renamed `backfillGrantExpansionColumn` → `backfillGrantDerivedColumns`) derives `has_external_match` while the proto is already unmarshaled for the expansion backfill. Sentinel branch is split by match-presence so each group still bulk-updates in one statement.
- New `GrantListOptions.ExternalMatchOnly` opted into by `syncer.listAllGrantsWithExpansion`. The in-loop annotation check at `processGrantsWithExternalPrincipals` is kept as a defensive fallback.
- `BATON_DISABLE_EXTERNAL_MATCH_FILTER=1` kill-switch for fast revert without re-release.

Clone and attached-diff paths are schema-agnostic (column discovery via `PRAGMA table_info`) so the new column flows through automatically.

## Benchmark

`BenchmarkListGrantsWithExpansion_ExternalMatchOnly` seeds N grants at varying match densities and compares the filtered path against the pre-fix behavior via the kill-switch. `benchtime=3x` on AMD Ryzen 7 5700X3D:

**n=10,000 grants** (pre-fix: ~14.5 MB, ~380k allocs, ~37 ms regardless of density):

| match % | allocs | B/op | alloc reduction | wall time |
|---:|---:|---:|---:|---:|
| 1% | 4,374 | 266 KB | **98.8%** | 0.6 ms (60×) |
| 5% | 21,164 | 970 KB | **94.4%** | 3.4 ms (11×) |
| 20% | 84,127 | 3.6 MB | **77.9%** | 10.3 ms (3.7×) |
| 100% | 419,924 | 17.7 MB | 0% (ceiling) | 42.9 ms (parity) |

**n=50,000 grants** (pre-fix: ~72 MB, ~1.9 M allocs, ~177 ms):

| match % | allocs | B/op | alloc reduction | wall time |
|---:|---:|---:|---:|---:|
| 1% | 21,174 | 969 KB | **98.9%** | 3.3 ms (54×) |
| 5% | 105,163 | 4.5 MB | **94.5%** | 15.9 ms (11×) |
| 20% | 420,127 | 17.7 MB | **78.4%** | 49.5 ms (3.7×) |
| 100% | 2,100,784 | 88.4 MB | 0% (ceiling) | 204.8 ms (parity) |

100% match density is pinned as a regression guard — the filter imposes no measurable cost vs. the pre-fix path when every row would match anyway.

## Pressure test

Before implementing, verified (in order):

1. `grantExtractFields` is called on every upsert with the full `*v2.Grant` — cheap `Any.MessageIs` works.
2. `listGrantsWithExpansionInternal` has only one caller in-repo (the one we convert).
3. `ExternalResourceMatch*` annotations are only set by connectors pre-`PutGrants`; no in-place annotation mutation exists. `newGrantForExternalPrincipal` copies them to expanded grants, which go back through `UpsertGrants(Replace)` — extraction runs fresh.
4. `state.HasExternalResourcesGrants()` is observation-based, not eager.
5. Migrations run synchronously in `NewC1File` → `InitTables` → `Migrations` before any read can happen.
6. `PreserveExpansion` upsert rewrites `data` from `EXCLUDED.data`, so `has_external_match` must track it — covered by the update map adjustment and a targeted test.

## Test plan

- [x] `TestHasExternalMatch_WriteExtraction` — every annotation flavor + co-existence with `GrantExpandable`
- [x] `TestHasExternalMatch_PreserveExpansionUpdatesColumn` — the critical PreserveExpansion refresh case
- [x] `TestListGrantsInternal_ExternalMatchOnlyFilter` — SQL predicate returns exactly the matching rows
- [x] `TestBackfillMigration_PopulatesHasExternalMatch` — all three backfill branches (expandable+match, sentinel+match, sentinel+no-match)
- [x] `TestExternalMatchFilter_KillSwitch` — env var disables the predicate
- [x] Existing `TestExternalResourceMatchIDWithExpandableRemapping` + full sync/dotc1z/synccompactor suites pass unchanged
- [x] `TestCloneSyncMigratedColumnOrder` updated to simulate old c1z with the new ALTER
- [x] `go vet ./...` clean
- [ ] Rollout: monitor alloc on `be-temporal-sync` after deploy; if the predicate mis-plans on a customer's c1z, `BATON_DISABLE_EXTERNAL_MATCH_FILTER=1` rolls back without a re-release

### DIFF
--- a/pkg/connectorstore/connectorstore.go
+++ b/pkg/connectorstore/connectorstore.go
@@ -143,6 +143,10 @@ type GrantListOptions struct {
 	// NeedsExpansionOnly filters rows to needs_expansion=1.
 	// Used by expansion-only modes.
 	NeedsExpansionOnly bool
+	// ExternalMatchOnly filters rows to grants carrying an ExternalResourceMatch
+	// annotation. Used by payload+expansion mode by processGrantsWithExternalPrincipals
+	// to skip the unmarshal work for rows the caller would otherwise discard.
+	ExternalMatchOnly bool
 
 	// PageToken and PageSize are used for pagination in all modes.
 	// SyncID is used for expansion-only modes.
--- a/pkg/dotc1z/c1file_store.go
+++ b/pkg/dotc1z/c1file_store.go
@@ -257,6 +257,41 @@ func (g c1FileGrantStore) ListWithAnnotations(ctx context.Context) iter.Seq2[Gra
 	}
 }
 
+// ListWithAnnotationsExternalMatchOnly implements GrantStore. Same shape as
+// ListWithAnnotations but filters at SQL via the has_external_match column,
+// skipping the per-row data-blob unmarshal for grants that processGrants-
+// WithExternalPrincipals would discard. Set BATON_DISABLE_EXTERNAL_MATCH_FILTER=1
+// to fall back to the unfiltered scan.
+func (g c1FileGrantStore) ListWithAnnotationsExternalMatchOnly(ctx context.Context) iter.Seq2[GrantAnnotation, error] {
+	return func(yield func(GrantAnnotation, error) bool) {
+		pageToken := ""
+		for {
+			if err := ctx.Err(); err != nil {
+				_ = yield(GrantAnnotation{}, err)
+				return
+			}
+			resp, err := g.c.listGrantsWithExpansionInternal(ctx, grantListOptions{
+				Mode:              grantListModePayloadWithExpansion,
+				ExternalMatchOnly: true,
+				PageToken:         pageToken,
+			})
+			if err != nil {
+				_ = yield(GrantAnnotation{}, err)
+				return
+			}
+			for _, ga := range grantAnnotationRowsFromInternal(resp.Rows) {
+				if !yield(ga, nil) {
+					return
+				}
+			}
+			if resp.NextPageToken == "" {
+				return
+			}
+			pageToken = resp.NextPageToken
+		}
+	}
+}
+
 // -----------------------------------------------------------------------------
 // SyncMeta
 // -----------------------------------------------------------------------------
--- a/pkg/dotc1z/clone_sync_test.go
+++ b/pkg/dotc1z/clone_sync_test.go
@@ -74,6 +74,9 @@ func TestCloneSyncMigratedColumnOrder(t *testing.T) {
 	_, err = srcFile.db.ExecContext(ctx,
 		fmt.Sprintf("ALTER TABLE %s ADD COLUMN needs_expansion integer not null default 0", grants.Name()))
 	require.NoError(t, err)
+	_, err = srcFile.db.ExecContext(ctx,
+		fmt.Sprintf("ALTER TABLE %s ADD COLUMN has_external_match integer not null default 0", grants.Name()))
+	require.NoError(t, err)
 
 	// Create partial indexes that the migration would add.
 	_, err = srcFile.db.ExecContext(ctx, fmt.Sprintf(
@@ -82,6 +85,10 @@ func TestCloneSyncMigratedColumnOrder(t *testing.T) {
 	require.NoError(t, err)
 	_, err = srcFile.db.ExecContext(ctx, fmt.Sprintf(
 		"CREATE INDEX IF NOT EXISTS idx_grants_sync_needs_expansion_v1 ON %s (sync_id) WHERE needs_expansion = 1",
+		grants.Name()))
+	require.NoError(t, err)
+	_, err = srcFile.db.ExecContext(ctx, fmt.Sprintf(
+		"CREATE INDEX IF NOT EXISTS idx_grants_sync_external_match_v1 ON %s (sync_id) WHERE has_external_match = 1",
 		grants.Name()))
 	require.NoError(t, err)
 
--- a/pkg/dotc1z/engine/pebble/adapter_grants_store.go
+++ b/pkg/dotc1z/engine/pebble/adapter_grants_store.go
@@ -256,3 +256,12 @@ func (g pebbleGrantStore) ListWithAnnotations(ctx context.Context) iter.Seq2[dot
 		}
 	}
 }
+
+// ListWithAnnotationsExternalMatchOnly implements GrantStore. The Pebble
+// engine does not maintain the has_external_match column the SQLite engine
+// uses for SQL-side filtering; fall back to the unfiltered iterator. The
+// caller (processGrantsWithExternalPrincipals) re-checks the annotation
+// in-Go, so correctness is preserved.
+func (g pebbleGrantStore) ListWithAnnotationsExternalMatchOnly(ctx context.Context) iter.Seq2[dotc1z.GrantAnnotation, error] {
+	return g.ListWithAnnotations(ctx)
+}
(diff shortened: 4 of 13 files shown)


arreyder:
## Reviving this PR with a fresh adaptation against current main

The branch was parked since 2026-05-22 on a 4-file rebase conflict against main. Main moved `GrantListOptions`/`GrantListMode`/`InternalGrantRow` from the public `connectorstore` package to an internal `pkg/dotc1z/internal_grant_options.go`, and replaced the raw `ListGrantsInternal` entry point with the `GrantStore.ListWithAnnotations*` family.

Re-authored as a single clean commit (`b707829b`) preserving the original schema, write-path, and SQL predicate. The `ExternalMatchOnly` field lives on the internal `grantListOptions` and is opted into via a new `GrantStore.ListWithAnnotationsExternalMatchOnly` method. Pebble adapter falls back to the unfiltered iterator (no SQL column there); the in-loop annotation check in `processGrantsWithExternalPrincipals` stays as a correctness fallback.

## Why this still matters

I benchmarked **current main** against the proposed PR with a fresh `BenchmarkListGrantsExternalMatch_PRRevival`, on AMD Ryzen 7 5700X3D, `benchtime=5x`:

**n=10,000 grants**

| match % | unfiltered (main) | filtered (this PR) | speedup |
|---:|---:|---:|---:|
| 1% | 35.8 ms / 15.6 MB / 390k allocs | **0.60 ms / 173 KB / 4,266 allocs** | **59× / 90× less mem / 92× fewer allocs** |
| 5% | 35.7 ms / 15.7 MB / 392k allocs | **3.29 ms / 819 KB / 20,260 allocs** | **11× / 19× / 19×** |
| 20% | 36.9 ms / 16.1 MB / 398k allocs | **10.18 ms / 3.2 MB / 80,224 allocs** | **3.6× / 5× / 5×** |
| 100% | 38.7 ms / 18.1 MB / 430k allocs | 40.5 ms / 16.2 MB / 400k allocs | parity (regression guard) |

**n=50,000 grants**

| match % | unfiltered (main) | filtered (this PR) | speedup |
|---:|---:|---:|---:|
| 1% | 177.1 ms / 78.2 MB / 1.95M allocs | **3.14 ms / 818 KB / 20,267 allocs** | **56× / 96× / 96×** |
| 5% | 178.7 ms / 78.7 MB / 1.96M allocs | **14.76 ms / 4.04 MB / 100,259 allocs** | **12× / 19× / 20×** |
| 20% | 179.6 ms / 80.5 MB / 1.99M allocs | **48.96 ms / 16.2 MB / 400,229 allocs** | **3.7× / 5× / 5×** |
| 100% | 196.9 ms / 90.4 MB / 2.15M allocs | 199.6 ms / 80.8 MB / 2.00M allocs | parity |

These numbers match the original PR's claim almost exactly (the PR claimed 60× at 1%/10k; actual is 59×). The slim-grants writer (#781) is in main's baseline and didn't close the gap — slim still requires an unmarshal, and the consumer filters in Go after that. Pushing the filter to SQL via the partial index sidesteps both costs.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./pkg/dotc1z/...` green (32s)
- [x] `TestCloneSyncMigratedColumnOrder` updated for the new `ALTER TABLE … ADD COLUMN has_external_match` migration step
- [x] Benchmark above shows the win is intact

## Out of scope vs. original PR

The original branch carried ~313 lines of test cases against the deprecated public `GrantListOptions` API. Porting those to main's new internal surface is a separate change; happy to follow up if reviewers want them. The benchmark above + the existing test suite (which exercises the new column on every grant write) cover the regression risk.

arreyder:
## Re-bench after critical-fix commit `26b535cd`

Sanity-checked the bench on the post-fix HEAD to confirm the perf claim survived adding the `has_external_match_backfilled` flag + dedicated backfill loop. The fix only touches the write/backfill path; the read path the bench measures is unchanged.

`benchtime=5x`, same AMD Ryzen 7 5700X3D, post-fix HEAD (`26b535cd`):

**n=10,000 grants**

| match % | unfiltered (main baseline) | filtered (this PR) | speedup |
|---:|---:|---:|---:|
| 1% | 35.03 ms / 15.63 MB / 390k allocs | **0.60 ms / 172 KB / 4,264 allocs** | **58× / 91× less mem / 92× fewer allocs** |
| 5% | 36.49 ms / 15.73 MB / 392k allocs | **3.19 ms / 819 KB / 20,258 allocs** | **11× / 19× / 19×** |
| 20% | 37.70 ms / 16.10 MB / 398k allocs | **9.80 ms / 3.23 MB / 80,225 allocs** | **3.8× / 5× / 5×** |
| 100% | 40.24 ms / 18.08 MB / 430k allocs | 41.12 ms / 16.16 MB / 400k allocs | parity (regression guard) |

**n=50,000 grants**

| match % | unfiltered (main baseline) | filtered (this PR) | speedup |
|---:|---:|---:|---:|
| 1% | 179.39 ms / 78.16 MB / 1.95M allocs | **3.14 ms / 817 KB / 20,266 allocs** | **57× / 96× / 96×** |
| 5% | 178.12 ms / 78.66 MB / 1.96M allocs | **15.79 ms / 4.05 MB / 100,266 allocs** | **11× / 19× / 20×** |
| 20% | 180.09 ms / 80.51 MB / 1.99M allocs | **47.64 ms / 16.17 MB / 400,229 allocs** | **3.8× / 5× / 5×** |
| 100% | 201.51 ms / 90.43 MB / 2.15M allocs | 201.85 ms / 80.83 MB / 2.00M allocs | parity |

The bench now also asserts row-count correctness: `b.Fatalf` if either path yields a count other than `numGrants * matchPercent / 100`. A regression that under-selects rows would abort the bench rather than silently look like a speedup.

### What changed since the prior comment

`26b535cd` addressed a critical correctness bug surfaced in a multi-agent code review of `b707829b`: `backfillGrantDerivedColumns` is gated on `sync_runs.grants_backfilled=0`, but every existing sync was already flipped to 1 by the prior expansion-column PR. Migration would have ALTERed in `has_external_match` with default 0, the gated backfill would no-op, the SQL predicate would filter to zero rows, and `processGrantsWithExternalPrincipals` would silently process nothing on every existing c1z. Fixed by introducing a separate `has_external_match_backfilled` flag column on `sync_runs` and a dedicated backfill loop. New regression test `TestBackfillHasExternalMatchColumn_OldSyncWithBackfilledFlag` reproduces the failure mode and asserts the fix.

PR #777, opened 2026-04-21
feat(dotc1z): record decompressed db size on save (zstd FCS + Info log)

arreyder:
## Summary

Two changes, one feature:

1. **Producer side** — `saveC1z` now writes the decompressed db size into the zstd frame header (Frame_Content_Size / FCS) and logs it at Info alongside the resulting compressed size. Upstream-spec-compatible — every existing zstd decoder ignores FCS and continues to stream exactly as today.
2. **Consumer side** — the dotc1z `decoder` peeks the frame header during init and exposes `DeclaredDecodedSize() (uint64, bool)`. When the declared size already exceeds `DecoderMaxDecodedSize`, the first `Read` returns `ErrMaxSizeExceeded` with the exact declared size embedded in the message — no bytes decompressed, no decoder warmed up.

## Why

The hit today isn't at sync time — it's during **grant expansion**. Expansion materializes one grant per (member, inherited access) tuple from the annotations the connector emits, and on some tenants that's enough to push the c1z past the 3 GiB decoder cap (observed: >200 GB decompressed). When downstream reopens the post-expansion c1z, it trips `ErrMaxSizeExceeded` with no signal about how much more data is in the stream. The only way to find the true size today is to download the c1z and re-decompress with a raised cap — operationally expensive, often not possible during an incident.

`saveC1z` runs once per `C1File.Close()`, so the final post-expansion save records the final post-expansion size. With FCS now in the header, any reader can learn that size directly — either via the 14-byte frame header (no dependency on this SDK) or via `DeclaredDecodedSize()` (this SDK). Fail-fast on cap means the next bad c1z reports its declared size in the error instead of a generic "cap exceeded".

### Producer (`pkg/dotc1z/file.go`)
- Stat the db file before encoding so we know its exact size.
- Initialize the zstd encoder, then call `ResetContentSize(outFile, dbSize)` so the frame header carries FCS. `ResetContentSize` requires the written byte count to equal the advertised size at `Close()`, so the `io.Copy` is replaced with `io.CopyN(..., dbSize)` to make that invariant explicit.
- After successful rename, log `c1z: saved` at Info with `decompressed_bytes` + `compressed_bytes` for operational visibility.

### Consumer (`pkg/dotc1z/decoder.go`)
- `decoder` grows `declaredSize` / `hasDeclaredSize` fields populated lazily during init.
- Init logic extracted into `ensureInit()`, called from both `Read` and the new `DeclaredDecodedSize()`.
- After `ReadHeader` strips the 5-byte C1ZF magic, wrap the post-magic reader in `bufio.Reader` so we can `Peek(zstd.HeaderMaxSize)` the zstd frame header without consuming those bytes — the zstd decoder reads them back from bufio's buffer when it starts.
- If FCS is advertised and `declaredSize > maxDecodedSize`, init stores an `ErrMaxSizeExceeded`-wrapped error with `"declared decoded size N exceeds cap M"`. The pooled / fresh zstd decoder is never even constructed in this case.
- Legacy c1zs (saved before this PR) don't advertise FCS — `DeclaredDecodedSize` returns `(0, false)`, and `Read` behaves exactly as before.

## Tests

Producer:
- `TestSaveC1z_SetsFrameContentSize` — decodes the frame header of a freshly-saved c1z, asserts `header.HasFCS` and `header.FrameContentSize == dbSize`.
- All existing `TestSaveC1z` / `TestLoadC1z` subtests pass (including the empty-db edge case).

Consumer:
- `TestDecoder_DeclaredDecodedSize` — `DeclaredDecodedSize` returns exact FCS before any `Read`; `Read` then streams the full payload; second call is idempotent.
- `TestDecoder_DeclaredDecodedSize_LegacyWithoutFCS` — hand-built legacy c1z (C1ZF magic + zstd frame without `ResetContentSize`) returns `(0, false)` and still decodes cleanly.
- `TestDecoder_DeclaredSizeExceedsCap_FailsFast` — cap set below declared size; first `Read` returns `ErrMaxSizeExceeded` with `"declared decoded size"` in the message; zero bytes produced.
- `TestC1ZDecoder_WindowSizeExceeded` (16 MiB random payload) — moved here from `TestC1ZDecoder`, since FCS-aware SingleSegment frames make small-payload window-exceeded checks impossible. Preserves coverage of `ErrWindowSizeExceeded`.

## Compat

- No c1z format change — the 5-byte `C1ZF\x00` magic and zstd stream layout are unchanged; FCS is an optional field inside the zstd frame header.
- Older decoders (baton-sdk ≤ v0.8.25) ignore FCS — no behavioral change for any existing reader.
- Callers that rely on `errors.Is(err, ErrMaxSizeExceeded)` keep working. The message now carries the declared size when known.

## Follow-ups (separate PRs, not in scope)

- Consumer-side (c1 repo): surface `DeclaredDecodedSize()` in the `ErrMaxSizeExceeded` log line at the fetch site so per-tenant c1z-growth is queryable; add an OTel metric (e.g. `c1z.size.decompressed_bytes` histogram) keyed on tenant.
- Optional: expose an `Inspect` helper on `dotc1z` that returns FCS without constructing a decoder, so ops tooling can parse the 14-byte header without any zstd setup.

### DIFF
--- a/pkg/dotc1z/decoder.go
+++ b/pkg/dotc1z/decoder.go
@@ -275,24 +275,38 @@ func (d *decoder) Read(p []byte) (int, error) {
 		return 0, d.o.ctx.Err()
 	}
 
-	// Check we have not exceeded our max decoded size. When the frame declared
-	// its size and was within the cap, this overrun path is effectively
-	// unreachable; it remains as defense against a mismatched FCS.
+	// Enforce max decoded size at both ends of the underlying Read:
+	// - Top-of-call: short-circuit subsequent Reads after we've already tripped.
+	// - Post-call: clip `n` so bytes beyond the cap never reach the caller,
+	//   even when a single zstd read straddles the cap boundary.
+	// When FCS is present and within the cap the post-call branch is
+	// unreachable; it's defense against a missing or mismatched FCS.
 	maxDecodedSize := d.getMaxDecodedSize()
 	if d.decodedBytes > maxDecodedSize {
-		if d.hasDeclaredSize {
-			return 0, fmt.Errorf(
-				"c1z: max decoded size exceeded: %d > %d (declared: %d): %w",
-				d.decodedBytes, maxDecodedSize, d.declaredSize, ErrMaxSizeExceeded,
-			)
-		}
-		return 0, fmt.Errorf("c1z: max decoded size exceeded: %d > %d: %w", d.decodedBytes, maxDecodedSize, ErrMaxSizeExceeded)
+		return 0, d.maxSizeExceededErr()
 	}
 
 	// Do underlying read
 	n, err := d.zd.Read(p)
 	//nolint:gosec // No risk of overflow/underflow because n is always >= 0.
 	d.decodedBytes += uint64(n)
+
+	// Clip any bytes that crossed the cap in this single Read so we never
+	// return data past maxDecodedSize, and surface the error immediately
+	// instead of waiting for the next Read.
+	if d.decodedBytes > maxDecodedSize {
+		overrun := d.decodedBytes - maxDecodedSize
+		//nolint:gosec // overrun <= n by construction (we just added n and went over).
+		if uint64(n) >= overrun {
+			n -= int(overrun)
+		} else {
+			n = 0
+		}
+		// Leave d.decodedBytes at its true post-read value so subsequent
+		// Reads hit the top-of-call guard above.
+		return n, d.maxSizeExceededErr()
+	}
+
 	if err != nil {
 		// NOTE(morgabra) This happens if you set a small DecoderMaxMemory
 		if errors.Is(err, zstd.ErrWindowSizeExceeded) {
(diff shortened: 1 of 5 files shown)


pkg/dotc1z/decoder.go
```diff
@@ -199,12 +275,17 @@ func (d *decoder) Read(p []byte) (int, error) {
...
 
-	// Check we have not exceeded our max decoded size
-	maxDecodedSize := d.o.maxDecodedSize
-	if maxDecodedSize == 0 {
-		maxDecodedSize = defaultMaxDecodedSize
-	}
+	// Check we have not exceeded our max decoded size. When the frame declared
+	// its size and was within the cap, this overrun path is effectively
+	// unreachable; it remains as defense against a mismatched FCS.
+	maxDecodedSize := d.getMaxDecodedSize()
```
kans, line 284: What if the subsequent read sends us over the limit?

arreyder: Good catch — fixed in 3b5df15d. `Read` now enforces the cap in two places: the existing top-of-call guard for subsequent calls plus a post-decode clip inside the same `Read` that trims the returned `n` so bytes past `maxDecodedSize` never reach the caller, even when a single zstd read straddles the boundary. `TestDecoder_ReadClipsBytesAtCap` pins the contract.

btipling (approved):

PR #779, opened 2026-04-22
feat(expand): periodic mid-expansion progress log

arreyder:
## Summary

Emits a log with live uncompressed db size and growth delta every time the syncer calls `ProgressLog.LogExpandProgress` (once per expansion step, rate-limited to the existing `WithLogFrequency` default of 10s), so operators can watch long grant-expansion runs grow in near-real-time instead of learning the final size only after `C1File.Close()` → `saveC1z`. Complements #777 (post-save size visibility) with during-run visibility.

## Why

A pathological tenant's grant expansion runs for ~10 hours and can grow non-linearly (transitive expansion: each newly-materialised grant can trigger more). With #777 alone we learn the final post-expansion size once saveC1z emits its `c1z: saved` line at the end — one data point after a 10-hour burn.

This PR gives the signal you actually want for that shape of workload: **delta bytes between successive samples**. Two logs at 45 GB → 120 GB tell you to abort immediately rather than wait 6 more hours for the final save.

### New capability interface: `pkg/connectorstore/connectorstore.go`
- `DBSizeProvider` — optional capability: returns current uncompressed working-set bytes. Additive; stores that don't implement it keep working. Placed alongside `LatestFinishedSyncIDFetcher` (matches existing pattern for optional store capabilities).

### `pkg/dotc1z/c1file.go`
- `(*C1File).CurrentDBSizeBytes() (int64, error)` — stats `dbFilePath` **plus the `-wal` sidecar** if present. WAL matters here: with `journal_mode=WAL` the main db file can stay flat while the WAL grows into hundreds of MB during active writes. Non-`ENOENT` stat errors on the `-wal` sidecar are surfaced (not silently swallowed) so a permission/EIO error can't under-report by hundreds of MB.
- `var _ connectorstore.DBSizeProvider = (*C1File)(nil)` — compile-time assertion so signature drift fails the build rather than silently disabling the size fields.

### `pkg/sync/progresslog/progresslog.go`
- `ProgressLog` gains optional `dbSize` + `lastLoggedDBSize` fields.
- New `WithDBSizeProvider(connectorstore.DBSizeProvider) Option`.
- `LogExpandProgress` — when a provider is attached, appends `decompressed_bytes` + `delta_bytes` to the existing `"Expanding grants"` log. Delta is computed across calls, not from zero (so growth rate is visible per sample).
- Rate-limiting reuses the existing `lastActionLog` field — no new env var. Tune via `WithLogFrequency` (10s default).

### `pkg/sync/syncer.go`
- In `NewSyncer`, type-asserts `s.store` against `connectorstore.DBSizeProvider` and calls `progresslog.WithDBSizeProvider(sp)` when the store implements it. `*dotc1z.C1File` does.

### Why state lives on `ProgressLog`, not `Expander`
`syncer.expandGrantsForEntitlements` constructs a fresh `Expander` on every step and runs `RunSingleStep` once before returning. Any state held on the `Expander` resets every step — including any "last log time" tracker, which would always be ~nanoseconds ago and suppress every log. `ProgressLog` is constructed once per sync and persists across steps, so it's the correct home for cross-step rate-limiting and size-delta accounting.

## Tests

In `pkg/sync/progresslog/progresslog_test.go`:

- `TestLogExpandProgress_SizeFieldsAcrossCalls` — first call emits with `delta_bytes == size` (baseline 0); calls inside the rate-limit window are suppressed; call after the window reports delta relative to the prior sample (catches the regression where per-step state reset would make delta == full size every time).
- `TestLogExpandProgress_NoSizeProviderOmitsFields` — pre-existing behavior preserved when no provider is attached; only `actions_remaining` appears.
- `TestLogExpandProgress_SizeErrorSkipsSizeFields` — stat error on the size provider suppresses only the size fields, not the whole log line.
- `TestLogExpandProgress_PerStepExpanderShape` — simulates the production call pattern: 50 back-to-back Expander constructions in a ~75ms window with a 15ms rate-limit. Asserts ≥2 logs emit, logs are rate-limited (not one-per-step), and the second log's `delta_bytes` < `decompressed_bytes` — proving the `ProgressLog` kept its prior sample across steps.

In `pkg/dotc1z/c1file_test.go`:

- `TestCurrentDBSizeBytes` — live sqlite file size grows after bulk writes; defensive error on empty `dbFilePath`.

The test file defines a tiny local `capturingCore` (~30 LOC) instead of pulling in `zaptest/observer` since that dependency isn't in the vendor tree.

## Compat

- Zero API break. `NewExpander` signature unchanged.
- `DBSizeProvider` is an optional interface — nothing requires implementing it.
- The log message stays `"Expanding grants"` with the existing `actions_remaining` field; the new size fields are additive.
- Rate-limiting uses the same `WithLogFrequency` knob (10s default) as every other progress log — no new env var.

## Next steps (follow-ups, not in scope)

- **Bounded expansion (baton-sdk)** — reuse the same `CurrentDBSizeBytes()` hook to *abort* instead of only log when size crosses a configurable ceiling. Shape: new `BATON_MAX_EXPANSION_DB_BYTES` env var (default: off), new `ErrExpansionSizeExceeded` sentinel wrapping the observed size and ceiling, abort check threaded through `Expander.RunSingleStep`. Abort-on-overrun semantics (not partial expansion). ~50 LOC + tests. Tune the default ceiling against the histogram this PR feeds, not by guessing. Tracked as **OPS-1258**.
- **Per-tenant `DecoderMaxDecodedSize` (c1)** — today the decoder cap is gated by the process-wide `BATON_DECODER_MAX_DECODED_SIZE_MB` env var, which is too coarse for multi-tenant `be-temporal-sync` / `be-temporal-worker` topology (raising it for one pathological tenant raises it for everyone). Add a tenant-keyed config value (feature flag or tenant-settings field) and pass `dotc1z.WithDecoderMaxDecodedSize(tenantLimit)` at each `NewDecoder` call site in c1, falling back to the existing 3 GiB default. ~20–30 LOC + config surface. Lets ops raise the cap for a specific tenant once this PR's growth log identifies who needs it, without lifting the floor globally. Tracked as **CXH-1398**.


arreyder:
Agent review addressed in cc4000ea. All four points confirmed and fixed:

**#1 [Critical] Progress log never fires — Expander recreated per step.** Confirmed by reading `syncer.go:2554`: each `expandGrantsForEntitlements` call constructs a fresh `Expander` and runs `RunSingleStep` once, so all my per-expander state (`expansionStart`, `lastProgressLog`, `actionsProcessed`, `lastLoggedSize`) reset every step. With the 30s default interval, elapsed is always ~ns and the gate short-circuits — the feature was dead in production.

Restructured per the reviewer's suggestion (a): moved everything onto `progresslog.ProgressLog`, which is constructed once per sync in `syncer.NewSyncer` and already handles cross-step rate limiting via `lastActionLog`. `LogExpandProgress` now takes an optional `DBSizeProvider` via new `WithDBSizeProvider` option and emits `decompressed_bytes` + `delta_bytes` alongside the existing `actions_remaining` field. The `syncer` type-asserts `s.store` to `DBSizeProvider` at counts construction — `*dotc1z.C1File` satisfies it automatically.

Also dropped the `BATON_EXPANSION_PROGRESS_INTERVAL` env var as redundant with the existing `WithLogFrequency` option (defaults to 10s).

**#2 [High] Missing per-step integration test.** Added `TestLogExpandProgress_PerStepExpanderShape` in `pkg/sync/progresslog/progresslog_test.go`: simulates 50 back-to-back Expander constructions in a ~75ms window with a 15ms rate-limit and asserts (a) ≥2 logs emit, (b) logs are rate-limited (not one-per-step), (c) the second log's `delta_bytes` is strictly less than its `decompressed_bytes`, proving state persisted across the simulated per-step calls. This would have caught #1 the moment it landed.

**#3 [Medium] Missing compile-time DBSizeProvider assertion.** Added `var _ connectorstore.DBSizeProvider = (*C1File)(nil)` in `pkg/dotc1z/c1file.go`. Also took the opportunity to move the interface from `pkg/sync/expand` → `pkg/connectorstore`: natural home (store capability, matches `LatestFinishedSyncIDFetcher` pattern, avoids a new `dotc1z → expand` import edge, and both `progresslog` and `dotc1z` already import `connectorstore`).

**#4 [Medium] WAL stat errors silently swallowed.** Replaced the `if err == nil { ... }` pattern with an explicit switch that only tolerates `errors.Is(err, os.ErrNotExist)`. Any other error (EACCES, EIO, ESTALE) now propagates out of `CurrentDBSizeBytes` instead of silently under-reporting the working-set size.

Also expands to the `Defer` items along the way:

- **Deferred #1 (actionsProcessed never asserted)** → N/A; that counter was removed as part of the restructure (it lived on the per-step Expander and wasn't useful there anyway).
- **Deferred #2 (env var not exercised e2e)** → N/A; env var was removed as redundant.

arreyder:
Multi-agent review fixes landed in 520331aa. Eleven items triaged; nine applied, four deferred (see commit message for rationale).

## Critical (merge-blocking, now fixed)

- **DBSizeProvider was dead in production.** The type assertion in `NewSyncer` ran against a nil `s.store` because both production callers (`pkg/tasks/c1api/full_sync.go`, `pkg/tasks/local/syncer.go`) use `WithC1ZPath` — the store is populated later inside `loadStore`. Moved to a new `SetDBSizeProvider` method called from `wireCountsDBSizeProvider()`, invoked idempotently from both `NewSyncer` (handles the `WithConnectorStore` eager-store path) and `loadStore` (handles the `WithC1ZPath` deferred-store path). My own integration test didn't cover this wiring — added `TestLogExpandProgress_NilProviderDoesNotPanic` to at least cover the typed-nil edge case the reviewer flagged.

## High (now fixed)

- **`os.Stat` held under the write mutex.** `LogExpandProgress` held `p.mu.Lock()` across `CurrentDBSizeBytes()` — which does two local stats. On a slow fs (exactly when we want this signal), that stalled every concurrent `AddResources` / `AddGrantsProgress`. Restructured to an RLock fast-path for the rate-limit check, a brief Lock to claim the log slot with double-checked locking, provider I/O outside any lock, and a separate brief Lock to publish the new size.
- **`TestLogExpandProgress_SizeFieldsAcrossCalls` flake margin.** 25ms/40ms (15ms margin) was too tight for coarse-clock Windows — same class as the flake commit `cf3a6c6a` fixed elsewhere. Widened to 10ms/100ms (10x margin).
- **WAL-sum path wasn't actually tested.** `TestCurrentDBSizeBytes` used WAL mode but never asserted the reported size included the sidecar. Added an assertion that compares against `stat(main) + stat(-wal)`.
- **Non-ENOENT WAL stat error path wasn't tested.** New `TestCurrentDBSizeBytes_WALStatErrorPropagates` plants a self-referential symlink at the `-wal` path, triggering ELOOP on `os.Stat`, and asserts the error surfaces with a message identifying WAL as the source.

## Medium (now fixed)

- **First-sample delta equals full size** → gated by a new `hasLoggedDBSize` flag; `decompressed_bytes_delta` is omitted on the first emission so Datadog graphs don't show a spurious baseline-vs-zero spike.
- **Field name `delta_bytes` ambiguous** → renamed to `decompressed_bytes_delta` for symmetry with `decompressed_bytes` and the capability naming.
- **Missing cross-reference to `BATON_DECODER_MAX_DECODED_SIZE_MB`** → added a paragraph to the `WithDBSizeProvider` godoc explaining the 3 GiB default and when a per-tenant override is the right next step.
- **Per-step test clock granularity** → 1500µs → 2ms inter-step sleep; also converted `require.Greater/Less` against `interface{}` to `.(int64)` type assertions for clarity.

## Deferred (documented in the commit, not blocking)

- 10s default rate-limit may be chatty over 10h expansion — will revisit with real histogram data.
- Negative delta after WAL checkpoint is real signal (not a bug).
- Dead `dbFilePath == ""` defensive check — harmless.
- Theoretical DoS via malicious `DBSizeProvider` — only implementer is `C1File`.

Full `go test ./...` + `go vet ./...` clean. 10 consecutive runs of the progresslog tests pass locally.

pkg/sync/progresslog/progresslog.go
```diff
@@ -215,18 +263,77 @@ func (p *ProgressLog) LogGrantsProgress(ctx context.Context, resourceType string
...
+		p.mu.RUnlock()
+		return
+	}
+	p.mu.RUnlock()
+
+	// Claim the log slot under Lock with a double-check against the
+	// rate-limit window, snapshot the provider + prior size, then release
+	// the lock before calling into the provider.
 	p.mu.Lock()
-	defer p.mu.Unlock()
```
kans, line 222: Why drop the defer?  Why not cover the call to `if sz, err := provider.CurrentDBSizeBytes(); err == nil {`, too?

arreyder: Fair call — addressed in 6c250d4d.

**Why I dropped the defer originally:** `p.mu` is shared with the sync hot path (`AddResources`, `AddGrantsProgress`, `AddEntitlementsProgress`, …). `provider.CurrentDBSizeBytes()` does `os.Stat` twice, and the signal this log is meant to emit is "something is going wrong during a long expansion" — i.e. the *exact* scenario where fs latency is worst. Holding `p.mu` across that stat would serialise the sync-hot-path counters on a slow-fs stall. So the split-lock pattern kept I/O outside any mutex.

**Why that answer was wrong anyway:** the resulting claim-slot / release / I/O / re-acquire-to-publish structure was awkward and easy to break on future edits — exactly your "why drop the defer" concern.

**Fix:** dedicated `expandMu sync.Mutex` for all expand-log state (`lastActionLog`, `dbSize`, `lastLoggedDBSize`, `hasLoggedDBSize`). `LogExpandProgress` is now `Lock` + `defer Unlock` over the whole body, the provider call *is* covered, and since `expandMu` is disjoint from `p.mu`, slow-fs stat still can't block hot-path counters. `SetDBSizeProvider` moves to `expandMu` for the same reason. Both concerns addressed, no cost.

`go test -race -count=3 ./pkg/sync/progresslog/...` clean.

btipling (approved):
I'm worried about a separate mutex resulting in a deadlock potential later, but there's no chance of it right now. That new lock is head for the duration of p.dbSize.CurrentDBSizeBytes(); which the bot says is fast. So that's safe I hope

PR #782, opened 2026-04-24
[CE-416] baton-sdk: use reuseTokenSOurce

Bencheng21:
see a lot of token exchange from lambda connector. see [here](https://us3.datadoghq.com/logs?query=%40http.url_details.route%3A%22%2Fauth%2Fv1%2Ftoken%22%20%22Token%3A%20Completed%20token%20exchange%22&agg_m=count&agg_m_source=base&agg_q=%40http.url_details.host&agg_t=count&analyticsOptions=%5B%22bars%22%2C%22dog_classic%22%2Cnull%2Cnull%2C%22value%22%5D&clustering_pattern_field_path=message&cols=service%2C%40http.url_details.host&flat_group_bys=true&fromUser=true&messageDisplay=inline&panel=%7B%22queryString%22%3A%22%40http.url_details.host%3Aplume.conductor.one%22%2C%22filters%22%3A%5B%7B%22isClicked%22%3Atrue%2C%22source%22%3A%22log%22%2C%22path%22%3A%22http.url_details.host%22%2C%22value%22%3A%22plume.conductor.one%22%7D%5D%2C%22queryId%22%3A%22a%22%2C%22timeRange%22%3A%7B%22from%22%3A1775766476000%2C%22to%22%3A1777062476000%2C%22live%22%3Atrue%7D%7D&refresh_mode=sliding&sort_m=count&sort_t=count&storage=hot&stream_sort=%40grpc.method%2Casc&top_n=500&top_o=top&viz=toplist&x_missing=true&from_ts=1770566143674&to_ts=1771862143674&live=true)

looks like we don't use `reuseTokenSource` in DPoPClient. 

## test. 
1. test with `expireIn: nil`. 
image

2. test with normal `reuseTokenSource` 
one `oauth token exchange` request saw on the support dashboard.
image

### DIFF
--- a/pkg/lambda/grpc/config/config.go
+++ b/pkg/lambda/grpc/config/config.go
@@ -96,7 +96,9 @@ func NewDPoPClient(ctx context.Context, clientID string, clientSecret string) (g
 		return nil, nil, nil, fmt.Errorf("new-dpop-client: failed to create token source: %w", err)
 	}
 
-	creds, err := dpop_grpc.NewDPoPCredentials(proofer, tokenSource, configurationHost, []dpop.ProofOption{
+	reuseTokenSource := oauth2.ReuseTokenSource(nil, tokenSource)
+
+	creds, err := dpop_grpc.NewDPoPCredentials(proofer, reuseTokenSource, configurationHost, []dpop.ProofOption{
 		dpop.WithValidityDuration(time.Minute * 5),
 		dpop.WithProofNowFunc(time.Now),
 	})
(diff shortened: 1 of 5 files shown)


pkg/lambda/grpc/config/config.go
```diff
@@ -78,7 +78,9 @@ func NewDPoPClient(ctx context.Context, clientID string, clientSecret string) (g
 		return nil, nil, nil, fmt.Errorf("new-dpop-client: failed to create token source: %w", err)
 	}
 
-	creds, err := dpop_grpc.NewDPoPCredentials(proofer, tokenSource, tokenHost, []dpop.ProofOption{
+	reuseTokenSource := oauth2.ReuseTokenSource(nil, tokenSource)
+
+	creds, err := dpop_grpc.NewDPoPCredentials(proofer, reuseTokenSource, tokenHost, []dpop.ProofOption{
```
kans: We aren't currently setting `token.Expiry` on our tokens when we unmarshal them.  I think `ExpiresIn` (in seconds or something) is on the wire, but we are supposed to set `Expiry` as a *time.Time*.  The go oauth library calls validate on the tokens we return form our source, which pretty much just checks they haven't expired.  There is an early return if the value isn't set (so we would never update the token).  We need to update the dpop library, then pull in that change over here.

kans (approved):

PR #784, opened 2026-04-27
Update SDK version to v0.8.28

### LINKED ISSUE #783: v0.8.27 tag: embedded Version constant still reads v0.8.26
## What

In the `v0.8.27` git tag, [`pkg/sdk/version.go`](https://github.com/ConductorOne/baton-sdk/blob/v0.8.27/pkg/sdk/version.go) still has:

\`\`\`go
const Version = "v0.8.26"
\`\`\`

The previous "Update SDK version to v0.8.X" commits before each release tag (e.g. [`52acb60`](https://github.com/ConductorOne/baton-sdk/commit/52acb60) for v0.8.26, [`772a94b`](https://github.com/ConductorOne/baton-sdk/commit/772a94b) for v0.8.25) appear to have been skipped between v0.8.26 and v0.8.27 — the v0.8.27 tag was placed directly on the `feat(expand): periodic mid-expansion progress log` commit ([`a2bdd0c`](https://github.com/ConductorOne/baton-sdk/commit/a2bdd0c)) without first bumping the constant.

## Impact

Anything that reports the SDK version at runtime — including baton-sdk's own [version subcommand](https://github.com/ConductorOne/baton-sdk/blob/main/pkg/cli/commands.go) and any downstream service log/metric that uses `sdk.Version` for diagnostics — will report v0.8.26 even when v0.8.27 is the actually-loaded module version. Cosmetic but confusing during incident triage.

## Found via

A code-review bot caught this while bumping ConductorOne/c1 from v0.8.23 → v0.8.27 (ductone/c1#17081). The vendored content is correct (matches the tag); only the embedded constant is stale.

## Suggested fix

Either:
1. Cut a v0.8.28 with the constant updated to `v0.8.28`, or
2. Move the v0.8.27 tag to a commit that bumps the constant to `v0.8.27`.

Option 1 is safer (no force-tag).

arreyder:
## Summary

Updates `pkg/sdk/version.go` so the next release can be tagged as v0.8.28 with a matching embedded constant.

Closes #783.

## Why

The v0.8.27 release tag was placed on the `feat(expand): periodic mid-expansion progress log` commit ([`a2bdd0c5`](https://github.com/ConductorOne/baton-sdk/commit/a2bdd0c5)) without the usual "Update SDK version to vX" preceding bump. The embedded constant at the v0.8.27 tag therefore still reads `v0.8.26`, which is cosmetically wrong on every consumer that logs `sdk.Version` at runtime.

## What about the auto-update workflow?

`update-hardcoded-version.yaml` ran on the v0.8.27 release event — workflow history shows three runs (one success, two failures). The successful run apparently no-op'd; no `Update SDK version to v0.8.27` commit appeared on main. Worth a separate look at why, but unblocking is more valuable than diagnosing here.

## Approach

Two options were possible:

1. **Cut v0.8.28 with the constant updated** (chosen here) — single small commit, no force-tag, lets v0.8.27 remain immutable.
2. Move the v0.8.27 tag to a new commit that bumps the constant to v0.8.27 — requires force-tagging an already-published release; risky for downstream consumers that pin to a specific commit SHA.

Picked option 1. Skips `v0.8.27` in the embedded constant (jumps directly to v0.8.28); the v0.8.27 tag stays where it is.

## Test plan

- [x] `make pkg/sdk/version.go VERSION=v0.8.28` produces the expected one-line diff
- [ ] CI green
- [ ] After merge: maintainer cuts v0.8.28 release, workflow optionally regenerates the constant (idempotent)

## Found via

internal platform-repo PR (bumping c1 from baton-sdk v0.8.23 → v0.8.27). The code-review bot on that PR flagged the version mismatch.

### DIFF
--- a/pkg/sdk/version.go
+++ b/pkg/sdk/version.go
@@ -1,3 +1,3 @@
 package sdk
 
-const Version = "v0.8.26"
+const Version = "v0.8.28"


gontzess:
We fixed this, a ci token had failed so the action failed. This is moving to the new gh app token

PR #785, opened 2026-04-27
support hot loading configs for lambdas

kans:
## Summary

- Adds lambda connector config hot loading keyed by the `x-baton-connector-config-version` request header.
- Rebuilds the full connector generation on version changes so connector-local caches/state are discarded instead of reused.
- Swaps registered lambda gRPC service implementations in place and waits for in-flight requests using the old generation to drain before closing it.
- Adds connector close forwarding in `connectorbuilder` for connectors that implement `Close(context.Context) error` or `Close() error`.
- Adds shared lambda client secret context wiring used by downstream credential flows.

PR #787, opened 2026-04-28
Add managed runtime snapshot reload support

gontzess:
Why

Axiomatic hotload needs one SDK-level runtime contract that can carry the effective connector config and runtime assets without teaching the SDK about axiomatic-specific artifact semantics. The SDK also needs to rebuild, readiness-check, swap, and drain full connector generations when C1 reports a new effective snapshot revision.

What this changes

Adds a managed runtime snapshot API with opaque role-based assets and threads it through RunTimeOpts. Lambda mode now fetches the effective snapshot, falls back to the legacy config RPC when needed, builds connector generations from the snapshot, checks GetMetadata before serving, and swaps generations when the x-baton-runtime-revision hint changes.

Depends on follow-up C1 work to serve populated runtime assets.

### DIFF
--- a/pkg/cli/lambda_managed_runtime__added.go
+++ b/pkg/cli/lambda_managed_runtime__added.go
@@ -1,0 +1,227 @@
+//go:build baton_lambda_support
+
+package cli
+
+import (
+	"context"
+	"crypto/ed25519"
+	"encoding/json"
+	"fmt"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/conductorone/baton-sdk/pkg/crypto/providers/jwk"
+	c1_lambda_grpc "github.com/conductorone/baton-sdk/pkg/lambda/grpc"
+	"github.com/conductorone/baton-sdk/pkg/types"
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	v1 "github.com/conductorone/baton-sdk/pb/c1/connectorapi/baton/v1"
+)
+
+const lambdaConnectorDrainTimeout = 30 * time.Second
+
+type managedRuntimeFetcher struct {
+	client     v1.ConnectorConfigServiceClient
+	privateKey ed25519.PrivateKey
+}
+
+func newManagedRuntimeFetcher(client v1.ConnectorConfigServiceClient, privateKey ed25519.PrivateKey) *managedRuntimeFetcher {
+	return &managedRuntimeFetcher{
+		client:     client,
+		privateKey: privateKey,
+	}
+}
+
+func (f *managedRuntimeFetcher) Fetch(ctx context.Context) (*ManagedRuntimeSnapshot, error) {
+	resp, err := f.client.GetManagedRuntimeSnapshot(ctx, &v1.GetManagedRuntimeSnapshotRequest{})
+	if err != nil {
+		if status.Code(err) == codes.Unimplemented {
+			return f.fetchLegacyConfig(ctx)
+		}
+		return nil, err
+	}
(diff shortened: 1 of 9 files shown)


pkg/cli/lambda_server__added.go
```diff
@@ -271,10 +190,58 @@ func OptionallyAddLambdaCommand[T field.Configurable](
...
-		connector.Register(runCtx, s, c, opts)
+		buildGeneration := func(ctx context.Context, snapshot *ManagedRuntimeSnapshot) (*lambdaConnectorGeneration, error) {
+			t, cfgViper, authMethodStr, schemaFields, err := makeLambdaManagedRuntimeConfiguration[T](v, connectorSchema, snapshot)
+			if err != nil {
+				return nil, err
+			}
 
-		aws_lambda.StartWithOptions(s.Handler, aws_lambda.WithContext(runCtx))
+			ops := RunTimeOpts{
+				SessionStore:        sessionStore,
```
kans, line 200: The session store should be reconstructed, too.

pkg/cli/lambda_server__added.go
```diff
@@ -271,10 +190,58 @@ func OptionallyAddLambdaCommand[T field.Configurable](
...
+					client: configClient,
+				}
+			}
+
+			c, err := getconnector(ctx, t, ops)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get connector: %w", err)
+			}
+
+			if _, err := c.GetMetadata(ctx, &connectorV2.ConnectorServiceGetMetadataRequest{}); err != nil {
```
kans, line 221: Why do we get metadata?

pkg/cli/lambda_managed_runtime__added.go
```diff
@@ -0,0 +1,227 @@
...
+	snapshot := resp.GetSnapshot()
+	if snapshot == nil {
+		return nil, fmt.Errorf("managed runtime snapshot response was empty")
+	}
+
+	return f.decryptSnapshot(snapshot.GetRevision(), snapshot.GetEncryptedConfig(), snapshot.GetAssets())
+}
+
+func (f *managedRuntimeFetcher) fetchLegacyConfig(ctx context.Context) (*ManagedRuntimeSnapshot, error) {
+	resp, err := f.client.GetConnectorConfig(ctx, &v1.GetConnectorConfigRequest{})
```
kans, line 57: This isn't a legacy config, is it?

pkg/cli/lambda_managed_runtime__added.go
```diff
@@ -0,0 +1,227 @@
...
+
+func newManagedRuntimeFetcher(client v1.ConnectorConfigServiceClient, privateKey ed25519.PrivateKey) *managedRuntimeFetcher {
+	return &managedRuntimeFetcher{
+		client:     client,
+		privateKey: privateKey,
+	}
+}
+
+func (f *managedRuntimeFetcher) Fetch(ctx context.Context) (*ManagedRuntimeSnapshot, error) {
+	resp, err := f.client.GetManagedRuntimeSnapshot(ctx, &v1.GetManagedRuntimeSnapshotRequest{})
```
kans, line 40: This is a strange contract.  The snapshot contains an encrypted connector config, but we don't populate it here for some connectors for some reason (even though they are on a version of the SDK that made the call), forcing them to make a different call for the same data?  It would make more sense to me if we do one of the following:

1. We return empty data here for everyone, don't bundle the config in this call, and always do a *GetConnectorConfig*.
2. We populate the entire payload and never do a *GetConnectorConfig*.  That gRPC method will live on to serve connectors on older versions of the SDK.
3. Put the data into the config.

gontzess:
going with https://github.com/ConductorOne/baton-sdk/pull/785

PR #792, opened 2026-04-30
Task Parallelism Config

kans:


pkg/field/defaults.go
```diff
@@ -325,8 +329,28 @@ var (
 		WithInt(func(r *IntRuler) {
 			r.Gte(1).Lte(1800)
 		}))
+
+	// TaskConcurrencyField limits concurrent Baton task execution in the runner
+	// (service mode). Semantics match [WorkerCountField] / sync worker parallelism.
+	TaskConcurrencyField = IntField("task-concurrency",
+		WithDescription("The number of Baton tasks to run concurrently in service mode. "+
+			"-1 for auto-detect, 0 for sequential, >0 for parallel. Tasks may include sync, grant, revoke, and more."),
```
ggreer: I think we should make this logic simpler and just have --task-concurrency specify the number of slots. The default can then be 3 and allowed values would be 1 to some ridiculous value like 25.

ggreer (approved):

PR #795, opened 2026-05-01
Fix Completing Tasks with an Error!

kans:
… an error state. This caused cascading failures in C1


ggreer (approved):

kans:
Lint errors are unrelated, for test files

PR #812, opened 2026-05-09
fix: SendHeader TOCTOU + 2 config-write Sync/Close hardening sites

robert-chiniquy:
## What

Three sibling fixes from the same static-analysis sweep that produced [#811](https://github.com/conductorone/baton-sdk/pull/811).

### \`pkg/lambda/grpc/server.go::TransportStream.SendHeader\` (real concurrency bug)

\`\`\`go
func (r *TransportStream) SendHeader(md metadata.MD) error {
    if r.headersSent.Load() {           // <-- check outside mutex
        return ErrIllegalSendHeader
    }
    r.mtx.Lock()
    defer r.mtx.Unlock()
    r.headersSent.Store(true)            // <-- write inside mutex
    r.headers = metadata.Join(r.headers, md)
    return nil
}
\`\`\`

Two concurrent callers can both pass the unlocked \`Load()\` check before either takes \`r.mtx\`. They then serialize through the mutex but each performs \`metadata.Join\` on \`r.headers\`, double-joining the metadata. The second caller also fails to return \`ErrIllegalSendHeader\`, violating the once-only contract gRPC \`SendHeader\` is meant to enforce.

Fix: re-check \`headersSent.Load()\` INSIDE the mutex (standard double-checked locking for an atomic + critical section).

### \`pkg/cli/service_windows.go::interactiveSetup\` (config-write hardening)

Bare \`defer f.Close()\` on \`os.Create\` followed by \`yaml.NewEncoder(f).Encode(config)\`. Close error silently discarded; user sees \"Config file created.\" log line for a possibly-torn config they then try to load on next start.

Same fix shape as #811: \`fClosed\` flag, explicit \`Sync\` + \`Close\` + err-check on the success path, deferred Close as safety net for error paths.

### \`pkg/config/generate.go::Generate\` (code-gen-output hardening)

Code generator that produces \`conf.gen.go\`. Bare \`defer f.Close()\` after \`tmpl.Execute\`. A torn \`conf.gen.go\` shows up as a downstream build failure rather than a clear generator panic; surface the flush error here.

Same fix shape.

## How found

[Static analysis](https://github.com/ductone/occult-go-analysis) detectors:
- \`read_modify_write_without_txn\` — flagged the SendHeader site (atomic Load outside, Store inside, no in-mutex re-check)
- \`error_swallowed_in_deferred_close\` + \`write_without_atomic_rename\` — flagged the two config-write sites

## Test plan

- [ ] \`go build ./pkg/lambda/grpc/... ./pkg/config/...\` (passed)
- [ ] \`GOOS=windows go build ./pkg/cli/...\` (passed)
- [ ] Existing tests pass
- [ ] No semantic change on success paths
- [ ] On Sync/Close failure: error surfaces before the function reports success

## Related

- [#811](https://github.com/conductorone/baton-sdk/pull/811) — first batch from this sweep (SaveC1Z + cpFile)


pkg/lambda/grpc/server.go
```diff
@@ -95,6 +95,15 @@ func (r *TransportStream) SendHeader(md metadata.MD) error {
 	}
 	r.mtx.Lock()
 	defer r.mtx.Unlock()
+	// Re-check inside the mutex: between the unlocked Load above and
+	// taking r.mtx, a concurrent caller may have already set
+	// headersSent and joined headers. Without this check, both
+	// callers reach the Store(true)/Join below and end up
+	// double-joining headers; the second caller would also fail to
+	// return ErrIllegalSendHeader, violating the once-only contract.
+	if r.headersSent.Load() {
```
kans, line 107: SetTrailer/SetHeader also have this problem.

robert-chiniquy: Good point

kans (approved):

PR #813, opened 2026-05-11
WIP/spike (parked): perf(dotc1z): transparent-paging Stream variants [OPS-1483]

arreyder:
> **🚧 WIP / EXPERIMENT — DRAFT, NOT FOR REVIEW.** First-cut spike for OPS-1483. Numbers and design are real and tests pass, but this is an experiment to validate the transparent-paging approach before committing to the c1-side migration. Do not merge. Feedback welcome; design may change.

## Summary

Adds three streaming methods on `*C1File`:

- `ListGrantsForEntitlementStream(ctx, req, scan func(*v2.Grant) error) error`
- `ListGrantsForPrincipalStream(ctx, req, scan func(*v2.Grant) error) error`
- `ListGrantsForResourceTypeStream(ctx, req, scan func(*v2.Grant) error) error`

Each iterates the matching grants via the `scan` callback, transparently paging through the underlying `grants` table in **1000-row batches**.

**Peak heap during iteration is bounded to ~1000 decoded grants** (independent of total grant count), vs up to `V2MaxPageSize = 10000` in the existing slice-returning RPCs. The new methods close the SQL cursor between batches, so `scan` may safely call back into other `*C1File` methods (e.g. `GetResourceType`) without deadlocking against the pinned writer connection — the failure mode that closed [PR #619](https://github.com/ConductorOne/baton-sdk/pull/619).

This is an additive change. Existing `ListGrants*` slice RPCs are untouched; callers migrate at their own pace.

## Why now

Sibling to (see internal platform-repo PR) — the "Address memory and GC issues in Temporal Worker" project. OPS-1193's streaming PRs bound the *caller-side* slice accumulators on the uplift hot path; this PR bounds the in-baton-sdk per-page proto buffer that those callers can't control.

Per the platform repo stored memory: `proto.Unmarshal(*v2.Grant)` via `fetchGrantsForEntitlementV2 → ListGrantsForEntitlement → listConnectorObjects` is 0.96 % inuse / 8.68 GB cum on `be-temporal-worker`. Page size of 1000 gives a 10× peak-heap reduction for the per-call buffer with manageable latency cost on realistic-size entitlements.

## Benchmark

`go test -bench '^BenchmarkListGrantsForEntitlement(Stream)?$' -benchtime=3x` on AMD Ryzen 7 5700X3D:

| N grants  | Slice ns/op | Stream ns/op | Slice B/op  | Stream B/op | Δ ns/op    |
|-----------|------------:|-------------:|------------:|------------:|------------|
| 100       |     510 µs  |     454 µs   |    110 KB   |    120 KB   | **-11 %**  |
| 1,000     |   3.45 ms   |   3.46 ms    |   1.05 MB   |   1.05 MB   |   ≈ 0 %    |
| 10,000    |  34.9 ms    |  86.5 ms     |  10.64 MB   |  10.58 MB   | 2.5× slower |
| 100,000   |   893 ms    |    6.2 s     |   106 MB    |   106 MB    | 7× slower |

- **Total allocations are within 1 %** across all N — both paths decode the same total number of protos.
- **Latency penalty at extreme N is the expected price** for bounding peak heap. The design goal is OOM-prevention on outsized entitlements, not raw throughput. Real-world median entitlements are well under 10K grants, where the cost is ≤ 2.5×.
- The per-query overhead in pure-Go `modernc.org/sqlite` is what amortizes badly at small page sizes; a single round trip costs the same whether you fetch 100 or 10,000 rows.

## Open questions before this leaves draft

- [ ] Is **page size 1000** the right default, or should it be exposed as a parameter / option? Different call sites may want different tradeoffs.
- [ ] Should we add the **dedicated read-only `*sql.DB` pool** in the same PR, or punt to a follow-up? Punted in this draft because `wal_checkpoint(TRUNCATE)` coordination needs design care.
- [ ] Do we want to **eliminate the slice variants** once consumers migrate, or keep them indefinitely as the back-compat path?
- [ ] **Server-streaming proto RPC** — out of scope here, but worth a design call: does the platform repo's `fetchGrantsForEntitlementV2` consume this as a Go API on `*C1File`, or do we expose it through `c1/reader/v2`?

## Test plan

- [x] `TestListGrantsForEntitlementStream_ParityWithSlice` — at 2500 grants (multi-page), streaming visits the exact same grants in the same order as the paginated slice path.
- [x] `TestListGrantsForEntitlementStream_ScanErrorPropagates` — scan error halts iteration immediately, error bubbles up.
- [x] `TestListGrantsForEntitlementStream_NilScanRejected` — nil scan is rejected with an error.
- [x] `TestListGrantsForEntitlementStream_NoDeadlockOnReentrantCall` — **regression test for PR #619's failure mode.** scan calls `GetResourceType` from inside the iterator across all 2500 grants; no deadlock against the single-conn pin.
- [ ] CI green
- [ ] Owner / design review

## Rollout (proposed — subject to design feedback)

- Additive only — no existing callers are affected.
- Tag a baton-sdk release once merged, then `go.mod` bump in `the platform repo`.
- Then `fetchGrantsForEntitlementV2` in `pkg/temporal/activity/app/fetch_v2.go` migrates to the streaming RPC under a feature flag (separate the platform repo PR).

## Out of scope (deferred — see Linear)

- Dedicated read-only `*sql.DB` pool. Real design problem (`wal_checkpoint(TRUNCATE)` coordination) that doesn't need to block this fix.
- Server-streaming proto RPC for `ListGrantsForEntitlement`. The new Stream methods are regular Go APIs on `*C1File`; that's enough for the in-process consumer in `the platform repo`.
- FST + roaring sidecar projection. Re-evaluate only if 2a + the platform repo migration still leaves meaningful alloc-rate pressure.

Linear: https://linear.app/ductone/issue/OPS-1483

### DIFF
--- a/pkg/dotc1z/grants_bench_test.go
+++ b/pkg/dotc1z/grants_bench_test.go
@@ -139,3 +139,45 @@ func BenchmarkListGrantsForEntitlement(b *testing.B) {
 		})
 	}
 }
+
+// BenchmarkListGrantsForEntitlementStream is the streaming counterpart to
+// BenchmarkListGrantsForEntitlement (OPS-1483). Compare alloc/op and ns/op
+// across the two paths. Total bytes allocated should be similar — both decode
+// the same N protos overall — but the streaming variant keeps only ~100 decoded
+// grants resident at a time, regardless of N. The peak-inuse difference is not
+// captured by standard benchmark metrics; see the issue for measurement notes.
+func BenchmarkListGrantsForEntitlementStream(b *testing.B) {
+	for _, numGrants := range grantCounts {
+		b.Run(fmt.Sprintf("grants=%d", numGrants), func(b *testing.B) {
+			f, entitlementID, cleanup := setupBenchmarkDB(b, numGrants)
+			defer cleanup()
+
+			ctx := b.Context()
+
+			entResp, err := f.GetEntitlement(ctx, reader_v2.EntitlementsReaderServiceGetEntitlementRequest_builder{
+				EntitlementId: entitlementID,
+			}.Build())
+			require.NoError(b, err)
+
+			req := reader_v2.GrantsReaderServiceListGrantsForEntitlementRequest_builder{
+				Entitlement: entResp.GetEntitlement(),
+			}.Build()
+
+			b.ReportAllocs()
+
+			for b.Loop() {
+				totalGrants := 0
+				err := f.ListGrantsForEntitlementStream(ctx, req, func(g *v2.Grant) error {
+					totalGrants++
+					return nil
+				})
+				if err != nil {
+					b.Fatal(err)
+				}
+				if totalGrants != numGrants {
+					b.Fatalf("expected %d grants, got %d", numGrants, totalGrants)
+				}
+			}
+		})
+	}
+}
--- a/pkg/dotc1z/grants_stream.go
+++ b/pkg/dotc1z/grants_stream.go
@@ -1,0 +1,159 @@
+package dotc1z
+
+import (
+	"context"
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+
+	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
+	reader_v2 "github.com/conductorone/baton-sdk/pb/c1/reader/v2"
+	"github.com/conductorone/baton-sdk/pkg/uotel"
+)
+
+// grantStreamPageSize is the internal page size used by the streaming grant readers.
+// Bounds peak heap to ~pageSize × sizeof(*v2.Grant) per call; the constant trades
+// heap-bounding strength against query overhead. modernc.org/sqlite (pure-Go) has
+// a meaningful per-query cost, so very small pages (<500) make the streaming
+// variant dramatically slower than the slice variant on large entitlements. 1000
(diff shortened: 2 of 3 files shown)


arreyder:
**Parking this PR as a spike.** Pulled a fresh \`be-temporal-worker\` prod heap profile (2026-05-11) and the picture has shifted: **\`proto.Unmarshal(*v2.Grant)\` is no longer in the top 15 alloc paths.** The 11 PRs under OPS-1193 successfully dealt with the original bottleneck.

The new dominant allocator is **CEL program compilation** in \`pkg/xcel\` — ~9-10 GB / 32.5 GB ≈ 30% of total alloc rate, called from \`pkg/mapper/user/attributes.go:91\` (\`evaluateUserAttributeCELExpression\`). No program cache exists; every user × every mapping × every evaluation re-compiles the same handful of expression texts.

Pivoting effort to the CEL cache (tracked in a new ductone/c1 Linear issue). This spike stays parked-as-draft for reference — the transparent-paging design is correct for if/when grant unmarshal returns to the hot list, but shipping it now would be solving last quarter's problem.

arreyder:
Closing — OPS-1483 cancelled. The transparent-paging streaming approach was a parked spike ("WIP/spike (parked)" in the title) and the broader effort is no longer being pursued in this shape. If the dotc1z streaming need returns we'll open a fresh PR from current main.

PR #817, opened 2026-05-13
Update release workflow for baton cli utility.

ggreer:


.github/workflows/release.yaml
```diff
@@ -7,14 +7,55 @@ on:
...
+        uses: sigstore/cosign-installer@v3
+
+      - name: Download syft
+        uses: anchore/sbom-action/download-syft@v0
+
+      - name: Configure AWS credentials via OIDC
+        uses: aws-actions/configure-aws-credentials@v5
+        with:
+          role-to-assume: arn:aws:iam::025044153841:role/GHA-Artifacts-${{ github.event.repository.owner.login }}-${{ github.event.repository.name }}
+          aws-region: us-west-2
```
gontzess: we should remove this step, were not uploading or recording this to the dist

.github/workflows/release.yaml
```diff
@@ -7,14 +7,55 @@ on:
...
-    secrets:
-      RELENG_GITHUB_TOKEN: ${{ secrets.RELENG_GITHUB_TOKEN }}
-      APPLE_SIGNING_KEY_P12: ${{ secrets.APPLE_SIGNING_KEY_P12 }}
-      APPLE_SIGNING_KEY_P12_PASSWORD: ${{ secrets.APPLE_SIGNING_KEY_P12_PASSWORD }}
-      AC_PASSWORD: ${{ secrets.AC_PASSWORD }}
-      AC_PROVIDER: ${{ secrets.AC_PROVIDER }}
-      DATADOG_API_KEY: ${{ secrets.DATADOG_API_KEY }}
+    runs-on: macos-latest
+    steps:
+      - name: Mint baton-ci app token
```
gontzess, line 12: this is the only step im not sure on if it will work, but we'll know right away when we try cutting a release

gontzess (approved):

PR #821, opened 2026-05-13
feat: add daily log rotation with gzip compression and configurable retention

MarcusGoldschmidt:


leet-c1:
Heads up: log rotation has been reworked as a standalone PR — **#746 (`feat: add size-based log rotation via lumberjack`)** — so this and #746 now fully overlap (both touch only the logging package + flag wiring).

#746 covers the same goal as this PR (SDK-level rotation + gzip compression + retention, wired into `logging.Init` and the Windows service default), but as a **vended implementation** rather than a hand-rolled rotator:

- Rotation is **size-based**, backed by [lumberjack](https://github.com/natefinch/lumberjack) (`gopkg.in/natefinch/lumberjack.v2`) — the standard Go rotation library zap recommends. That removes the custom `DailyRotator` / `rotation.go` (~900 lines incl. tests) and, with it, the bespoke calendar-day/DST/timezone/collision/concurrency handling we'd otherwise own and maintain.
- Size-based rotation also maps directly to the actual problem (unbounded `baton.log` growth) instead of per-calendar-day files.
- Config: `--log-file`, `--log-max-size-mb`, `--log-max-backups`, `--log-retention-days` — defaults (100 MB × 10) bound rotated history to ~1 GB; retention is enforced by whichever of count/age is hit first.
- It's squashed to a single reviewable commit and has been through code review plus an adversarial pass.

Since the two are duplicate/competing implementations, I'd suggest **closing this PR in favor of #746** to avoid divergence. Flagging here so the work and discussion aren't lost — happy to fold in anything from this branch that #746 doesn't already cover. cc @MarcusGoldschmidt

PR #826, opened 2026-05-14
Add index for entitlement + sync ID in grants table.

ggreer:


pkg/dotc1z/grants.go
```diff
@@ -38,7 +38,8 @@ create table if not exists %s (
 create index if not exists %s on %s (resource_type_id, resource_id);
 create index if not exists %s on %s (principal_resource_type_id, principal_resource_id);
 create index if not exists %s on %s (entitlement_id, principal_resource_type_id, principal_resource_id);
-create unique index if not exists %s on %s (external_id, sync_id);`
+create unique index if not exists %s on %s (external_id, sync_id);
+create index if not exists %s on %s (entitlement_id, sync_id);`
```
kans: we need the grant id in here, too, or else we have to order it on the fly.

kans:
How long does creating the index take on a big c1z that wasnt' created with the index?  (is this safe).  We could have a system to only add new indexes to fresh c1zs.

kans (approved):

PR #830, opened 2026-05-15
dotc1z: add OTel spans on C1File.Close, saveC1z, rawDb.Close

arreyder:
## Summary

Today only the inner WAL truncate is spanned (`C1File.truncateWAL`). The outer `C1File.Close`, the pure-Go SQLite handle close, and `saveC1z` (zstd recompression of the database back into a `.c1z` file) are all unspanned. We cannot answer "is closing the c1z reader slow?" with a single Datadog query — the hypothesis raised by Geoff Greer during a recent prod investigation.

Wrap with `tracer.Start` + `uotel.EndSpanWithError`:

- `C1File.Close` (outer) — top-level operation
- `C1File.closeRawDB` — pure-Go SQLite handle close (the normal-path one at line 430; error-path closes at 406/420 stay direct since they're already inside the parent Close span and only fire on WAL checkpoint failure)
- `C1File.saveC1z` — zstd recompression on writer path

Attributes on `Close`: `read_only`, `db_updated`, `db_path`. Attributes on `saveC1z`: `db_path`, `output_path`, `encoder_concurrency`, `input_size_bytes` (from `os.Stat` of the SQLite file before compression).

No behavior changes; `saveC1z` keeps its current signature.

## Test plan

- [ ] CI green
- [ ] After vendor-bump into c1 and deploy to prod-usw2, verify spans appear in be-temporal-sync:
  - `@operation_name:C1File.Close env:prod-usw2 service:be-temporal-sync`
  - `@operation_name:C1File.saveC1z env:prod-usw2`
- [ ] Compare p99 duration of `C1File.Close` for `@db_updated:true` (writer/sync) vs `@db_updated:false` (read-only/uplift). Geoff's hypothesis is that the writer-side close (saveC1z driven) is the slow one.

## Note on pre-existing test failures

`TestPutResources` and `TestGrantStatsGroupByParity` fail on plain `main` too with `unable to open database file (14)` — pre-existing flake in this repo's test environment, unrelated to this change.

## Companion PR (c1)

A companion platform-repo PR adds spans around the c1-side wrappers (`connectorReaderCache.{Get, Close, evictApp, evictConnector}`, `runStore.{getOrFetchArtifact, fetchArtifact, Close}`). The two together give us end-to-end visibility on the uplift reader-close path.

Refs OPS-1564.

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/doug-martin/goqu/v9"
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -358,7 +359,10 @@ var checkpointTimeout, _ = strconv.ParseInt(os.Getenv("BATON_WAL_CHECKPOINT_TIME
 // with our changes. The provided context is used for the WAL checkpoint operation. If the context is already expired,
 // a fresh context with a timeout is used to ensure the checkpoint completes.
 func (c *C1File) Close(ctx context.Context) error {
+	ctx, span := tracer.Start(ctx, "C1File.Close")
 	var err error
+	defer func() { uotel.EndSpanWithError(span, err) }()
+
 	l := ctxzap.Extract(ctx)
 
 	c.closedMu.Lock()
@@ -367,6 +371,12 @@ func (c *C1File) Close(ctx context.Context) error {
 		l.Warn("close called on already-closed c1file", zap.String("db_path", c.dbFilePath))
 		return nil
 	}
+
+	span.SetAttributes(
+		attribute.Bool("read_only", c.readOnly),
+		attribute.Bool("db_updated", c.dbUpdated),
+		attribute.String("db_path", c.dbFilePath),
+	)
 
 	if c.rawDb != nil {
 		// CRITICAL: Force a full WAL checkpoint before closing the database.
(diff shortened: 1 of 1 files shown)


arreyder:
Review pass found one Medium: the deferred `uotel.EndSpanWithError(span, err)` closure was capturing the outer `var err error` by reference, but five early-return paths returned wrapped errors without updating that variable — so the span would have recorded `codes.Ok` for the very failure modes we're trying to instrument (WAL checkpoint incomplete, etc.).

Fixed in [a568f833](https://github.com/ConductorOne/baton-sdk/commit/a568f8334e8b32b46e2b91bb09f7e63e07fa2bc7) by converting `Close` to use a named return. No behavior change; only fixes what the span records.

kans (approved):

arreyder:
Picked up @ggreer's feedback from [#829](https://github.com/ConductorOne/baton-sdk/pull/829) (parallel attempt at the same change). Both concerns addressed in [b6ee08c1](https://github.com/ConductorOne/baton-sdk/commit/b6ee08c167dd3e5d6b0d6038e57d35feba51b5e4):

1. **"Should we set the rawDb to nil at this point?"** — moved `c.rawDb = nil` and `c.db = nil` into `closeRawDB` so the helper owns the full "after close, drop handles" pattern. Removed three duplicated nil-out blocks at the call sites. As a bonus, the WAL-error paths can now use the helper too (they were calling `c.rawDb.Close()` directly and missing the span).

2. **"Maybe we can inline saveC1z? I'm not sure why it's a standalone function instead of a method on the c1file."** — `runSaveC1z` was just a wrapper. The underlying `saveC1z` in `file.go` is a free function used by tests with arbitrary paths; converting it to a method would churn the tests. So I inlined the span at the single call site in `Close` instead and removed the helper.

Net 19 insertions / 40 deletions. No behavior change.

ggreer (approved):

PR #831, opened 2026-05-15
dotc1z: add WithV2GrantsWriter to local manager

btipling:
SDK plumbing for the per-connector slim-grants rollout. Adds a local manager option so c1's sync activity can toggle slim per connector instance, a normalization helper so `baton diff` treats slim and full grants for the same logical row as equal, and two counters to confirm a flip is actually producing slim rows.

`resolveSyncIDForPayloadQuery` now returns `sql.ErrNoRows` instead of silently running an unscoped grants query when no sync can be resolved. Mirrors the sibling `resolveSyncIDForInternalQuery`; small behavior change for callers reading c1z files with no completed syncs.


pkg/dotc1z/grants_expandable_query.go
```diff
@@ -170,9 +170,7 @@ func (c *C1File) listGrantsWithExpansionInternal(ctx context.Context, opts grant
 		q = q.Where(goqu.C("resource_type_id").Eq(resource.GetId().GetResourceType()))
 	}
 
-	if syncID != "" {
-		q = q.Where(goqu.C("sync_id").Eq(syncID))
-	}
+	q = q.Where(goqu.C("sync_id").Eq(syncID))
```
btipling: This is and below is an unrelated bug fix that would lead to an unscoped query. It's minor, flagged by the bot

ggreer: resolveSyncIDForPayloadQuery() can return empty string, so won't this return no rows if that is the case? For example, if there are no completed syncs in the c1z, but part of a sync did run and was saved.

btipling: It's unclear to me when a sync would not have a sync id. I assumed it always does, and with the new sync sessions it needs a sync id. This is not really part of the change so I can remove it if this is risky.

ggreer: I think it's risky. I'd keep the empty check.

ggreer (approved):

PR #832, opened 2026-05-18
[CE-720 ]add license traits

Bencheng21:
tested in https://github.com/ConductorOne/baton-github/pull/157/changes

### DIFF
--- a/proto/c1/connector/v2/annotation_trait.proto
+++ b/proto/c1/connector/v2/annotation_trait.proto
@@ -154,7 +154,7 @@ message LicenseProfile {
   // Consumed seats from the vendor API. Zero when not available.
   int64 consumed_seats = 3;
   // Optional per-seat cost from the vendor API. Zero when not available.
-  double cost_per_unit = 4;
+  int64 cost_per_unit_in_cents = 4;
   // ISO 4217 currency code. Defaults to "USD" when unset.
   string currency = 5 [(validate.rules).string = {
     max_bytes: 8
--- a/proto/c1/connector/v2/annotation_trait.proto
+++ b/proto/c1/connector/v2/annotation_trait.proto
@@ -132,3 +132,34 @@ message SecretTrait {
   connector.v2.ResourceId created_by_id = 5;
   connector.v2.ResourceId identity_id = 6;
 }
+
+// LicenseTrait is the trait annotation for resources with TRAIT_LICENSE. A
+// license resource represents a vendor-defined seat or plan that users on
+// the connector can hold (e.g. Microsoft "E3", Atlassian "Business Plus").
+// The entitlements listed in profile.entitlement_ids identify, on this
+// connector, the grants that mean "the principal holds this license seat" —
+// typically the membership entitlement of a license-backing group/role.
+message LicenseTrait {
+  LicenseProfile profile = 1;
+}
+
+message LicenseProfile {
+  // Human-readable license name (e.g. "Enterprise", "Business Plus", "E3").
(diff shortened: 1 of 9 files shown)


proto/c1/connector/v2/annotation_trait.proto
```diff
@@ -132,3 +132,34 @@ message SecretTrait {
...
+  string license_name = 1 [(validate.rules).string = {
+    max_bytes: 1024
+    ignore_empty: true
+  }];
+  // Total seats from the vendor API. Zero when not available.
+  int64 purchased_seats = 2;
+  // Consumed seats from the vendor API. Zero when not available.
+  int64 consumed_seats = 3;
+  // Optional per-seat cost from the vendor API. Zero when not available.
+  double cost_per_unit = 4;
```
ggreer: I think this should be an integer storing the cost in cents. Floating point can lead to precision errors. For example, if you have 213 licenses that cost $1.30, using floating point will give you a total cost of $276.90000000000003.

proto/c1/connector/v2/annotation_trait.proto
```diff
@@ -132,3 +132,34 @@ message SecretTrait {
...
 }
+
+// LicenseTrait is the trait annotation for resources with TRAIT_LICENSE. A
+// license resource represents a vendor-defined seat or plan that users on
+// the connector can hold (e.g. Microsoft "E3", Atlassian "Business Plus").
+// The entitlements listed in profile.entitlement_ids identify, on this
+// connector, the grants that mean "the principal holds this license seat" —
+// typically the membership entitlement of a license-backing group/role.
+message LicenseTrait {
+  LicenseProfile profile = 1;
```
ggreer: Why put all the license info in a profile attribute instead of just putting it on the LicenseTrait? On other resource traits, we usually make the profile a struct that is a grab bag of miscellaneous data.

Bencheng21: https://www.notion.so/conductorone/MVP-Spec-App-Utilization-Connector-Native-3454694ad8468148b12efa7d0f8913c2#44e4f80871d84d3599642b73b61a6fb6

In the docs, it’s called LicenseProfile. I think in c1, we would have a struct called `licenceProfile` as well. While it technically doesn’t need to contain all license information within the profile attributes, I think it’s better to keep the naming consistent across components.

incremental-sync is another good example. In baton-sdk, we call it targetedSync, while in C1 we use incrementalSync for the same concept. I’d rather avoid having two different names for the same thing.

proto/c1/connector/v2/annotation_trait.proto
Bencheng21: hm, Changing to cost in cents doesn’t feel very user-friendly since users now have to multiply everything by 100 for connectors. 

> if you have 213 licenses that cost $1.30, using floating point will give you a total cost of $276.90000000000003.
We should probably round or limit the result to 2 decimal places in c1 side, IMO

but i updated it with your suggestion

proto/c1/connector/v2/annotation_trait.proto
ggreer: I'd call it LicenseProfileTrait then, and put the profile info directly on it.

The original name was partial sync, which we use consistently in baton-sdk... except the FullSyncTask, which got partial sync shoehorned into it. I don't know why the name became incremental sync in C1. It's more syllables. Maybe because that feature uses both event feeds & partial syncs to update the data?

ggreer (approved):

PR #833, opened 2026-05-19
Add opaque connector config version

gontzess:
**Why**

C1 needs to stop treating connector config versions as timestamps once runtime-backed Lambdas use logical runtime state. The SDK already compares invoke versions as strings, but cold start still initialized the current generation from `last_updated`, forcing C1 to keep returning timestamp-shaped versions.

**What this changes**

Adds an opaque `version` field to `GetConnectorConfigResponse` and makes Lambda cold start prefer it when present. Older servers remain compatible because the SDK still falls back to `last_updated`.

**Testing**

- `make protogen`
- `buf lint`
- `buf breaking --against 'https://github.com/conductorone/baton-sdk.git#branch=main'`
- `go test -tags=baton_lambda_support -v ./pkg/cli ./pkg/lambda/grpc`
- `git diff --check`

`make lint` currently reports existing gosec findings outside this change.

### DIFF
--- a/proto/c1/connectorapi/baton/v1/config.proto
+++ b/proto/c1/connectorapi/baton/v1/config.proto
@@ -18,6 +18,7 @@ message GetConnectorConfigRequest {
 message GetConnectorConfigResponse {
   bytes config = 1;
   google.protobuf.Timestamp last_updated = 2;
+  string version = 3;
 }
 
 message SignedHeader {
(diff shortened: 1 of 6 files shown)


proto/c1/connectorapi/baton/v1/config.proto
```diff
@@ -18,6 +18,7 @@ message GetConnectorConfigRequest {
 message GetConnectorConfigResponse {
   bytes config = 1;
   google.protobuf.Timestamp last_updated = 2;
+  string version = 3;
```
ggreer, line 21: Should we add some validation rules to this to make it a semver? I don't have a strong opinion here. The regex could be too complicated and/or restrictive.

ggreer (approved):

proto/c1/connectorapi/baton/v1/config.proto
kans: Its better to control that on the other side IMO.

gontzess:
not strictly necessary at the moment, can reopen if that changes

PR #835, opened 2026-05-19
CE-721: add crypto.GenerateRandomPasswordWithPolicy for connector-local password policies

cvkramer:
## Summary

- Adds `crypto.GenerateRandomPasswordWithPolicy(rp, policy)` — same algorithm as the existing `GenerateRandomPassword` plus excluded-character filtering, length-bound enforcement, and required-class coverage per the supplied policy.
- Adds `RandomPasswordPolicy` struct, `PasswordCharClass` enum, `PolicyEnforcement` enum (`Strict` / `Clamp` / `WarnOnly`), and sentinel errors `ErrPolicyViolation` and `ErrPolicyConflict`.
- Exports the character-class constants (`UpperCaseLetters`, `LowerCaseLetters`, `Digits`, `Symbols`) so callers can reason about the alphabet without duplicating literals. The previously-unexported names are kept as internal aliases — no source-level breakage in this package.

Linear: [CE-721](https://linear.app/ductone/issue/CE-721/baton-sdk-add-cryptogeneraterandompasswordwithpolicy-for-connector). Closes the dead-field bug fixed at the connector level in [CXP-543](https://linear.app/ductone/issue/CXP-543) and unblocks a similar fix in baton-nosql / baton-http.

## Why connector-local, not proto / wire

The policy describes how a connector wants random passwords generated **when it has to generate one** — the default `CreateAccount` path with `LocalCredentialOptions_RandomPassword`. C1's workflow `Generate Password` step generates server-side per its own `GeneratePasswordPolicy` and hands the connector a pre-baked password as `EncryptedPassword`; that path doesn't ask the connector to generate, so it doesn't need to know the connector's policy. No proto changes, no wire changes, no C1 backend involvement.

Earlier RFC drafts that put the policy on the wire (RFC-A through RFC-E in the design discussion on CE-721) are explicitly deferred. This PR is the minimum, connector-local subset.

## Backward compatibility

- `GenerateRandomPassword` is **unmodified**. Existing callers see zero change.
- `GenerateRandomPasswordWithPolicy(rp, nil)` is behavior-equivalent to `GenerateRandomPassword(rp)`.
- The internal `upperCaseLetters` / `lowerCaseLetters` / `digits` / `symbols` names are preserved as aliases for the new exported constants — the existing `password_test.go` references compile unchanged.

## Tests

`pkg/crypto/policy_test.go` adds 7 test functions and ~25 subtests:

- **Nil and zero-value policy** — confirms back-compat with `GenerateRandomPassword`.
- **Excluded characters** — 200-iteration assertion that disallowed runes never appear; covers constraint filtering and the "every char excluded" conflict.
- **Length enforcement** — `Strict` rejects with `ErrPolicyViolation`; `Clamp` clamps with the right side; `WarnOnly` proceeds; bounded values pass.
- **Required classes** — replace-vs-additive semantics covered; conflict with `ExcludedCharacters` returns `ErrPolicyConflict`; `PasswordCharClassUnspecified` returns a tailored error.
- **Combined realistic policy** — min/max + excluded + required classes, all together.
- **Exported constants** — smoke test that values match the original literals.

Existing `TestGeneratePassword` / `TestConvertCredentialOptions` continue to pass unmodified.

## End-to-end validation (off the PR; performed locally)

To prove the SDK helper is sufficient by itself, I checked out baton-sql `main` in a separate worktree, added a `go.work` pointing at this branch, and wired the helper into baton-sql's `generatePassword` with ~30 lines of changes (no local algorithm, no `pkg/bsql/password.go`). Then ran a battery of live tests against a local Postgres 17:

| Live test | Result |
|---|---|
| `disallowed_characters` honored end-to-end (5 iters) | PASS |
| Length bounds: within / over-max / under-min / combined / unspecified | PASS (5/5) |
| Bcrypt credentials roundtrip — generated password authenticates, wrong password fails (10 iters) | PASS |
| Postgres native `CREATE ROLE … LOGIN PASSWORD` with SCRAM-SHA-256 verifier reconstruction, 5 policy scenarios (alphanumeric / no-ambiguous / no-digits / required-letters-only / no-SQL-meta) | PASS (5/5) |
| `Rotate` path — new password authenticates, old no longer does | PASS |
| Password containing SQL meta-chars (`'\";\\`) handled through parameterized queries (20 iters) | PASS |
| 50 distinct generations under the same policy (no PRNG degeneracy) | PASS |

The downstream baton-sql / baton-nosql / baton-http migrations are separate PRs.

## Out of scope

- Proto changes.
- C1 backend involvement.
- Wire-level enforcement / merge between connector and workflow policies (deferred per the design discussion).
- Other credential types (`EncryptedPassword`, `PlaintextPassword`, `SSO`, `NoPassword`).
- Connector migrations.

### DIFF
--- a/pkg/crypto/password.go
+++ b/pkg/crypto/password.go
@@ -11,12 +11,22 @@ import (
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
 )
 
+// Character classes used by random password generation. Exposed so callers
+// (e.g., connectors building their own policies) can reason about the alphabet
+// without duplicating the literals.
 const (
-	upperCaseLetters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-	lowerCaseLetters = "abcdefghijklmnopqrstuvwxyz"
-	digits           = "0123456789"
-	symbols          = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~"
-	characters       = upperCaseLetters + lowerCaseLetters + digits + symbols
+	UpperCaseLetters = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	LowerCaseLetters = "abcdefghijklmnopqrstuvwxyz"
+	Digits           = "0123456789"
+	Symbols          = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~"
+
+	// Internal aliases. Kept lowercase to avoid touching unrelated call sites
+	// in this package; new code should reference the exported names above.
+	upperCaseLetters = UpperCaseLetters
+	lowerCaseLetters = LowerCaseLetters
+	digits           = Digits
+	symbols          = Symbols
+	characters       = UpperCaseLetters + LowerCaseLetters + Digits + Symbols
 )
 
 var ErrInvalidCredentialOptions = errors.New("unknown credential options")
--- a/pkg/crypto/policy.go
+++ b/pkg/crypto/policy.go
@@ -1,0 +1,295 @@
+package crypto //nolint:revive,nolintlint // we can't change the package name for backwards compatibility
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
+)
+
+// RandomPasswordPolicy is a connector-local policy describing how a random
+// password should be generated. It is intentionally NOT a proto type — it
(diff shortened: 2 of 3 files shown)


JavierCarnelli-ConductorOne:
This was originally made because of the issue: [CXP-543](https://linear.app/ductone/issue/CXP-543/baton-sql-honor-disallowed-characters-in-random-password-generation)

The issue was addressed with a different approach. Changes were delivered to the connector. PR: https://github.com/ConductorOne/baton-sql/pull/128

PR #842, opened 2026-05-20
fix: 4 latent bugs surfaced by occult-go-analysis 2026-05-19 scan

robert-chiniquy:
Four real bugs found by [occult-go-analysis](https://github.com/ductone/occult-go-analysis)'s full-scan of baton-sdk on 2026-05-19.

### 1. \`pkg/lambda/grpc/config/config.go::lambdaHTTPClient\`

The constructed \`*http.Client\` had no \`Timeout\`. A hung lambda endpoint would hang the caller forever. Added an explicit 30s timeout, matching the budget used elsewhere in this SDK (\`pkg/uhttp/transport.go\`).

### 2. \`pkg/healthcheck/server.go::Start\`

The \`Serve\` goroutine had no \`defer recover()\` -- a panic in any \`/healthz\` handler crashed the whole connector. Added a recover wrapper that logs the panic + stack via the existing zap logger (pattern matched from \`pkg/ugrpc/interceptors.go\`).

### 3. \`pkg/uhttp/dbcache.go::load\`

\`d.location\` was assigned **before** \`sql.Open\` was attempted, leaving the receiver half-initialised on error (location set, db nil). A caller retrying \`load\` (or inspecting \`.location\` after the error) would see a broken state. Reordered so \`location\`, \`db\`, and \`rawDb\` are committed together only after \`sql.Open\` succeeds.

### 4. \`pkg/dotc1z/grants.go::backfillGrantExpansionColumn\`

Two \`*sql.Rows\` values were manually closed in every exit branch but never deferred -- a panic mid-iteration would leak the underlying sql connection. Replaced the manual epilogues with deferred Close. The per-page Rows inside the for-loop is scanned in an inline closure so \`Close\` fires per iteration rather than accumulating defers across pages.

### How found

occult-go-analysis baton-sdk full-scan (2026-05-19). Detector kinds: \`http_client_literal_without_timeout\`, \`goroutine_without_recover\`, \`abstract_state_leak_on_error_path\`, \`sql_rows_close_missing\`.

### DIFF
--- a/pkg/healthcheck/server.go
+++ b/pkg/healthcheck/server.go
@@ -91,6 +91,14 @@ func (s *Server) Start(ctx context.Context) error {
 	s.started = true
 
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				l.Error("health check server goroutine panicked",
+					zap.Any("panic", r),
+					zap.Stack("stack"),
+				)
+			}
+		}()
 		l.Info("health check server starting", zap.String("address", addr))
 		if err := s.server.Serve(listener); err != nil && err != http.ErrServerClosed {
 			l.Error("health check server error", zap.Error(err))
--- a/pkg/healthcheck/server.go
+++ b/pkg/healthcheck/server.go
@@ -91,14 +91,11 @@ func (s *Server) Start(ctx context.Context) error {
 	s.started = true
 
 	go func() {
-		defer func() {
-			if r := recover(); r != nil {
-				l.Error("health check server goroutine panicked",
-					zap.Any("panic", r),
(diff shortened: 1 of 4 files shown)


pkg/healthcheck/server.go
```diff
@@ -91,6 +91,14 @@ func (s *Server) Start(ctx context.Context) error {
 	s.started = true
 
 	go func() {
+		defer func() {
```
kans: Crashing is probably more correct behavior.  Recovery means we have a dead health server that thinks its alive.

ggreer (approved):

PR #845, opened 2026-05-20
fix: recover panics in background goroutines

robert-chiniquy:
## what changed

Four goroutines in the SDK launch background work whose body can panic, but none of them install a `recover`. A panic in a goroutine without a deferred recover takes the entire process down -- the connector daemon dies and any in-flight RPC the client is awaiting hangs forever.

The four sites:

- **`pkg/actions::invokeGlobalAction`** launches the user-supplied global action handler. A panic in a handler now sets the action's error state instead of crashing.
- **`pkg/actions::invokeResourceAction`** launches the user-supplied resource-scoped action handler. Same fix.
- **`pkg/cli::closeLambdaConnectorGeneration`** launches the connector's `Close()` (via reflection on `lambdaConnectorCloserWithContext` / `lambdaConnectorCloser`). A panic in `Close` now flows back to the caller through the same `errCh` the success path uses.
- **`internal/connector::runServer`** launches the gRPC session store server in a goroutine. A panic in the gRPC server (handler, codec, TLS, interceptor) now logs and lets the deferred listener close fire, instead of crashing the connector wrapper.

## why this matters

Action handlers are connector-supplied user code -- the SDK has no way to know what they will do. Same for the connector `Close` lifecycle method. Treating these as trusted-not-to-panic is fragile and has bitten us in similar shapes in connector PRs (see baton-google-workspace `fetchUserTokens` goroutine recover).

## how I found this

Static analysis with [occult-go-analysis](https://github.com/conductorone/occult-go-analysis), c1 profile, `goroutine_without_recover` detector.

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -272,6 +272,13 @@ func (cw *wrapper) runServer(ctx context.Context, serverCred *tlsV1.Credential) 
 		cw.SessionServer = server
 		go func() {
 			defer sessionListener.Close()
+			defer func() {
+				if r := recover(); r != nil {
+					l.Error("panic in session store server goroutine",
+						zap.Any("panic", r),
+						zap.Stack("stack"))
+				}
+			}()
 			serverErr := session.StartGRPCSessionServerWithOptions(ctx, sessionListener, server,
 				grpc.Creds(credentials.NewTLS(tlsConfig)),
 				grpc.ChainUnaryInterceptor(ugrpc.UnaryServerInterceptor(ctx)...),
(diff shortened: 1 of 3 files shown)


madison-c-evans (approved):
good catch well scoped lgtm

internal/connector/connector.go
```diff
@@ -272,6 +272,13 @@ func (cw *wrapper) runServer(ctx context.Context, serverCred *tlsV1.Credential)
 		cw.SessionServer = server
 		go func() {
 			defer sessionListener.Close()
+			defer func() {
```
kans: We need this panic, right?  If we can't start the server, there is no way to communicate that outside the goroutine.

kans (changes requested):

internal/connector/connector.go
robert-chiniquy: Agreed — reverted that recover in 5d78fa2e. The panic was the only failure signal to the parent and recovering hid it. The other four recover sites stay (pkg/actions x2, lambda closers x2) — those goroutines have no upstream startup-failure semantics to preserve.

kans (approved):

PR #848, opened 2026-05-21
CXH-1473: Add source_type fields to GroupTrait and AppTrait

al-conductorone:
Connectors can now report what kind of group or app each entry is (security group, distribution list, dynamic, etc.) so C1 stops dropping that classification when it pulls data from identity providers.

### DIFF
--- a/pkg/types/resource/app_trait.go
+++ b/pkg/types/resource/app_trait.go
@@ -53,6 +53,20 @@ func WithAppHelpURL(helpURL string) AppTraitOption {
 	}
 }
 
+func WithAppSourceType(sourceType string) AppTraitOption {
+	return func(at *v2.AppTrait) error {
+		at.SetAppSourceType(sourceType)
+		return nil
+	}
+}
+
+func WithRawAppSourceType(raw string) AppTraitOption {
+	return func(at *v2.AppTrait) error {
+		at.SetRawAppSourceType(raw)
+		return nil
+	}
+}
+
 // NewAppTrait creates a new `AppTrait` with the given help URL, and profile.
 func NewAppTrait(opts ...AppTraitOption) (*v2.AppTrait, error) {
 	at := &v2.AppTrait{}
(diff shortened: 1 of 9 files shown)


pkg/types/resource/app_trait.go
```diff
@@ -53,6 +53,20 @@ func WithAppHelpURL(helpURL string) AppTraitOption {
 	}
 }
 
+func WithAppSourceType(sourceType string) AppTraitOption {
```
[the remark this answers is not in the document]
al-conductorone: no vocab exists yet for app source types so no constants to create

pkg/types/resource/app_trait.go
```diff
@@ -53,6 +53,20 @@ func WithAppHelpURL(helpURL string) AppTraitOption {
...
+	return func(at *v2.AppTrait) error {
+		at.SetAppSourceType(sourceType)
+		return nil
+	}
+}
+
+func WithRawAppSourceType(raw string) AppTraitOption {
```
FeliLucero1, line 65: both WithAppSourceType and WithRawAppSourceType are plain string, callers can swap them with no compiler error

al-conductorone: constrained WithAppSourceType

luisina-santos (approved):

pkg/types/resource/group_trait.go
```diff
@@ -8,6 +8,17 @@ import (
...
 
+type GroupSourceType string
+
+const (
+	GroupSourceTypeNative          GroupSourceType = "native"
+	GroupSourceTypeAppImported     GroupSourceType = "app_imported"
+	GroupSourceTypeBuiltIn         GroupSourceType = "built_in"
+	GroupSourceTypeDirectorySynced GroupSourceType = "directory_synced"
+	GroupSourceTypeDynamic         GroupSourceType = "dynamic"
+	GroupSourceTypeDistribution    GroupSourceType = "distribution"
```
ggreer, line 19: It might be nice to add validation rules in the protos to limit the values to these.

ggreer (approved):
Only noticed one thing, but it's not a blocker.

PR #851, opened 2026-05-21
sync: pre-build entitlement-id set in processGrantsWithExternalPrincipals

btipling:
## Summary

C1 uplift perf campaign — second baton-sdk PR (sibling to #849).

`processGrantsWithExternalPrincipals` issues a `s.store.GetEntitlement` per slug per principal-match while iterating every external-source grant via `ListWithAnnotations`. The pattern repeats in both the `MatchID` and `MatchExternal` branches. At tenant scales the campaign targets (millions of grants per sync, large entitlement counts per resource), this is material wall-clock during c1's post-sync uplift.

This pre-builds the entitlement-ID set once per call via one paginated `ListEntitlements`. Inner-loop existence check becomes O(1) set membership. No behavior change.

## What changes

- `pkg/sync/syncer.go`: New `listAllEntitlementIDs` helper paginates `ListEntitlements` into a `mapset.Set[string]` (existing project idiom — see L1901, L2486). `processGrantsWithExternalPrincipals` builds the set once at function entry and replaces both per-slug `GetEntitlement` existence checks with `knownEntitlementIDs.ContainsOne(newExpandableEntId)`. The existing "found no entitlement with entitlement id generated from external source sync" error log is preserved; the surrounding `sql.ErrNoRows` plumbing is no longer needed.

## Perf rationale

| Scenario | Before | After |
|---|---|---|
| Inner existence check | 1 SQL `GetEntitlement` per (grant × matched slug) | O(1) set lookup |
| Total DB calls | O(G × matched × slugs) `GetEntitlement` | 1 paginated `ListEntitlements` |
| Set size | n/a | 10^2–10^4 entitlement IDs (fits comfortably) |

For a tenant with 10^5 external-source grants and an average match/slug fanout of 5, this collapses ~500k per-row `GetEntitlement` queries into a single paginated list.

No benchmark currently covers `processGrantsWithExternalPrincipals`; before/after numbers are not included.

## Backwards compatibility

Function signature unchanged. No new params, no new return values, no interface changes. All existing callers continue to compile and behave identically.

The single new helper (`listAllEntitlementIDs`) is private to the package.

## Self-review

- Diff matches the candidate spec — `+40/-14` lines, one file, no interface changes.
- Set is constructed once at function entry, outside both nested grant-iteration loops.
- Set type (`mapset.Set[string]` from `github.com/deckarep/golang-set/v2`) matches the existing idiom in the same file.
- Behavior unchanged: every grant that was processed before is still processed, in the same order, with the same side effects. Missing-entitlement log line preserved.
- Existing `TestExternalResource*` coverage (`TestExternalResourceMatchID`, `TestExternalResourceMatchIDWithExpandableRemapping`, `TestExternalResourceEmailMatch`, `TestExternalResourceGroupProfileMatch`, `TestExternalResourceWithGrantToEntitlement`) all pass.
- `gofmt` clean.
- Comments explain WHY (perf rationale) where non-obvious; do not restate the code.

Self-review clean.

## Test plan

- [x] `go build ./...`
- [x] `go test -count=1 ./pkg/sync/ -v` — all pass, including `TestExternalResource*`
- [x] `go test ./...` — all pass
- [ ] Manual: benchmark with realistic resource sizing (no existing bench covers this path)

## Related

- Sibling PR in the same campaign: #849 (compareProto -> proto.Equal)
- Campaign tracker: stargate `investigations/c1-uplift-perf-tracker.md`

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -2150,6 +2150,31 @@ func (s *syncer) listExternalResourcesForResourceType(ctx context.Context, resou
 	return resources, nil
 }
 
+// listAllEntitlementIDs paginates every entitlement in the local store and
+// returns just the ID set. processGrantsWithExternalPrincipals needs only
+// existence, not the full entitlement blob, so this avoids the per-slug
+// GetEntitlement N+1 inside its nested match-by-id / match-by-key loops.
+func (s *syncer) listAllEntitlementIDs(ctx context.Context) (mapset.Set[string], error) {
+	ids := mapset.NewSet[string]()
+	pageToken := ""
+	for {
+		resp, err := s.store.ListEntitlements(ctx, v2.EntitlementsServiceListEntitlementsRequest_builder{
+			PageToken: pageToken,
+		}.Build())
+		if err != nil {
+			return nil, err
+		}
+		for _, ent := range resp.GetList() {
+			ids.Add(ent.GetId())
+		}
+		pageToken = resp.GetNextPageToken()
+		if pageToken == "" {
+			break
+		}
+	}
+	return ids, nil
+}
+
 func (s *syncer) listExternalEntitlementsForResource(ctx context.Context, resource *v2.Resource) ([]*v2.Entitlement, error) {
 	ents := make([]*v2.Entitlement, 0)
 	entitlementToken := ""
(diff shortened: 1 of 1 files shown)


pkg/sync/syncer.go
```diff
@@ -2238,6 +2263,15 @@ func (s *syncer) processGrantsWithExternalPrincipals(ctx context.Context, princi
 
 	l := ctxzap.Extract(ctx)
 
+	// One ListEntitlements paginate replaces the per-slug GetEntitlement
+	// existence checks below. Entitlement counts are 10^2-10^4 on real
+	// tenants, so the in-memory set fits comfortably and the grant-iteration
+	// loop's inner GetEntitlement fanout collapses to set membership.
+	knownEntitlementIDs, err := s.listAllEntitlementIDs(ctx)
```
kans, line 2270: 1. Do you know the biggest entitlement count, currently?  Some connectors have dynamic ents.
2. We do this before we know that we need it.  We could make it lazy (use a once or something).  IF the answer to 1 is actually 10^4, I don't care anymore.

kans (approved):
Address comment before merge.

robert-chiniquy:
LGTM but I'm scared of not knowing enough about uplift and lifecycle

PR #853, opened 2026-05-21
validate exclusion group is only tied to one resource type/has at most one default entitlement

laurenleach:
- validate exclusion group id is only tied to one resource type
- validate exclusion group only has one default entitlement

### DIFF
--- a/pkg/sync/state.go
+++ b/pkg/sync/state.go
@@ -52,6 +52,14 @@ type State interface {
 	// updated (if needed) and ("", false) is returned. The existing return
 	// value is meaningful only when conflict is true.
 	CheckAndSetExclusionGroupDefault(exclusionGroupID, entitlementID string) (existing string, conflict bool)
+	// IncrementExclusionGroupCount increments the running count of entitlements
+	// observed for exclusionGroupID and returns the new count. The count is
+	// global per exclusion group id (across all resources) and survives resume.
+	IncrementExclusionGroupCount(exclusionGroupID string) uint32
+	// ClearExclusionGroupTracking drops the exclusion group bookkeeping maps so
+	// they don't bloat the token of a completed sync, mirroring
+	// ClearEntitlementGraph.
+	ClearExclusionGroupTracking(ctx context.Context)
 }
 
 // ActionOp represents a sync operation.
@@ -179,6 +187,7 @@ type state struct {
 	completedActionsCount           uint64
 	exclusionGroupResourceTypes     map[string]string
 	exclusionGroupDefaults          map[string]string
+	exclusionGroupCounts            map[string]uint32
 }
 
 // Original serialized token format. Needed to parse/resume syncs started by older versions of baton-sdk.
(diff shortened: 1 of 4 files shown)


pkg/sync/state.go
```diff
@@ -484,3 +516,37 @@ func (st *state) GetCompletedActionsCount() uint64 {
...
+	defer st.mtx.Unlock()
+
+	if st.exclusionGroupResourceTypes == nil {
+		st.exclusionGroupResourceTypes = make(map[string]string)
+	}
+	if existing, ok := st.exclusionGroupResourceTypes[exclusionGroupID]; ok {
+		if existing != resourceTypeID {
+			return existing, true
+		}
+		return existing, false
```
MarcusGoldschmidt: Should we return `""` ?

pkg/sync/state.go
```diff
@@ -484,3 +516,37 @@ func (st *state) GetCompletedActionsCount() uint64 {
...
+	defer st.mtx.Unlock()
+
+	if st.exclusionGroupDefaults == nil {
+		st.exclusionGroupDefaults = make(map[string]string)
+	}
+	if existing, ok := st.exclusionGroupDefaults[exclusionGroupID]; ok {
+		if existing != entitlementID {
+			return existing, true
+		}
+		return existing, false
```
MarcusGoldschmidt: Should we return "" ?

MarcusGoldschmidt (approved):

pkg/sync/state.go
```diff
@@ -38,6 +38,13 @@ type State interface {
 	ShouldSkipGrants() bool
 	SetShouldSkipGrants()
 	GetCompletedActionsCount() uint64
+	// CheckAndSetExclusionGroupResourceType records that exclusionGroupID is
+	// used on resourceTypeID. If the group was already recorded against a
+	// different resource type, the prior value is returned with seen=true and
+	// the map is not modified. If the group was unseen or already maps to
+	// resourceTypeID, the map is updated and (resourceTypeID, false) is
+	// returned.
+	CheckAndSetExclusionGroupResourceType(exclusionGroupID, resourceTypeID string) (existing string, conflict bool)
```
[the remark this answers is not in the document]
ggreer: Fortunately, nothing else implements this. It's just meant to allow for different ways to serialize/deserialize state.

pkg/sync/state.go
```diff
@@ -191,16 +207,20 @@ type serializedTokenV1 struct {
...
-		currentActionID:  0,
-		entitlementGraph: nil,
-		needsExpansion:   false,
+		actions:                     make(map[string]Action),
+		actionOrder:                 []string{},
+		currentActionID:             0,
+		entitlementGraph:            nil,
+		needsExpansion:              false,
+		exclusionGroupResourceTypes: make(map[string]string),
+		exclusionGroupDefaults:      make(map[string]string),
```
ggreer, line 233: We should clear these out at the end of a sync so that we don't have a big value in the token column of a completed sync, similar to how we clear the entitlement graph with ClearEntitlementGraph().

ggreer (approved):
It would be nice to clear these after the sync is complete, but otherwise the implementation seems fine.

pkg/sync/state.go
laurenleach: addressed in https://github.com/ConductorOne/baton-sdk/pull/853/changes/e8225c45c85d03cb972ba0b46c72ee972efe3bfc

ggreer (approved):

PR #857, opened 2026-05-22
Preserve lambda Viper StringMap keys

gontzess:
Why

The typed-config fix preserves StringMap key case for generated connector configs, but lambda connectors that intentionally use *viper.Viper still receive the effective lambda config. Viper.Set recursively lowercases nested map[string]any keys, so dynamic StringMap fields can still lose case-sensitive keys.

What this changes

This makes the lambda effective config schema-aware for StringMapField values. Those payloads are stored as JSON strings before entering Viper, which keeps GetStringMap and GetStringMapString behavior while avoiding Viper's recursive map-key lowercasing. The PR also adds a regression test for the Viper path and keeps the existing typed-config test intact.

Depends on #856.

Validation

- go test -tags baton_lambda_support ./pkg/cli
- go test ./pkg/cli

### DIFF
--- a/pkg/cli/lambda_server__added.go
+++ b/pkg/cli/lambda_server__added.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/debug"
+	"strings"
 	"sync"
 	"time"
 
@@ -334,7 +335,7 @@ func OptionallyAddLambdaCommand[T field.Configurable](
 			}
 
 			connectorConfig := configStruct.AsMap()
-			effectiveConfig := effectiveLambdaConfig(v, connectorConfig)
+			effectiveConfig := effectiveLambdaConfig(v, connectorConfig, connectorSchema)
 			logLevelConfig, err := lambdaLogLevelConfigFromViper(effectiveConfig)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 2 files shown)


pkg/cli/lambda_server__added.go
```diff
@@ -477,14 +478,48 @@ func cloneViperSettings(v *viper.Viper) *viper.Viper {
...
 		effectiveConfig.Set(key, value)
 	}
 	return effectiveConfig
 }
 
+func lambdaStringMapFields(connectorSchema field.Configuration) map[string]struct{} {
+	out := make(map[string]struct{})
+	for _, schemaField := range connectorSchema.Fields {
+		if schemaField.Variant == field.StringMapVariant {
+			out[strings.ToLower(schemaField.FieldName)] = struct{}{}
```
kans, line 509: Why do we want to lower the field names here at all?  WHat is consuming `out` that needs lowered names?

gontzess: Example config value:
```
map[string]any{
	"user-custom-fields": map[string]any{
		"customString10":     "Custom String 10",
		"departmentNav/name": "Department",
	},
}
```

Non-Lambda CLI

Top-level field names are effectively Viper-managed and case-insensitive. The nested map keys are the customer’s payload and should stay as-is.
```
v.GetStringMapString("user-custom-fields")
// map[string]string{
//   "customString10":     "Custom String 10",
//   "departmentNav/name": "Department",
// }
```

Lambda, Typed Connector

This is the SuccessFactors-style path. With PR #856, the typed generated config decodes the raw lambda payload directly, bypassing Viper.Set for the connector struct.
```
cfg.UserCustomFields
// map[string]any{
//   "customString10":     "Custom String 10",
//   "departmentNav/name": "Department",
// }
```

PR #857 does not materially affect this path.

Lambda, Axiomatic / `*viper.Viper`

This is the future footgun path. Axiomatic receives *viper.Viper, so without PR #857 the nested keys would pass through Viper.Set as a map and become:
```
map[string]any{
	"customstring10":     "Custom String 10",
	"departmentnav/name": "Department",
}
```

With PR #857, only schema-declared StringMapField values are converted before Viper.Set:

```
"user-custom-fields" => `{"customString10":"Custom String 10","departmentNav/name":"Department"}`
```

Then Axiomatic reads it normally:
```
v.GetStringMapString("user-custom-fields")
// map[string]string{
//   "customString10":     "Custom String 10",
//   "departmentNav/name": "Department",
// }
```

we haven't used this field type yet for axiomatic but i think we want to set the precident as wanting case sensative - map keys are often external identifiers, not user-facing option names. They may be JSON paths, OData paths, SCIM attributes, API field names, headers, claim names, or vendor-specific IDs. Many of those are case-sensitive, or at least case-preserving.

kans (approved):

PR #863, opened 2026-05-24
perf(expand): prefetch descendant grants to eliminate per-principal SQL queries

arreyder:
## Summary

- Replace O(N) per-principal `ListGrantsForEntitlement` SQL queries in `runAction` with a single bulk pre-fetch into an in-memory map
- For a large tenant (180K users, 139K groups, 72 GB c1z), a single expansion action was taking **38 minutes** — 180K individual SQLite queries through pure-Go modernc.org/sqlite
- After: one paginated scan of the descendant entitlement + O(1) map lookups per source grant → expected **seconds** instead of minutes

## Problem

Profile evidence from be-temporal-sync prod (1h aggregate):
- 100% of CPU in `dotc1z.listGrantsGeneric → ListGrantsForEntitlement → database/sql.Rows.Next → modernc.org/sqlite._sqlite3VdbeExec`
- Block profile: 0 blocked samples — CPU-bound, not I/O-bound
- 277 pending grant-expansion actions × ~38 min each = days to complete one sync

The inner loop in `runAction` (previously lines 278-350) called `ListGrantsForEntitlement` with a specific `PrincipalId` for each source grant. Even though the compound index `(entitlement_id, principal_resource_type_id, principal_resource_id)` makes each a point lookup returning 0-1 rows, the per-query overhead in pure-Go SQLite (~0.1-1ms per call) compounds to 38 minutes at 180K source grants.

## Fix

Pre-fetch ALL grants on the descendant entitlement into `map[principalKey][]*v2.Grant` before entering the source grants loop. Each source grant then does a map lookup (O(1)) instead of a SQL query.

The pre-fetch is per-`runAction` call, not cached across actions. This ensures correctness in diamond graphs: if action A→C creates grants that action B→C must see, the B→C call does a fresh pre-fetch that includes the writes from A→C.

## Memory tradeoff

For the worst case (descendant with 180K grants × ~2KB each = ~360MB peak), this is within budget for worker pods (32GB RAM). Most entitlements are much smaller.

## Test plan

- [x] `go test ./pkg/sync/expand/...` — all pass (including diamond graph, multi-level, mixed directness, cycle tests)
- [x] `go test ./pkg/sync/...` — all pass
- [x] `go test -short ./pkg/dotc1z/...` — all pass
- [x] `go test -bench=BenchmarkExpand ./pkg/sync/expand/...` — 70% improvement

## Benchmark

```
                              main (baseline)      branch (prefetch)     delta
BenchmarkExpandSmall            89.0s                26.4s              -70.3%
BenchmarkExpandSmallMedium     113.5s                34.4s              -69.7%
allocs/op (Small)              299M                 129M                -57.0%
B/op (Small)                  14.8 GB               7.2 GB             -51.4%
```

### DIFF
--- a/pkg/sync/expand/expander.go
+++ b/pkg/sync/expand/expander.go
@@ -48,9 +48,6 @@ type ExpanderStore interface {
 type Expander struct {
 	store ExpanderStore
 	graph *EntitlementGraph
-
-	prefetchedDescendantID string
-	prefetchedDescendants  map[string][]*v2.Grant
 }
 
 // NewExpander creates a new Expander with the given store and graph.
@@ -110,8 +107,6 @@ func (e *Expander) RunSingleStep(ctx context.Context) error {
 			// Action is complete - mark edge expanded and remove from queue
 			e.graph.MarkEdgeExpanded(action.SourceEntitlementID, action.DescendantEntitlementID)
 			e.graph.Actions = e.graph.Actions[1:]
-			e.prefetchedDescendantID = ""
-			e.prefetchedDescendants = nil
 		}
 	}
 
@@ -159,22 +154,18 @@ func descendantGrantKey(resourceTypeID, resourceID string) string {
 	return resourceTypeID + "\x00" + resourceID
 }
 
-func (e *Expander) getDescendantGrants(
+func prefetchDescendantGrants(
 	ctx context.Context,
+	store ExpanderStore,
 	entitlement *v2.Entitlement,
 ) (map[string][]*v2.Grant, error) {
-	entID := entitlement.GetId()
-	if e.prefetchedDescendantID == entID && e.prefetchedDescendants != nil {
-		return e.prefetchedDescendants, nil
-	}
-
 	result := make(map[string][]*v2.Grant)
 	pageToken := ""
 	for page := 0; page < maxPrefetchPages; page++ {
 		if err := ctx.Err(); err != nil {
 			return nil, err
 		}
-		resp, err := e.store.ListGrantsForEntitlement(ctx,
+		resp, err := store.ListGrantsForEntitlement(ctx,
 			reader_v2.GrantsReaderServiceListGrantsForEntitlementRequest_builder{
 				Entitlement: entitlement,
 				PageToken:   pageToken,
@@ -192,9 +183,6 @@ func (e *Expander) getDescendantGrants(
 			break
 		}
 	}
-
-	e.prefetchedDescendantID = entID
-	e.prefetchedDescendants = result
 	return result, nil
 }
 
@@ -258,7 +246,7 @@ func (e *Expander) runAction(ctx context.Context, action *EntitlementGraphAction
 		return "", fmt.Errorf("runAction: error fetching source grants: %w", err)
 	}
 
-	descendantByPrincipal, err := e.getDescendantGrants(ctx, descendantEntitlement.GetEntitlement())
+	descendantByPrincipal, err := prefetchDescendantGrants(ctx, e.store, descendantEntitlement.GetEntitlement())
 	if err != nil {
 		l.Error("runAction: error prefetching descendant grants", zap.Error(err))
 		return "", fmt.Errorf("runAction: error prefetching descendant grants: %w", err)
--- a/pkg/sync/expand/expander.go
+++ b/pkg/sync/expand/expander.go
@@ -324,6 +324,14 @@ func (e *Expander) runAction(ctx context.Context, action *EntitlementGraphAction
 	var newGrants = make([]*v2.Grant, 0)
 	for _, sourceGrant := range sourceGrants.GetList() {
 		if sourceGrant.GetPrincipal() == nil {
+			// Malformed grant with no principal. main passed this through to
+			// ListGrantsForEntitlement with a nil PrincipalId, which the store
+			// treats as "no filter" (pkg/dotc1z/grants.go) — fetching every
+			// descendant grant and adding this source to all of them,
+			// corrupting unrelated principals. Skip it, but log so the upstream
+			// connector bug is visible instead of silently dropped.
+			l.Warn("runAction: skipping source grant with nil principal",
+				zap.String("source_grant_id", sourceGrant.GetId()))
 			continue
 		}
 		if action.Shallow {
--- a/pkg/sync/expand/expander.go
+++ b/pkg/sync/expand/expander.go
@@ -191,6 +191,14 @@ func prefetchDescendantPrincipals(
 			return nil, false, fmt.Errorf("prefetchDescendantPrincipals: %w", err)
 		}
 		for _, g := range resp.GetList() {
+			// Skip malformed grants with no principal: they would otherwise
+			// add an empty key to the set, polluting the negative oracle the
+			// caller relies on. Consistent with the nil-principal guard in
+			// runAction. (Opaque getters are nil-safe, so this is correctness,
+			// not panic-avoidance.)
+			if g.GetPrincipal() == nil {
+				continue
+			}
 			pid := g.GetPrincipal().GetId()
 			set[descendantGrantKey(pid.GetResourceType(), pid.GetResource())] = struct{}{}
 		}
--- a/pkg/sync/expand/expander.go
+++ b/pkg/sync/expand/expander.go
@@ -148,6 +148,39 @@ func (e *Expander) IsDone(ctx context.Context) bool {
 	return e.graph.IsExpanded()
 }
 
+func descendantGrantKey(resourceTypeID, resourceID string) string {
+	return resourceTypeID + "\x00" + resourceID
+}
+
+func prefetchDescendantGrants(
+	ctx context.Context,
+	store ExpanderStore,
+	entitlement *v2.Entitlement,
+) (map[string][]*v2.Grant, error) {
+	result := make(map[string][]*v2.Grant)
+	pageToken := ""
+	for {
+		resp, err := store.ListGrantsForEntitlement(ctx,
+			reader_v2.GrantsReaderServiceListGrantsForEntitlementRequest_builder{
+				Entitlement: entitlement,
+				PageToken:   pageToken,
+			}.Build())
+		if err != nil {
+			return nil, fmt.Errorf("prefetchDescendantGrants: %w", err)
+		}
+		for _, g := range resp.GetList() {
+			principal := g.GetPrincipal().GetId()
+			key := descendantGrantKey(principal.GetResourceType(), principal.GetResource())
+			result[key] = append(result[key], g)
+		}
+		pageToken = resp.GetNextPageToken()
+		if pageToken == "" {
+			break
+		}
+	}
+	return result, nil
+}
+
 // runAction processes a single action and returns the next page token.
 // If the returned page token is empty, the action is complete.
 //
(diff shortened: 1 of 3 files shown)


arreyder:
## Benchmark results

Ran on the existing `BenchmarkExpand{Small,SmallMedium}` fixtures (232-node graph, 787 edges, real c1z file). These use full-blob mode (slim writer OFF) — same configuration as Lilly's prod connector.

```
goos: linux
goarch: arm64
cpu: Neoverse-V2

                              main (baseline)      branch (prefetch)     delta
─────────────────────────────────────────────────────────────────────────────────
BenchmarkExpandSmall            89.0s                26.4s              -70.3%
BenchmarkExpandSmallMedium     113.5s                34.4s              -69.7%

                              main                 branch               delta
─────────────────────────────────────────────────────────────────────────────────
allocs/op (Small)              299M                 129M                -57.0%
allocs/op (SmallMedium)        384M                 170M                -55.8%
B/op (Small)                  14.8 GB               7.2 GB             -51.4%
B/op (SmallMedium)            19.0 GB               9.5 GB             -50.0%
```

**70% wall-clock improvement, 57% fewer allocations, 51% less memory allocated.**

The benchmark fixture is a 232-node graph — much smaller than Lilly's 139K-group tenant. The improvement scales with graph density: the more grants per entitlement, the more per-principal SQL queries the old code issued. Lilly's 180K-member "all employees" group (the action that takes 38 minutes today) should see an even larger relative speedup because:

1. The fixed per-query overhead (statement prepare, cursor open/close) dominates at high N
2. The prefetch cache (`getDescendantGrants`) amortizes across all 18 pages of source grants within a single action

alan-lee-12:
re. the benchmark above:

my 🤖 pointed out that it used one Expander for the whole run, while production syncer uses a new Expander per RunSingleStep in syncer.go (line 2555), so the cache does not persist across pages.

```
Since each source-grant page is one step, production will prefetch the descendant
entitlement once per source page, not once per action.

The changes in this PR may still be much faster than 180,000 separate point queries.
But it means the claimed performance model is not quite true in production, and the
benchmark is likely too optimistic because it uses expander.Run(ctx), which keeps the
same Expander alive across pages.

I’d make the cache lifetime explicit and owned by the syncer, not by a throwaway Expander.
```

pkg/sync/expand/expander.go
```diff
@@ -148,6 +153,51 @@ func (e *Expander) IsDone(ctx context.Context) bool {
...
+	entitlement *v2.Entitlement,
+) (map[string][]*v2.Grant, error) {
+	entID := entitlement.GetId()
+	if e.prefetchedDescendantID == entID && e.prefetchedDescendants != nil {
+		return e.prefetchedDescendants, nil
+	}
+
+	result := make(map[string][]*v2.Grant)
+	pageToken := ""
+	for page := 0; page < maxPrefetchPages; page++ {
```
kans, line 181: What happens if we hit maxPrefetchPages?

arreyder: Good catch — traced the caller and silent truncation is actually unsafe. In `runAction` (expander.go:277-285), a missing key in the prefetched map is treated identically to "this principal has no existing descendant grant" → we call `newExpandedGrant` and append a fresh grant instead of merging sources into the existing one. So a partial map produces duplicate grants with wrong source attribution, not just an incomplete result.

Fixed in cb8a2b63: `prefetchDescendantGrants` now returns `ErrPrefetchPageCapExceeded` (with the entitlement ID and page count wrapped in) when it hits the cap, so the failure surfaces to `runAction` and bubbles up as a normal expand error rather than silently corrupting output. Added `TestPrefetchDescendantGrants_PageCapExceeded` driving a stub store that always returns a non-empty `NextPageToken` to lock the behavior in.

The 10000-page cap is preserved as a defense-in-depth bound; hitting it in practice still means we have a real data-scale issue worth surfacing rather than papering over.

kans: We need to change the algo a bit. 

1. We should be able to handle very large c1zs in principal, even if it takes way too long without exploding.
2. We should lower the bounds on the page cap to bound memory to a smaller value.
3. As a general rule, progress should be captured, somehow or at least resumable, because syncs get interrupted.  This also argues in favor of a smaller cap.

arreyder: Addressed in 2b6f25f2 by switching the prefetch to a principal-key set + per-hit per-principal query:

- **Memory bound**: the prefetched map is now `map[string]struct{}` (only the principal-key string), not `map[string][]*v2.Grant`. Per-entry footprint drops from ~2KB (full grant payload) to ~80B (resource-type + resource-id). For a 180K-principal descendant that's ~14MB instead of ~360MB.
- **Lower page cap**: `maxPrefetchPages` is now 100 (down from 10000). With the default 10000-rows/page, that bounds the scan at 1M principals → ~80MB peak set size.
- **Graceful degradation on cap exceeded**: instead of returning `ErrPrefetchPageCapExceeded`, `prefetchDescendantPrincipals` now returns `(set, complete=false, nil)`. `runAction` logs a warning, drops the partial set, and falls back to per-principal queries for the rest of the page. Very large c1zs run slowly but don't OOM or fail.
- **Source loop**: if the key-set is complete and a principal is *not* in it, we skip the SQL query and create a new grant directly (the fast path that drove the Lilly speedup — most principals in a fresh expansion have no existing descendant grant). If the principal *is* in the set, we issue one `ListGrantsForEntitlement(ent, PrincipalId=p)` to fetch and merge.

Resumability: the action's source-grant `PageToken` already captures progress across `RunSingleStep` calls; the per-action prefetch is rebuilt each page, which is the bounded-memory tradeoff. No per-runtime state to lose on interruption.

pkg/sync/expand/expander.go
```diff
@@ -208,79 +241,52 @@ func (e *Expander) runAction(ctx context.Context, action *EntitlementGraphAction
...
-				if err != nil {
-					l.Error("runAction: error updating descendant grants", zap.Error(err))
-					return "", fmt.Errorf("runAction: error updating descendant grants: %w", err)
-				}
-				break
-			}
-
-			// Add the source entitlement as a source to all descendant grants.
-			grantsToUpdate := make([]*v2.Grant, 0)
+			newGrants = append(newGrants, descendantGrant)
```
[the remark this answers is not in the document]
arreyder: Fixed in 2b6f25f2. The source loop now keeps a per-page `descendantsByKey` cache that is seeded with newly created grants — so a second source grant for the same principal in the same page takes the merge path against the in-memory grant instead of falling through to `newExpandedGrant` again.

`TestExpanderInPageDuplicatePrincipal` locks the behavior in: two source grants on the same entitlement for the same principal must produce exactly one expanded grant on the descendant.

pkg/sync/expand/expander.go
```diff
@@ -148,6 +153,51 @@ func (e *Expander) IsDone(ctx context.Context) bool {
...
+	return resourceTypeID + "\x00" + resourceID
+}
+
+func (e *Expander) getDescendantGrants(
+	ctx context.Context,
+	entitlement *v2.Entitlement,
+) (map[string][]*v2.Grant, error) {
+	entID := entitlement.GetId()
+	if e.prefetchedDescendantID == entID && e.prefetchedDescendants != nil {
+		return e.prefetchedDescendants, nil
```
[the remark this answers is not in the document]
arreyder: The Expander-level cache that this comment was concerned about was already removed in a90304c7. The prefetch now happens inside `runAction` — one call per source-grants page, no carry-over state. A grant written by page N is persisted by `PutGrantsInChunks` before page N+1 runs, and page N+1's fresh prefetch picks it up via the key-set scan.

The in-page same-principal case (cache invisible to *later iterations within the same page*) is the dual of this concern; 2b6f25f2 fixes that by seeding the local cache with newly created grants.

arreyder:
## Re-bench with prod-shaped harness

Per @alan-lee-12's note that `syncer.expandGrantsForEntitlements` builds a new `Expander` per `RunSingleStep`, I added a `*PerStep` benchmark variant that mirrors that shape exactly — fresh `NewExpander(store, graph)` before each `RunSingleStep`, loop until `IsDone`. Same fixture (`36zGvJw3uxU1QMJKU2yPVQ1hBOC`, 232 nodes / 787 edges), `-count=3 -benchtime=1x`, i7-1255U.

| Branch | Harness | Wall (avg) | Bytes/op | Allocs/op |
|---|---|---|---|---|
| main | one-Expander-per-run | 100.4 s | 14.80 GB | 299,271,478 |
| main | new-Expander-per-step | 118.2 s | 14.80 GB | 299,271,562 |
| this PR | one-Expander-per-run | **80.6 s** | **12.83 GB** | **250,710,693** |
| this PR | new-Expander-per-step | **82.7 s** | **12.85 GB** | **251,529,176** |

**Win vs main, both harness shapes:**
- Allocations: **−16.2%** (~48M fewer per run)
- Bytes: **−13.3%**
- Wall: −19.7% (one-Expander) / −30.0% (per-step). Wall-clock variance on this machine is wide (each iter rebuilds the c1z tmp copy), but bytes/allocs are deterministic — that's the load-bearing signal.

**Why harness shape doesn't change the win:** `prefetchDescendantPrincipals`'s key-set and `descendantsByKey` cache are both `make`'d inside `runAction` itself — neither lives on the `Expander` struct (that was 5a7ddeeb's whole point). The prefetch wins per action, not per expander lifetime, so it's invariant under "1 Expander × N RunSingleStep" vs "N Expanders × 1 RunSingleStep". Bench confirms it: alloc count is identical (within ±0.3M) across the two shapes within each branch.

Pushed the `*PerStep` variant in 2efc545c so you can re-run it. Happy to add `BenchmarkExpandSmallMediumPerStep` measurements too if you want the larger fixture re-confirmed.

pkg/sync/expand/expander.go
```diff
@@ -289,21 +394,15 @@ func (e *Expander) runAction(ctx context.Context, action *EntitlementGraphAction
 				if updated {
 					sources := v2.GrantSources_builder{Sources: sourcesMap}.Build()
 					descendantGrant.SetSources(sources)
-					grantsToUpdate = append(grantsToUpdate, descendantGrant)
+					newGrants = append(newGrants, descendantGrant)
```
kans, line 422: I think we are dropping some is_directs here that would have been updated in the unbuffered case in main.  As in, we see an indirect then a direct, we don't update, it.  A failing (or passing test) would be good.... 

edit: although, I'm not sure if the behavior in main is sane.  I'd vote for correctness, I think, over exactly reproducing main.

pkg/sync/expand/expander.go
```diff
@@ -208,79 +295,97 @@ func (e *Expander) runAction(ctx context.Context, action *EntitlementGraphAction
...
+	// same page merges into the in-memory grant instead of producing a
+	// duplicate (same deterministic grant ID would otherwise upsert and lose
+	// the first iteration's source attribution).
+	descendantsByKey := make(map[string][]*v2.Grant)
+
 	var newGrants = make([]*v2.Grant, 0)
 	for _, sourceGrant := range sourceGrants.GetList() {
-		// If this is a shallow action, then we only want to expand grants that have no sources
-		// which indicates that it was directly assigned.
+		if sourceGrant.GetPrincipal() == nil {
```
kans, line 334: This is another strange case.  Main doesn't do this (obviously).  Instead, it filters by the nil principal, which is interpreted as no filter!... which maybe results in us updating all descendant grants!?... or erroring out.  This only effects malformed data, but I don't think the SDK actually rejects grants with no principal (we should do that too).  So, it _may_ have been adding a self source (direct)... 

@ggreer ^

pkg/sync/expand/expander.go
arreyder: Good catch — confirmed it's a real regression and fixed in 2493c894.

**The bug:** in the in-page cache merge path, the second iteration for a repeated principal only wrote a source when it was *absent* (`sourcesMap[SourceEntitlementID] == nil`), so an indirect-then-direct sequence kept `IsDirect=false`. main got the upgrade "for free": it re-queried the store every iteration, the first write was still buffered (un-flushed), so the second iteration re-created the grant via `newExpandedGrant` and last-write-wins promoted it to direct.

**The fix:** I went with correctness over reproducing main, per your edit. The merge path now explicitly upgrades indirect→direct:

```go
if existingSource := sourcesMap[action.SourceEntitlementID]; existingSource == nil {
    sourcesMap[action.SourceEntitlementID] = &v2.GrantSources_GrantSource{IsDirect: isSourceDirect}
    updated = true
} else if isSourceDirect && !existingSource.GetIsDirect() {
    existingSource.SetIsDirect(true)
    updated = true
}
```

This is also strictly *more* correct than main, which was order-dependent: main's last-write-wins only produced the right answer when the direct grant happened to come second. Direct-first-then-indirect would have left it `false`. The new code is order-independent — direct always wins.

`TestExpanderInPageDirectnessUpgrade` locks it in (indirect grant first, direct second, asserts the descendant source is direct). It fails before the change, passes after.

pkg/sync/expand/expander.go
arreyder: Confirmed the analysis. In production `ListGrantsForEntitlement` only filters by principal when `PrincipalId != nil` (`pkg/dotc1z/grants.go:199-204`), so main's nil principal → nil `PrincipalId` → **no filter** → it fetched *every* grant on the descendant entitlement and ran the source-merge loop over all of them. The `len(sourcesMap)==0` branch would also stamp a direct self-source on each — so yes, a single malformed grant could rewrite the sources of every principal on the descendant. Definitely a (latent) corruption bug, only reachable with malformed data.

The rewrite is safe here two ways:
1. **Explicit skip** (the `continue` you flagged).
2. **Structurally** — the new path never issues an unfiltered descendant query. It builds a principal-key *set* and then does point lookups via `listDescendantGrantsForPrincipal(ent, principal)`. A nil principal just yields an empty key that matches nothing, so even without the guard it can't fan out across all descendant grants the way main did. (I verified: with the guard removed, a nil-principal source grant is a plain no-op — no writes, no error.)

Two changes pushed in ec40d778:
- **`l.Warn` on skip** — silently dropping malformed grants hides connector bugs, so we now log the offending `source_grant_id`.
- **`TestExpanderNilPrincipalSourceGrant`** — locks in the property: a nil-principal source grant must not add the source entitlement to an unrelated principal's pre-existing descendant grant. (It's a forward-looking regression guard rather than a fails-on-main test, since the new architecture already can't reproduce the fan-out.)

On the SDK-side point — agreed it'd be better to reject grants with no principal at write time rather than rely on every consumer to defend against them. That's a broader contract change outside this perf PR; happy to file a follow-up issue for it. @ggreer — thoughts on rejecting nil-principal grants at ingest?

pkg/sync/expand/expander.go
```diff
@@ -148,6 +148,93 @@ func (e *Expander) IsDone(ctx context.Context) bool {
...
+			reader_v2.GrantsReaderServiceListGrantsForEntitlementRequest_builder{
+				Entitlement: entitlement,
+				PageToken:   pageToken,
+			}.Build())
+		if err != nil {
+			return nil, false, fmt.Errorf("prefetchDescendantPrincipals: %w", err)
+		}
+		for _, g := range resp.GetList() {
+			pid := g.GetPrincipal().GetId()
+			set[descendantGrantKey(pid.GetResourceType(), pid.GetResource())] = struct{}{}
```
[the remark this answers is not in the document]
arreyder: Done in 0d06400c. Added the `continue` guard.

One correction: this wouldn't actually panic — the opaque proto getters are nil-safe (`GetPrincipal().GetId().GetResourceType()` returns `""` on a nil chain rather than dereferencing). The real bug is subtler: a nil-principal descendant grant would insert an empty `"\x00"` key into the principal-key set, which then acts as a bogus entry in the negative oracle `runAction` uses to decide whether to skip the per-principal query. Guarding keeps junk out of the set and stays consistent with the nil-principal skip in `runAction`. Verified existing tests still pass.

kans (approved):

PR #864, opened 2026-05-24
perf(dotc1z): add prepared statement cache for repeated SQL queries

arreyder:
## Summary

- Add a `map[string]*sql.Stmt` cache on `*C1File` that eliminates repeated SQLite statement compilation for read queries
- Profile evidence: `modernc.org/sqlite.(*conn).bind` consumes **163s/hr (5.6% CPU)** recompiling the same ~10 SQL shapes thousands of times per sync
- With `MaxOpenConns(1)`, all statements are bound to a single connection for the lifetime of the sync activity — ideal for caching

## How it works

goqu already uses `.Prepared(true)` on all read paths, generating stable `?`-placeholder SQL. The SQL text is deterministic for each query shape (determined by which WHERE clauses are active). `getOrPrepare(ctx, query)` checks the cache; on miss, prepares the statement and stores it. On hit, returns the cached `*sql.Stmt` immediately.

## Call sites converted (7)

| File | Function | Hot path |
|---|---|---|
| `sql_helpers.go` | `listConnectorObjects` | Main sync list loop |
| `sql_helpers.go` | `getConnectorObject` | GetEntitlement, GetResource |
| `sql_helpers.go` | `getResourceObject` | GetResource variant |
| `grants.go` | `listGrantsGeneric` | ListGrantsForEntitlement |
| `grants_expandable_query.go` | `listExpandableGrantsInternal` | Expansion work queue |
| `grants_expandable_query.go` | `listGrantsWithExpansionInternal` | Expansion data read |
| `grants_hydrate.go` | `hydrateSingleGrant` | Slim-grant hydration |

## What's NOT changed

- Write paths (`executeChunkedInsert`) — already use transactions, INSERT shapes vary with batch size
- The goqu query builder still runs (SQL text still built each call to use as cache key) — eliminating goqu is a future follow-up

## Benchmark

```
                              main (baseline)      stmt-cache           delta
BenchmarkExpandSmall            89.0s                73.3s              -17.7%
BenchmarkExpandSmallMedium     113.9s                94.4s              -17.1%
```

17-18% wall-clock improvement. Allocations flat (goqu still builds the SQL string — savings are pure CPU from eliminating bytecode recompilation).

## Test plan

- [x] `go test -short ./pkg/dotc1z/...` — all pass
- [x] `go test ./pkg/sync/...` — all pass
- [x] `go vet ./pkg/dotc1z/...` — clean

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -67,6 +67,11 @@ type C1File struct {
 	slowQueryThreshold    time.Duration
 	slowQueryLogFrequency time.Duration
 
+	// Prepared statement cache: keyed by SQL text, lazily populated.
+	// With MaxOpenConns(1) all stmts are bound to the single connection.
+	stmtCache   map[string]*sql.Stmt
+	stmtCacheMu sync.Mutex
+
 	// Sync cleanup settings
 	syncLimit   int
 	skipCleanup bool
@@ -216,6 +221,7 @@ func NewC1File(ctx context.Context, dbFilePath string, opts ...C1FOption) (*C1Fi
 		slowQueryThreshold:    5 * time.Second,
 		slowQueryLogFrequency: 1 * time.Minute,
 		encoderConcurrency:    1,
+		stmtCache:             make(map[string]*sql.Stmt),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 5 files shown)


arreyder:
## Benchmark results

Same machine, same fixtures as PR #863 benchmarks. This branch has ONLY the stmt cache (no prefetch from #863) — measuring the cache in isolation.

```
goos: linux
goarch: amd64
cpu: AMD Ryzen 7 5700X3D 8-Core Processor

                              main (baseline)      stmt-cache           delta
─────────────────────────────────────────────────────────────────────────────────
BenchmarkExpandSmall            89.0s                73.3s              -17.7%
BenchmarkExpandSmallMedium     113.9s                94.4s              -17.1%

                              main                 stmt-cache           delta
─────────────────────────────────────────────────────────────────────────────────
allocs/op (Small)              299M                 300M                ~flat
allocs/op (SmallMedium)        384M                 385M                ~flat
B/op (Small)                  14.8 GB              14.8 GB              ~flat
B/op (SmallMedium)            19.0 GB              19.0 GB              ~flat
```

**17-18% wall-clock improvement, allocations flat.** 

The savings are pure CPU: eliminating repeated SQLite bytecode compilation for the same ~10-20 query shapes that execute thousands of times per sync. Allocations stay flat because goqu still builds the SQL string each call (used as the cache key) — cutting that is Phase 3 (raw SQL templates), a future PR.

This stacks with PR #863 (prefetch): #863 reduces the NUMBER of queries by 10,000x; #864 makes each remaining query faster by ~18%. Together they compound.

pkg/dotc1z/c1file.go
```diff
@@ -492,12 +498,48 @@ func (c *C1File) closeRawDB(ctx context.Context) error {
...
+		return nil, fmt.Errorf("getOrPrepare: %w", err)
+	}
+
+	c.stmtCacheMu.Lock()
+	if existing, ok := c.stmtCache[query]; ok {
+		c.stmtCacheMu.Unlock()
+		stmt.Close()
+		return existing, nil
+	}
+	c.stmtCache[query] = stmt
```
alan-lee-12, line 619: [🤖] P1: `getOrPrepare` can still panic on the cache-miss path if `closeRawDB` runs between preparing the new statement and re-acquiring `stmtCacheMu`. `closeRawDB` sets `c.stmtCache = nil`, so this write can become a write to a nil map. Please mirror the entry nil check after the second lock, close the newly prepared statement, and return `ErrDbNotOpen` before checking/storing the cache entry.

arreyder: Addressed in 426e8854 — added the nil re-check inside the second `stmtCacheMu.Lock()` section (c1file.go:533-537). On cache miss, after re-acquiring the lock we check whether `closeRawDB` nilled `stmtCache` while we were preparing, and if so we close the freshly prepared stmt and return `ErrDbNotOpen` before any write to the (now-nil) map. Mirrors the existing pre-prepare nil guard.

PR #875, opened 2026-05-25
feat: c1z sanitizer v0.1 — library + CLI

btipling:
## Summary
- New `pkg/c1zsanitize` package + `cmd/baton-c1z-sanitize` CLI that transform a `.c1z` into an identity-stripped copy via `connectorstore.Reader`/`Writer`. Per-c1z HMAC-SHA256 secret drives every transform.
- Cross-references stay coherent because the transform is deterministic within a c1z — every place an id appears (`Grant.principal`, `Grant.entitlement`, `Grant.sources` map keys, `Entitlement.resource`, `Resource.parent_resource_id`) maps through the same `sanitize_id`.
- Annotation dispatch is a whitelist of Any type URLs. v0.1 handles `UserTrait`, `GroupTrait`, `AppTrait`, `RoleTrait`, `SecretTrait`, `LicenseProfileTrait`, `ScopeBindingTrait`. Unknown annotations are dropped by default with a log line; `-allow-unknown-annotations` flips to pass-through.
- Timestamps use anchor-and-shift (newest source timestamp → anchor, single Δ applied uniformly) so relative deltas survive. Asset payloads get a content-type-matched placeholder while the ref chain is preserved.

Implementation follows the design in §6.2 of the investigation document. The sanitizer code never imports `c1.storage.v3`; it works entirely through `connectorstore.Reader/Writer` and the connector-v2 wire types as the investigation prescribed.

## Output format choice
v2 (sqlite-zstd). The investigation's §7 question 5 punted on v2 vs v3 with the proviso "v0.1 should write v3 by default if PRs #870/#871/#872 have landed; otherwise v2." At the time of this PR, the storage-engine-v4 stack (#867–#872) is all still **open** on `main`, so v0.1 writes v2 and v0.2 swaps to v3 once the writer adapter ships.

## Open questions / choices for ambiguous items
- **Resource type IDs preserved.** §2.3.1 of the investigation flagged that some connectors might construct `resource_type_id` with tenant data. v0.1 preserves them; the §7 question 1 audit hasn't run yet.
- **Asset placeholder content.** Image content types get a 1x1 PNG; everything else gets a single zero byte. `PutAsset` silently drops empty data, so the single byte is the minimum that keeps the cross-reference alive. Document as known-lossy.
- **Sync ID minting.** The destination Writer's `StartNewSync` mints a fresh KSUID rather than accepting a deterministic transform of the source sync id. Parent linkage is preserved via an in-memory `srcSyncID → dstSyncID` map maintained for the call. The v2 `connectorstore.Writer` interface doesn't expose `SetSyncID`, so the deterministic-KSUID approach from the investigation §6.4 is deferred.
- **All sync runs copied.** No `-max-sync-runs` flag yet (investigation §7 question 6). Add when needed.
- **Reversibility.** Operator-supplied secret via `-secret-file`, or the CLI generates one and writes it next to `-out` with mode 0600. Archive or shred — the sanitizer doesn't choose.
- **Role scope conditions.** §3.7's expression strings are HMACed wholesale. Lossy by design (see §7 question 3). Connector-side structured alternatives are the v0.2 path.

## Out of scope for v0.1 (per §6.3)
- S3 fetch / upload, Union Station / Temporal integration, content-store registration, opt-in tracking, encryption at rest, reversibility tooling, selective re-sanitization.

## Test plan
- [x] `go vet ./...`, `gofmt -l`, `golangci-lint run` clean on new code
- [x] `go test ./pkg/c1zsanitize/` passes — all unit + invariant tests green
- [x] `go test ./...` passes — no existing tests broken
- [x] Hand-crafted fixture exercises every annotation handler in v0.1's whitelist (UserTrait/GroupTrait/AppTrait/RoleTrait/SecretTrait/LicenseProfileTrait/ScopeBindingTrait)
- [x] Cross-reference integrity invariant: for every source id that appears N times, `sanitize_id(id)` appears exactly N times in dst
- [x] Graph integrity invariants: `Grant.principal`/`Grant.entitlement`/`Entitlement.resource`/`Resource.parent_resource_id` all resolve in dst
- [x] Unknown annotation type dropped by default with `DropUnknownAnnotations=true`
- [x] CLI end-to-end smoke test: build a small fixture, run `baton-c1z-sanitize -in src -out dst`, assert exit 0 and dst exists


cmd/baton-c1z-sanitize/main.go
```diff
@@ -0,0 +1,150 @@
...
+
+func main() {
+	if err := run(); err != nil {
+		fmt.Fprintln(os.Stderr, "baton-c1z-sanitize:", err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	fs := flag.NewFlagSet("baton-c1z-sanitize", flag.ContinueOnError)
```
kans: This could go in the baton utility.

pkg/c1zsanitize/annotations.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil
+	}
+	out := make([]*anypb.Any, 0, len(in))
+	for _, a := range in {
+		if a == nil {
+			continue
+		}
+		handler, ok := s.handlers[a.GetTypeUrl()]
+		if !ok {
+			if s.dropUnknownAnnotations {
```
kans, line 59: This should probably be true by default?

pkg/c1zsanitize/sanitize.go
```diff
@@ -0,0 +1,216 @@
...
+// id is the per-sanitizer hot path. SanitizeID stays as the
+// allocation-y reference implementation; this one reuses a single
+// hmac.Hash so the SHA-256 key schedule isn't redone every call.
+// Sanitize runs single-threaded, so no locking is needed.
+func (s *sanitizer) id(input string) string {
+	if input == "" {
+		return ""
+	}
+	s.idHmac.Reset()
+	s.idHmac.Write([]byte(input))
```
kans: This transformation isn't application aware.  For a grant, we can split the parts and hmac the substrings then reconstruct the string so that the references align.

kans:
1. Make `Sanitize` handle source syncs with other types - `partial_upserts` / `partial_deletions` produced by `GenerateSyncDiffFromFile`. Currently `sanitizeSync` passes those through to SQLite `StartNewSync`, which rejects them. Preserve the source sync graph semantics where possible, including parent/linked sync relationships if supported, or choose a deliberate compatible representation and test it with a diff-sync fixture.

2. Move CLI validation before secret generation. `baton sanitize` currently writes the generated secret before checking that `--file` exists and `--out` does not. Validate input/output paths first so failed invocations do not leave a stray `.secret` file. Add/adjust command tests.

3. We don't currently sanitize the members of the composite IDs.  We could split them and sanitize the components to not leak any PII ... I'm not sure this is necessary.

kans (approved):

PR #887, opened 2026-05-27
Add grant cancellation error

alan-lee-12:
## Summary
- add `grant.ErrGrantCancelled` and `grant.NewErrGrantCancelled(reason)` for connectors that intentionally decline a grant

Motivation:
- C1 can now detect rejected grants with `errors.As(err, *grant.ErrGrantCancelled)` and handle them as terminal cancellations instead of connector failures.


laurenleach (approved):

proto/c1/connector/v2/annotation_grant.proto
```diff
@@ -29,6 +29,11 @@ message GrantImmutable {
 // Grant was not created because the entitlement already existed.
 message GrantAlreadyExists {}
 
+// Grant was not created because connector policy rejected it.
+message GrantRejected {
+  string reason = 1;
+}
```
ggreer: Why not have the connector's Grant() method return a GRPC status code of FailedPrecondition or InvalidArgument, along with an error message?

alan-lee-12: the customer's ask is that when a grant is rejected due to certain constraint, the ticket should be canceled. InvalidArgument/FailedPrecondition are treated as non-retryable failures and produce CONNECTOR_ACTION_STATE_FATAL_ERROR, GRANT_OUTCOME_ERROR, and app-owner reassignment/error handling. these don't make sense to them since the ticket didn't really "fail"/"error".

If we reuse existing, i'm worried that we could overload them, or necessitate special-casing of the error, which probably requires adding other status details.

alan-lee-12: discussed off-line, and i'm going to change the approach to let sdk export a named error instead of this new anno.

ggreer (approved):

PR #888, opened 2026-05-27
[wip] failing pebble tests

kans:
working through pebble bugs 

image


pkg/dotc1z/engine/pebble/adapter.go
```diff
@@ -802,22 +820,44 @@ func (a *Adapter) currentSyncID() string {
...
+//     pkg/dotc1z/sql_helpers.go `resolveSyncID`.
+//  3. The adapter's current sync (set by StartNewSync / SetCurrentSync).
+//  4. The most-recent finished sync of any type. Lets Reader calls
+//     against a closed c1z resolve to its stored data (caught by the
+//     cross-engine parity test).
+//
+// Returns ("", nil) when no sync resolves. A malformed SyncDetails
+// annotation surfaces as a non-nil error so callers don't silently
+// fall through to the wrong sync.
+func (a *Adapter) resolveActiveSync(reqSyncID string, annos []*anypb.Any) (string, error) {
```
ggreer: Nitpick: Can we call it resolveSyncID to match SQLite? To make things a little cleaner, maybe it can take an interface that implements GetAnnotations() so we can avoid `req.GetAnnotations()` in every call.

Also there are some request structs that have a GetActiveSyncId() (such as ResourceTypesServiceListResourceTypesRequest) or GetSyncId(). By passing the whole request in, in the future we could be clever about checking if those methods exist and obey them before returning the current active sync.

pkg/dotc1z/engine/pebble/adapter.go
```diff
@@ -802,22 +820,44 @@ func (a *Adapter) currentSyncID() string {
...
+		return "", fmt.Errorf("pebble: read sync_id from annotations: %w", err)
+	}
+	if annoSyncID != "" {
+		return annoSyncID, nil
 	}
 	if id := a.currentSyncID(); id != "" {
-		return id
+		return id, nil
 	}
 	if id, err := a.LatestFinishedSyncID(context.Background(), connectorstore.SyncTypeAny); err == nil && id != "" {
```
ggreer: Maybe this method can take a context to be used here.

pkg/dotc1z/engine/pebble/adapter_reader.go
```diff
@@ -494,20 +522,36 @@ func v3SyncTypeToString(t v3.SyncType) string {
...
+// even though the data is right there on disk; the cross-engine parity
+// test caught that divergence.
 //
-// Without (2) a Reader call after EndSync would return ErrNoCurrentSync
-// even though the data is right there on disk. SQLite already does this
-// fallback; the cross-engine parity test caught the divergence.
-func (a *Adapter) resolveActiveSyncForReader(_ []*anypb.Any) string {
+// A malformed SyncDetails annotation surfaces as a non-nil error so
+// callers don't silently fall through to the wrong sync.
+func (a *Adapter) resolveActiveSyncForReader(annos []*anypb.Any) (string, error) {
```
ggreer: Instead of repeating the code in resolveActiveSync, this method should just call resolveActiveSync() and return empty string if no sync is found. The same goes for resolveSyncIDForRead() in bulk_by_ids.go.

pkg/dotc1z/engine/pebble/index_migrations.go
```diff
@@ -273,6 +278,68 @@ func backfillGrantPrincipalResourceTypeIndex(ctx context.Context, e *Engine) err
...
+// entries (including the new idxGrantByEntitlementResource) for every
+// sync. writeGrantIndexes is idempotent — re-emitting the older
+// indexes (entitlement / principal / principal_resource_type /
+// needs_expansion) is a cheap no-op Set.
+//
+// Required for any c1z written before idxGrantByEntitlementResource
+// existed. Without this backfill, Adapter.ListGrants(Resource=...)
+// and pebbleGrantStore.ListWithAnnotationsForResourcePage would
+// return empty results on existing files until they were re-synced.
+func backfillGrantEntitlementResourceIndex(ctx context.Context, e *Engine) error {
```
ggreer: Ideally we could avoid writing migrations for pebble until we are actually using it. I'm sure the data model will be tweaked a few times before we settle on what's best for production usage.

kans: Yeah, the migrations are versioned, but we can try the tests without hte migrations.

ggreer (approved):

PR #892, opened 2026-05-28
Verify uploaded c1z size against local read so truncated objects do not go live

arreyder:
## Summary

Adds `us3.PutWithVerify` (and the existing `Put` delegates to it) so every S3 upload is followed by:

1. **Counting reader** — bytes the SDK read from us matches what the caller declared as expected size (catches short-reads at the source, including the io.Pipe-style "encryption goroutine returned clean EOF on cancel" case once the corresponding c1 fix lands).
2. **HeadObject content-length assertion** — what S3 stored matches what we sent, catching the truncated-multipart-completed-as-success failure mode regardless of source.

On any mismatch the just-uploaded object is `DeleteObject`'d on a detached short-deadline context so a worker drain cannot keep the corrupt object live in S3.

## Changes

- `pkg/us3/s3.go` — adds `PutWithVerify(ctx, key, r, expectedSize, contentType)`. `Put` delegates with `expectedSize=-1`. `PutWithVerify` validates non-empty key, enforces `maxFileSize` for known sizes (closes a long-standing gap where the existing `WithMaxDownloadFilesize` docstring said "uploaded or downloaded" but uploads never checked), and wraps the upload body in a `countingReader`. Verification logic extracted to `verifyUpload()` so it is unit-testable without an S3 round-trip.
- `pkg/dotc1z/manager/s3/s3.go` — `SaveC1Z` stats the tmp file and forwards its size into `PutWithVerify` so the source-short-read assertion fires on the caller side too. Adopts a small `s3PutClient` interface so the manager is mockable.
- `pkg/us3/s3_test.go` (new) — 9 unit tests covering `countingReader` (happy + n>0+err path) and `verifyUpload` for all branches: happy, expectedSize=-1, expectedSize=0 happy, expectedSize=0 over-read, source short-read, source over-read, HeadObject error, HeadObject size mismatch (with and without expectedSize).
- `pkg/us3/s3_integration_test.go` (new) — end-to-end tests using a thin `httptest` fake S3 (PutObject + HeadObject + DeleteObject + GetBucketLocation, enough for the SDK's single-PutObject path below its 8 MiB part-size threshold). Covers happy path, head-size mismatch triggers delete, source short-read triggers delete without head.
- `pkg/dotc1z/manager/s3/s3_test.go` (new) — six tests that capture `PutWithVerify`'s args via a fake `s3PutClient` and confirm the local file size is forwarded correctly, including the empty-file edge case.

## Notes for reviewers

- `Put`'s contract is now slightly stricter: any caller of `Put` will perform a HeadObject after upload. New IAM dependency: callers need `s3:HeadObject` + `s3:DeleteObject` in addition to `s3:PutObject`. This should be a no-op for typical baton consumers (the manager's S3 client always had these) but worth a release-note callout.
- `DeleteObject` was briefly added as a public method during development and has been kept private (`deleteObject`) — the only caller is `deleteAfterTruncation` in the same package.

## Test plan

- [x] `go test ./...` clean
- [x] `make lint` clean on the touched files
- [x] End-to-end tests assert PutObject + HeadObject + DeleteObject call counts
- [x] Short-read assertion fires before HeadObject is reached


pkg/dotc1z/manager/s3/s3.go
```diff
@@ -19,8 +19,17 @@ import (
...
+// NewS3Manager (it satisfies the interface); tests inject a fake that
+// records the call without standing up an S3 fake.
+type s3PutClient interface {
+	PutWithVerify(ctx context.Context, key string, r io.Reader, expectedSize int64, contentType string) error
+	Get(ctx context.Context, key string) (io.Reader, error)
+}
+
 type s3Manager struct {
-	client         *us3.S3Client
+	client         s3PutClient
```
ggreer: I don't think we're using the s3 manager anywhere in c1. Matt and I were planning on removing it.

arreyder:
Good catch. Pulled the `manager/s3` changes out in 6c7912bd — c1 never constructs the S3 manager (none of the `manager.New(...)` call sites pass `WithS3()`), so the prevention there isn't load-bearing for the corruption fix in c1, and I'd rather not give your deletion a merge conflict. The `us3.PutWithVerify` primitive stays on its own; any future caller can opt in without going through `manager/s3`. Test file `manager/s3/s3_test.go` deleted too since it goes with the package.

ggreer:
I don't think we're using any s3 code anywhere. If I vendor https://github.com/ConductorOne/baton-sdk/pull/894 into c1, everything builds and runs fine.

arreyder:
Closing: this PR adds `PutWithVerify` to `pkg/us3`, but main removed the entire `us3` / `s3manager` subsystem as dead code in #894 ("Remove s3manager and s3 util. They were never used by anything."). There's no longer a code path to rebase onto. The c1z-truncation protection this was part of lives in the ctx-detach work (#891) and the c1 upload path.

PR #899, opened 2026-05-30
synccompactor: pass runCtx into doOneCompaction so c.runDuration actually bounds the loop

arreyder:
## Summary

`Compact()` builds `runCtx = context.WithTimeout(ctx, c.runDuration)` but the loop calls `doOneCompaction(ctx, ...)`. The deadline only gates the pre-flight select; individual compactions then run under the unbounded parent ctx and can blow past `c.runDuration`.

This is the sibling of the bug fixed in #893 (`WriteEnvelope` walked under an unbounded context). Same shape: a deadline-bearing context exists but inner work uses a different one, so the deadline dangles instead of bounding the work.

## Change

- `pkg/synccompactor/compactor.go`: pass `runCtx` into `doOneCompaction` so `c.runDuration` actually limits the loop.
- When `runCtx` fires *due to its own deadline* (not parent cancellation), surface the same clean **\"compaction run duration has expired\"** message the pre-flight select already uses, instead of letting a bare `context.DeadlineExceeded` bubble out of inner SQLite operations. The error is still returned so callers can decide whether to retry.

## Impact

Caller-visible behavior change for callers that pass `WithRunDuration`:

- **Before:** `c.runDuration` only checked at entry; individual compactions could run as long as the parent ctx allowed. If they did eventually fail (e.g. heartbeat timeout during a long SQLite operation), the error was a raw `context deadline exceeded` from inside `ExecContext`.
- **After:** each `doOneCompaction` is bounded by `c.runDuration`. When the deadline fires, the loop exits with `compaction run duration has expired: context deadline exceeded`, and the structured log identifies which sync was in flight and how many remain.

Callers that don't use `WithRunDuration` are unaffected — `runCtx == ctx` in that branch.

## Test plan

- [x] `go build ./pkg/synccompactor/...`
- [x] `go vet ./pkg/synccompactor/...`
- [x] `go test ./pkg/synccompactor/...` (all existing tests pass)
- [ ] CI

## Context

Surfaced by a code-review bot on a separate compaction stuck-state investigation. Lilly's Entra connector compaction was failing at ~98 min with `context deadline exceeded` from inside `CompactGrants` even though `WithRunDuration(1h)` was passed — because `runCtx` was created but not threaded into the loop, the deadline never fired cleanly; instead the activity hit its Temporal heartbeat timeout while SQLite was blocking.

This PR alone won't unblock Lilly (the compaction still doesn't fit in 1h), but it makes the failure intentional and legible, and matches the contract the option name implies.

### DIFF
--- a/pkg/synccompactor/compactor.go
+++ b/pkg/synccompactor/compactor.go
@@ -216,9 +216,22 @@ func (c *Compactor) Compact(ctx context.Context) (*CompactableSync, error) {
 	l.Debug("new empty partial sync created", zap.String("sync_id", newSyncId))
 
 	// Base sync is c.entries[0], so compact in reverse order. That way we compact the biggest sync last.
+	// Pass runCtx (not the outer ctx) so c.runDuration actually bounds the loop. Without this, the
+	// deadline set on runCtx only gates the pre-flight check above and individual doOneCompaction
+	// calls run under the unbounded parent ctx.
 	for i := len(c.entries) - 1; i >= 0; i-- {
-		err = c.doOneCompaction(ctx, c.entries[i])
+		err = c.doOneCompaction(runCtx, c.entries[i])
 		if err != nil {
+			// When runCtx fires due to c.runDuration, surface the same clean message the pre-flight
+			// check uses instead of bubbling a bare context.DeadlineExceeded out of the inner sqlite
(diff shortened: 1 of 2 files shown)


btipling (approved):
semantically, it should be correct for doOneCompaction to use the c.runDuration context.

mj-palanker (approved):

PR #901, opened 2026-05-30
Remove c1z manager.

ggreer:
After s3manger was removed, there was only one implementation (localmanager). Instead of having another layer of indirection, just open the c1zs directly.

This also lets us avoid having to add more manager options that mirror the c1zoptions (such as which DB engine to use)


cmd/baton/optimize.go
```diff
@@ -52,11 +46,6 @@ func runOptimizeDb(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	err = m.SaveC1Z(ctx)
```
kans, line 55: This previously copied the file to a different location?

ggreer: The previous version copied the c1z file, decompressed it, optimized it, compressed it, and stomped over the original file. The new code just decompresses the c1z, optimizes it, and compresses it over the original.

I can change this to copy it manually and move the compressed c1z to the original location so that there's no chance of corrupting the original. In practice, I think this command is only used for us testing/benchmarking the vacuum/optimize code.

kans (approved):

PR #906, opened 2026-06-02
feat: baton rollback-expansion subcommand

btipling:
## Validation — rollback / re-expand round-trip

Validated end-to-end on a real Microsoft Entra c1z: `rollback-expansion` deleted exactly the 41 expansion-derived grants while leaving the 824 direct grants untouched, and `--replay` re-derived the identical 41 to restore the exact 865-grant set with identical `baton stats`.

| Metric | BEFORE | ROLLBACK | RE-EXPAND |
|---|---|---|---|
| total live grants | 865 | 824 | 865 |
| derived | 41 | 0 | 41 |
| direct | 824 | 824 | 824 |

Rollback's own log line reported `deleted 41`.

**Assertion 1 — rollback yields direct-only: PASS.** 865 → 824 (41 derived removed), 0 derived grants remain, the 824 direct grants are byte-identical to the baseline's direct set, and none of the 41 derived grants survived.

**Assertion 2 — re-expand matches the baseline: PASS.** 865 == 865 set-equal, derived 41 == 41, direct 824 == 824, and the `baton stats` diff is clean against the baseline.

### How to reproduce

```
go build -o /tmp/baton-rollback ./cmd/baton
# on a c1z whose latest sync actually ran grant expansion (its expansion grants carry Grant.Sources):
/tmp/baton-rollback -f sync.c1z rollback-expansion --dry-run                          # reports N expansion-derived grants to delete
/tmp/baton-rollback -f sync.c1z rollback-expansion --out rolled-back.c1z              # deletes them; direct grants untouched
/tmp/baton-rollback -f rolled-back.c1z rollback-expansion --replay --out re-expanded.c1z   # re-derives the expansion grants
# verify live counts before/after:
/tmp/baton-rollback -f <c1z> grants -o json | jq '[.[] | select(.sources != null and (.sources|length>0))] | length'   # derived
```

Caveat: a c1z whose sync never ran grant expansion (no `supports_diff` marker / no expansion grants) will report 0 to delete — use a sync that expanded. Also note that `baton stats` reads the stored sync-run stats blob, so it may still show the pre-rollback total; the authoritative live count is `grants -o json` (and the command's own `deleted N` log line).

---

## What

Adds `baton rollback-expansion` — a subcommand that reverts grant expansion in a c1z back to its directly-synced grants so the expansion can be re-run. Two uses:

- **Regression testing** — roll back, replay expansion through the current code, and compare against the original to confirm a change to grant expansion did not alter the output.
- **Performance validation** — roll back, then time the replay, to measure whether a grant-expansion performance improvement actually helps on real production data.

Every change is confined to the subcommand and its store helper — there is no sync-write change, no schema change, and no new column.

## Mechanism

In plain terms: during a sync the expander turns indirect access into explicit grants — e.g. "group A is granted into group B" becomes a derived grant for each member of A — and tags each derived grant with the source grant(s) it came from. Rollback deletes exactly those expander-created grants (keeping the directly-synced ones, and clearing the source-tags off any direct grant the expander also touched) and resets the sync token, all in one transaction. With the token reset, `--replay` re-runs the real production expansion to regenerate them. Net: rollback strips expansion back to the raw synced grants; replay rebuilds it identically.

The detail:

Provenance comes from `Grant.Sources`, which the expander already populates on existing c1z files. For every grant in the target sync the rollback inspects the `GrantSources` map:

- **No Sources** → a plain direct grant the expander never touched. Left untouched.
- **Sources present, none keyed to the grant's own entitlement id** → a purely expander-derived grant. Deleted.
- **Sources present including a self-source for the grant's own entitlement** → a direct grant the expander touched on first use. Kept, with its Sources cleared to empty so it matches the pre-expansion shape.

The grant's own entitlement id is read from the `entitlement_id` **column**, not the data blob — `slimGrantForWrite` nils `Entitlement` on write, so the blob would yield no id and mis-classify every slim grant. Sources are not slimmed, so they are read from the blob.

The deletes, the source clears, and a `sync_token=''` reset run in **one transaction**, so the rollback is all-or-nothing. The empty token makes a later replay seed a fresh init step that re-pushes the grant-expansion step; without it a finished sync's completed token would replay to a no-op.

`RollbackResult` reports `GrantsDeleted`, `SourcesCleared`, and `SuspectConnectorSourced` — the last counts deleted grants that carried Sources but no self-source **and** lacked the `GrantImmutable` annotation every expander-created grant carries. That shape looks connector-set rather than expander-set, so the count surfaces any real connector-sourced grant at risk of being swept up. Zero in the common case.

Pagination reads each 1000-row page fully and closes its cursor before any mutation runs, so reads and writes never share an open cursor on the same SQLite connection.

## Replay

`--replay` re-runs expansion through the public syncer path exactly as the compactor does — empty connector + the existing store + `WithOnlyExpandGrants` — so it runs the full production expansion: graph load, cycle fixing honoring `BATON_DONT_FIX_CYCLES`, and the expander. A replay matches a fresh sync rather than a hand-rolled subset.

## Safety

- **`--out` mandatory for any write** — the input c1z is never modified. A write without `--out` is refused; `--dry-run` previews counts read-only.
- **Writes go to a fresh clone** via `CloneSync`, which copies only the targeted sync and refuses an existing `--out`. A partial `--out` is removed if any step after the clone fails, so an identical retry is not blocked.
- **Refuses an unfinished sync** (`ErrSyncNotFinished`): its partially-expanded graph lives in the sync token and would be stranded, causing irreversible grant loss on a later resume.
- **Refuses a sync that never expanded** (`ErrSyncNotExpanded`, no `supports_diff` marker): nothing to roll back.
- Default `--sync-id` is the latest finished sync.

## Tests

`go test ./pkg/dotc1z/ ./cmd/baton/` green. New tests: dry-run counts, unfinished-sync refusal, `--out`-required refusal, and an end-to-end command run asserting the input is untouched, the expander-derived grant is gone in `--out`, and the direct grant survives.

## Verification

- `go build ./...` ✓
- `go test ./pkg/dotc1z/ ./cmd/baton/` ✓
- `gofmt -l` clean ✓

## Open items

- A replay-correctness integration test over a multi-level + cyclic fixture. The replay routes through the verified sync-path engine, so the current test is a runs-clean smoke.
- The branch is behind `main`; the GitHub diff shows three pre-merged files (`c1file.go`, `sync_runs.go`, `sync_runs_test.go`) and the `v0.12.1` bump that are already byte-identical in `main`. Rebasing reduces the diff to the four subcommand files; the net merge footprint is already only those four.

### DIFF
--- a/pkg/dotc1z/rollback_expansion.go
+++ b/pkg/dotc1z/rollback_expansion.go
@@ -87,11 +87,28 @@ func (c *C1File) RollbackExpansion(ctx context.Context, syncID string, dryRun bo
 		return nil, err
 	}
 	deleteQuery := fmt.Sprintf("DELETE FROM %s WHERE derived_by_expansion = 1 AND sync_id = ?", tableName)
-	if _, err := tx.ExecContext(ctx, deleteQuery, syncID); err != nil {
-		if rbErr := tx.Rollback(); rbErr != nil {
-			return nil, errors.Join(rbErr, err)
+	// Clearing the sync token leaves the sync ready to be re-expanded.
+	// A completed sync's token persists a finished state with an empty
+	// action queue; on replay the syncer would resume that token, find
+	// nothing to do, and exit without expanding. An empty token instead
+	// makes the syncer seed a fresh InitOp, which pushes the grant-
+	// expansion step. Done in the same transaction as the delete so the
+	// rollback is all-or-nothing.
+	resetTokenQuery := fmt.Sprintf("UPDATE %s SET sync_token = '' WHERE sync_id = ?", syncRuns.Name())
+	txErr := func() error {
+		if _, err := tx.ExecContext(ctx, deleteQuery, syncID); err != nil {
+			return err
 		}
-		return nil, fmt.Errorf("c1z: rollback delete failed: %w", err)
+		if _, err := tx.ExecContext(ctx, resetTokenQuery, syncID); err != nil {
+			return err
+		}
+		return nil
+	}()
+	if txErr != nil {
+		if rbErr := tx.Rollback(); rbErr != nil {
+			return nil, errors.Join(rbErr, txErr)
+		}
+		return nil, fmt.Errorf("c1z: rollback failed: %w", txErr)
 	}
 	if err := tx.Commit(); err != nil {
 		return nil, err
--- a/pkg/dotc1z/rollback_expansion.go
+++ b/pkg/dotc1z/rollback_expansion.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
 	"go.uber.org/zap"
@@ -44,14 +46,40 @@ type RollbackResult struct {
 	// SourcesCleared is the number of surviving direct grants whose
 	// Sources were reset to empty, restoring the pre-expansion shape.
 	SourcesCleared int
-	// SuspectConnectorSourced counts deleted grants that had Sources but
-	// no self-source AND lacked the GrantImmutable annotation that every
+	// SuspectConnectorSourced counts grants that had Sources but no
+	// self-source AND lacked the GrantImmutable annotation that every
 	// expander-created grant carries — a shape that looks connector-set
-	// rather than expander-set. These are still deleted; the count is
-	// surfaced so a real connector-sourced grant being swept up can be
-	// detected. Zero in the common case.
+	// rather than expander-set. By default these are deleted (and counted
+	// in GrantsDeleted); with WithPreserveSuspectGrants they are kept
+	// instead (counted in SuspectPreserved, not GrantsDeleted). Either way
+	// the count is surfaced so a real connector-sourced grant is visible.
+	// Zero in the common case.
 	SuspectConnectorSourced int
-	DryRun                  bool
+	// SuspectPreserved counts suspect grants kept rather than deleted
+	// because WithPreserveSuspectGrants was set. Always 0 in the default
+	// (delete) mode; a subset of SuspectConnectorSourced when set.
+	SuspectPreserved int
+	DryRun           bool
+}
+
+// rollbackConfig holds the tunable behavior of a rollback.
+type rollbackConfig struct {
+	preserveSuspect bool
+}
+
+// RollbackOption configures RollbackExpansion.
+type RollbackOption func(*rollbackConfig)
+
+// WithPreserveSuspectGrants makes rollback KEEP suspect connector-sourced
+// grants (Sources present, no self-source, no GrantImmutable) instead of
+// deleting them. Default OFF: a suspect grant is an expander artifact in
+// the common case and is deleted, but because Grant.Sources is public
+// connector data and replay runs an empty connector, a genuinely
+// connector-set grant of that shape would be permanently dropped — this
+// option trades a possibly-stale expander grant left in the output for
+// never silently dropping real connector data.
+func WithPreserveSuspectGrants() RollbackOption {
+	return func(c *rollbackConfig) { c.preserveSuspect = true }
 }
 
 // queryContexter is the read surface shared by *goqu.Database and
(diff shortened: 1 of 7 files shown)


btipling:
this is not ready for reviews it has issues

kans:
For validation, lets also make sure that the grant sources are the same before and after, too.

kans:
Curated AI feedback:

1. `pkg/dotc1z/rollback_expansion.go`: `RollbackExpansion` deletes any grant with `Grant.Sources` but no self-source. It only increments `SuspectConnectorSourced` when the grant lacks `GrantImmutable`, but still deletes it. Since `Grant.Sources` is public connector data and replay uses an empty connector, this can permanently drop real connector-sourced grants from the output c1z. Consider refusing rollback when suspect grants exist or preserving them.

2. `pkg/dotc1z/rollback_expansion.go` / `pkg/dotc1z/c1file.go`: rollback mutates grant rows and resets `sync_token`, but does not clear/recompute `sync_runs.stats`. `Stats()` returns cached stats when present, so `baton stats` on a rollback-without-replay output may report the pre-rollback grant count. Clear `stats` during rollback or recompute after mutation/replay.

3. Performance: rollback scans grants with:
   `SELECT id, entitlement_id, data FROM v1_grants WHERE sync_id = ? AND id > ? ORDER BY id LIMIT 1000`
   but the grants table has no `(sync_id, id)` index. Direct dry-run on multi-sync files can scan unrelated rows. Consider changing the scan to use an existing selective index/path.`.

pkg/dotc1z/rollback_expansion.go
```diff
@@ -0,0 +1,261 @@
...
+
+// ErrSyncNotExpanded is returned when a rollback targets a finished sync
+// that never ran grant expansion (no supports_diff marker). There is
+// nothing to roll back, and a missing marker also means a resumable
+// graph was never produced.
+var ErrSyncNotExpanded = errors.New("c1z: sync did not run grant expansion (no supports_diff marker); nothing to roll back")
+
+// rollbackPageSize bounds how many grant rows are read per round so a
+// large sync does not buffer the whole table in memory.
+const rollbackPageSize = 1000
```
ggreer, line 36: This could probably be 10k rows. I don't know if you benchmarked it.

pkg/dotc1z/rollback_expansion.go
```diff
@@ -0,0 +1,395 @@
...
+	cfg := &rollbackConfig{}
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	sr, err := c.getSync(ctx, syncID)
+	if err != nil {
+		return nil, fmt.Errorf("c1z: rollback could not load sync %q: %w", syncID, err)
+	}
+	if sr.EndedAt == nil {
```
ggreer, line 130: Should we error if the sync type is not something that can contain grants? Such as a resource only sync. I could see "correct" behavior being either option. If we do detect a sync without grants, we could probably return early.

pkg/dotc1z/rollback_expansion.go
```diff
@@ -0,0 +1,395 @@
...
+		// makes the syncer seed a fresh InitOp, which pushes the grant-
+		// expansion step.
+		//
+		// stats is cleared in the same statement: rollback mutates grant
+		// rows, but Stats() returns the cached sync_runs.stats verbatim when
+		// present, so without clearing it `baton stats` on a
+		// rollback-without-replay output reports the pre-rollback grant
+		// count. NULL stats forces Stats()'s slow path to recompute from the
+		// rolled-back rows (replay recomputes stats at sync end regardless).
+		resetQuery := fmt.Sprintf("UPDATE %s SET sync_token = '', stats = NULL WHERE sync_id = ?", syncRuns.Name())
```
ggreer, line 180: I think we should parse the sync token and set NeedsExpansion to true. That would help us avoid certain bugs that might arise if we delete other stuff in the sync token.

pkg/dotc1z/rollback_expansion.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+		return res, nil
+	}
+
+	tableName := grants.Name()
+	tx, err := c.db.BeginTx(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	txErr := func() error {
```
ggreer, line 155: This is a pretty big transaction since it's updating/deleting lots grants. In some cases that could be the majority of grants in the table. Do we want to avoid a transaction here so we don't make a giant rollback journal?

pkg/dotc1z/rollback_expansion.go
```diff
@@ -0,0 +1,395 @@
...
+	return nil
+}
+
+// GrantSourcesForSync returns, for every grant in the sync, a canonical
+// string form of its GrantSources keyed by the grant's external id. It is
+// used to validate a rollback+replay round trip: every grant present
+// before must be present after with identical Sources, since replay
+// re-derives exactly what rollback removed. Pages off the same
+// (external_id, sync_id) index the classify scan uses.
+func (c *C1File) GrantSourcesForSync(ctx context.Context, syncID string) (map[string]string, error) {
```
ggreer, line 334: For a c1z with lots of grants, this could take up a lot of memory. I think you could use the StreamGrants() method in a loop to avoid reading all the grants into a giant map.

At the very least, I think this function could live over in the baton command and not on C1File. We probably have a method on the C1File that reads grants by sync ID that works for this use case (such as StreamGrants or just ListGrants).

btipling:
STATS+SOURCES PRESERVED, exact identity, independently of the PR's own --validate

kans (approved):

PR #909, opened 2026-06-02
Propagate grant cancellation status details

alan-lee-12:
_Follow-up to #887._

Add grant status helpers to encode and decode grant cancellation markers in google.rpc.Status, and preserve those details through c1api task finish and lambda gRPC error responses.


alan-lee-12:
closing in favor of #913

PR #910, opened 2026-06-02
feat(sync): stamp connector identity on sync/dotc1z spans + c1z size metric

arreyder:
## Why

Sync and dotc1z spans already exist (`syncer.Sync`, `syncer.SyncResources/Grants/Entitlements`, `C1File.Vacuum/Cleanup/finalize`, …) but carry **no connector identity**, so a single connector's sync/c1z work can't be filtered per-connector in APM. In a recent incident a connector's sync silently looped + VACUUMed for days and was only findable via CPU **profile labels** — the spans were useless for it. This closes that gap and adds the c1z-size signal that loop would have surfaced immediately.

## What

- **`uotel.SyncIdentity`** — context-propagated `{tenant_id, connector_id, catalog_id, catalog_name}`. Attribute keys match the platform's `pprof.Do` labels, so spans and CPU profiles line up on the same dimensions.
- **`sync.WithSyncIdentity(...)` option** — `Sync()` injects identity into the run context, so every descendant span (sync **and** dotc1z, via the threaded ctx) gets stamped. Explicit option wins; otherwise inherits whatever the caller already set on ctx.
- **Identity stamped** on every linked-root sync operation span + `C1File.Vacuum/Cleanup/finalize/saveC1z`.
- **Bounded summary span** on the parallel fan-out (`syncParallel`): one span per batch with `sync.op` / `sync.action_count` / `sync.worker_count`. The per-action work already starts its own `StartWithLink` root, so this stays a handful of spans per sync — **traces do not grow with resource/grant count**.
- **`C1File.Vacuum`** records `c1z.vacuum.{size_before,size_after,reclaimed}_bytes`.
- **`c1z_size_bytes` gauge** (`phase=vacuum|finalize`), tagged **`catalog_name` only** — `connector_id` is too high-cardinality for a metric and is already on the spans for per-connector drill-down.

## Trace-length safety

No new per-item spans. Identity stamping is attributes on existing spans; the only new span is one bounded summary per fan-out batch. The existing `StartWithLink` (new-root-linked-to-parent) pattern is preserved throughout.

## Wiring

Identity flows by context value, so this lights up once the platform caller sets it (the `ductone/c1` sync activity already computes `tenant_id`/`connector_id`/`catalog_id`/`catalog_name` for its `pprof.Do` labels — a follow-up passes the same values via `WithSyncIdentity` after this SDK release). Until then the changes are inert no-ops.

## Test

- New `pkg/uotel/identity_test.go` (Attrs/IsZero/context round-trip).
- `go build` + `go vet` clean (incl. `-tags=baton_lambda_support`); `pkg/uotel`, `pkg/dotc1z`, `pkg/sync` tests pass.

### DIFF
--- a/pkg/uotel/identity.go
+++ b/pkg/uotel/identity.go
@@ -19,6 +19,7 @@ type SyncIdentity struct {
 	CatalogName string
 }
 
+// IsZero reports whether id has no fields set.
 func (id SyncIdentity) IsZero() bool {
 	return id == SyncIdentity{}
 }
@@ -43,7 +44,9 @@ func (id SyncIdentity) Attrs() []attribute.KeyValue {
 
 type syncIdentityKey struct{}
 
-// WithSyncIdentity returns a context carrying id for downstream span stamping.
+// WithSyncIdentity returns a child of ctx carrying id. It only stores the
+// identity; spans are stamped where SetSyncIdentityAttrs(ctx, span) is called,
+// not here.
 func WithSyncIdentity(ctx context.Context, id SyncIdentity) context.Context {
 	return context.WithValue(ctx, syncIdentityKey{}, id)
 }
--- a/pkg/uotel/identity.go
+++ b/pkg/uotel/identity.go
@@ -14,6 +14,7 @@ import (
 // CPU profiles line up on the same dimensions.
 type SyncIdentity struct {
 	TenantID    string
+	AppID       string
 	ConnectorID string
 	CatalogID   string
 	CatalogName string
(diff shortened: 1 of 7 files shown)


pkg/uotel/identity.go
```diff
@@ -0,0 +1,66 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// SyncIdentity carries the platform identifiers for a connector sync. It is
+// propagated through the context so spans deep in the sync and dotc1z call
+// trees can be filtered to a single connector in APM. The attribute keys
+// match the pprof.Do labels set by the platform sync activity, so spans and
+// CPU profiles line up on the same dimensions.
+type SyncIdentity struct {
```
gontzess, line 15: since were sticking all these in here, should we include app id too? the tenant, app, connector ID trilogy gives us the ability to build dashboard links in alerting, etc

arreyder: Good call — added `AppID` to `SyncIdentity` + the `app_id` span attribute in 709e851 (additive, omitted when empty; tests extended). Also threaded it through the c1 wiring (ductone/c1#18658), including the `pprof.Do` labels, so spans and CPU profiles stay aligned on the same dimensions.

gontzess (approved):

PR #911, opened 2026-06-03
fix(dotc1z): delete sync runs in committed batches to survive retries

arreyder:
## Problem

`DeleteSyncRun` (called by `Cleanup`) issues one unbounded `DELETE FROM <table> WHERE sync_id = ?` per table. On a large c1z the grants DELETE is a **single multi-million-row transaction**. If the sync activity's deadline fires mid-statement, SQLite rolls the whole statement back — so a retry starts from zero, makes **no forward progress**, and cleanup loops until the connector's c1z is effectively abandoned.

Observed in prod: a sync spent ~1h+ in cleanup across retries and never finished deleting even the first table (no `Removed old sync data.` log, never reached VACUUM).

## Fix

Delete in bounded batches — `DELETE … WHERE id IN (SELECT id … WHERE sync_id=? LIMIT 50000)` in a loop per table. Each batch autocommits, so an interrupted/retried cleanup **resumes where it left off** instead of restarting. No API or schema change; same call sites.

## Why batches (not an index, not rebuild)

Backed by an autoresearch benchmark (1M-grant old generation + 100k keep, SQLite c1z):

| approach | delete time | notes |
|---|---:|---|
| baseline (current, one DELETE/table) | 18.1s | single transaction; rolls back on interrupt |
| **chunked (this PR)** | **12.4s** | ~30% faster **and** resumable |
| index `sync_id` then delete | 17.0s | no help — other grants indexes dominate |

Batching is the **smallest safe** win: it breaks the retry/timeout loop and is incidentally faster (smaller per-commit journal + index churn). It does not reclaim space on its own — that's still `Cleanup`'s VACUUM step, and a larger follow-up (rebuild-keep-into-fresh-file via the existing `CloneSync`, ~11× faster + self-compacting) is the deeper fix.

## Test

- Full `pkg/dotc1z` suite passes (incl. `TestCleanupVacuum`, clone/diff paths that use `DeleteSyncRun`).
- `gofmt` + `go vet` clean.

### DIFF
--- a/pkg/dotc1z/sync_runs.go
+++ b/pkg/dotc1z/sync_runs.go
@@ -31,7 +31,10 @@ const syncRunsTableName = "sync_runs"
 
 // deleteSyncRunBatchSize bounds each DELETE in DeleteSyncRun so progress
 // commits incrementally and survives an interrupted/retried cleanup.
-const deleteSyncRunBatchSize = 50000
+// Var (not const) only so tests can shrink it to exercise the multi-batch
+// loop without inserting 50k rows; not tuned at runtime.
+var deleteSyncRunBatchSize = 50000
+
 const syncRunsTableSchema = `
 create table if not exists %s (
     id integer primary key,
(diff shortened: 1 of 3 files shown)


pkg/dotc1z/sync_runs.go
```diff
@@ -901,20 +908,39 @@ func (c *C1File) DeleteSyncRun(ctx context.Context, syncID string) error {
...
+			}
+			n, raErr := res.RowsAffected()
+			if raErr != nil {
+				err = fmt.Errorf("failed to read rows affected for sync run delete: %w", raErr)
+				return err
+			}
+			deleted += n
+			if n == 0 {
+				break
+			}
```
ggreer: I just noticed we probably have a bug in the original code: We iterate over allTableDescriptors, and syncRuns is before grants, assets, and sessionStore. We should probably delete from syncRuns last, otherwise we'll keep old sync data in the other tables.

ggreer (approved):

PR #914, opened 2026-06-03
Add options to CloneSync().

ggreer:
Also add NeedsExpansion() helper to check if a c1z needs expansion.

These are useful for speeding up syncs for service mode connectors in c1.

### DIFF
--- a/pkg/sync/state.go
+++ b/pkg/sync/state.go
@@ -62,6 +62,15 @@ type State interface {
 	ClearExclusionGroupTracking(ctx context.Context)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/sync/state.go
```diff
@@ -62,6 +62,15 @@ type State interface {
 	ClearExclusionGroupTracking(ctx context.Context)
 }
 
+func NeedsExpansion(stateStr string) bool {
+	state := newState()
+	err := state.Unmarshal(stateStr)
+	if err != nil {
```
kans, line 68: Return an err.

kans (approved):

PR #917, opened 2026-06-04
fix(dotc1z): drop malformed grants (nil principal/entitlement) at write time

### LINKED ISSUE #916: dotc1z silently persists malformed grants (nil principal/entitlement) with empty identity columns
## Summary

The local grant write path (`pkg/dotc1z`) does not validate that a grant has a principal and entitlement, so a malformed grant is **silently persisted with empty identity columns** rather than rejected. Every other layer of the SDK already treats principal/entitlement as required — the write path is the lone gap.

## Evidence

| Layer | Behavior |
|---|---|
| Proto (`proto/c1/connector/v2/grant.proto`) | `principal` and `entitlement` are both `(validate.rules).message = {required: true}` |
| Builder (`pkg/types/grant/grant.go` `NewGrant()`) | **panics** on a nil principal |
| Task handler (`pkg/tasks/c1api/grant.go:35`) | **rejects** nil principal/entitlement as a "malformed grant task" (non-retryable) |
| **Write path** (`pkg/dotc1z/grants.go` `upsertGrantsInternal` → `baseGrantRecord`) | **no check** |

`baseGrantRecord` reads `grant.GetPrincipal().GetId().GetResourceType()` etc. via nil-safe opaque getters. So a nil-principal grant does **not** panic — it writes `principal_resource_id = ""` / `entitlement_id = ""` (empty satisfies the `not null` constraint). The row is unqueryable by principal and invisible as a problem.

Verified locally: `PutGrants(nilPrincipal, good, nilEntitlement)` persists all three rows; two carry empty identity columns.

This was surfaced during review of #863 (grant-expansion perf). Expansion already defends against nil-principal grants by skipping them, but the right fix is to stop the bad data entering the store in the first place.

## Proposed fix

Validate in `upsertGrantsInternal` (covers both `PutGrants` and `StoreExpandedGrants`). PR #917 implements **drop-and-continue**: malformed grants are dropped, logged (with grant id), and counted via a new `c1z_grant_malformed_dropped_total` metric — so one buggy connector grant can't abort an entire sync's grant write.

## Open decision (needs a call)

Drop-and-continue (PR default, resilient — matches the graceful-degradation already used in expansion) **vs.** hard-fail the batch (strict, matches the task handler's non-retryable rejection). Flagging for @ggreer — the PR is a one-line change away from either policy.

## Out of scope

Enforcing the proto `validate.rules` at the connector/gRPC boundary, and whether to reject vs. coerce elsewhere. This issue is specifically about the local c1z write path silently persisting junk.

arreyder:
## Summary

Drop malformed grants (nil principal or nil entitlement) at the c1z write path instead of silently persisting them with empty identity columns.

Fixes #916.

## Problem

`baseGrantRecord` extracts a grant's principal/entitlement identity columns straight off the live proto using nil-safe opaque getters. A grant with a nil principal or entitlement therefore does **not** panic and is **not** rejected — it's written with `principal_resource_id = ""` / `entitlement_id = ""` (empty satisfies the `not null` constraint). The result is an unqueryable junk row that nothing surfaces.

Every other layer already treats these fields as required:
- `grant.proto` marks `principal` and `entitlement` `(validate.rules).message = {required: true}`
- `NewGrant()` panics on a nil principal
- the grant task handler (`pkg/tasks/c1api/grant.go:35`) rejects nil principal/entitlement as a non-retryable "malformed grant task"

The local write path was the only place that let them through. Surfaced during review of #863.

## Fix

`upsertGrantsInternal` (which backs both `PutGrants` and `StoreExpandedGrants`) now runs `dropMalformedGrants` first:
- drops any grant with a nil principal or entitlement
- logs the dropped grant id at error level
- increments a new `c1z_grant_malformed_dropped_total` metric
- fast-paths (no reallocation) when every grant is well-formed

**Policy choice — drop-and-continue vs. hard-fail.** This PR drops-and-continues so one buggy connector grant can't abort an entire sync's grant write, matching the graceful-degradation already used in grant expansion. The strict alternative (fail the batch, matching the task handler) is a one-line change. @ggreer — flagging for your call; happy to flip it.

## Test plan

- [x] `TestPutGrants_DropsMalformedGrants` — interleaved `nilPrincipal, good, nilEntitlement` batch → only the well-formed grant persists
- [x] `TestPutGrants_AllMalformedIsNoop` — all-malformed batch writes nothing, no error
- [x] `TestDropMalformedGrants_AllValidReturnsSameSlice` — fast path returns input untouched
- [x] Pre-fix verification: with the drop disabled, all three grants persist (two with empty identity columns)
- [x] `go test -short ./pkg/dotc1z/...` — pass
- [x] `golangci-lint run pkg/dotc1z/...` — 0 issues

### DIFF
--- a/pkg/dotc1z/dotc1z.go
+++ b/pkg/dotc1z/dotc1z.go
@@ -34,6 +34,14 @@ var (
 		metric.WithDescription("Grants kept full-blob by the unsafeForSlim escape hatch on slim-enabled writers."),
 	)
 
+	// Increments when a malformed grant (nil principal or entitlement) is
+	// dropped at write time instead of being persisted. A non-zero value
+	// means an upstream connector is emitting out-of-contract grants.
+	grantMalformedCounter, _ = meter.Int64Counter(
+		"c1z_grant_malformed_dropped_total",
+		metric.WithDescription("Grants dropped at write time for missing a principal or entitlement."),
+	)
+
 	// Pre-built so we don't allocate a fresh attribute.Set on every
 	// grant write. WithAttributes sorts and dedups on each call;
 	// WithAttributeSet on a shared Set is allocation-free.
--- a/pkg/dotc1z/grants.go
+++ b/pkg/dotc1z/grants.go
@@ -560,17 +560,37 @@ func slimGrantForWrite(grant *v2.Grant) {
 	grant.SetPrincipal(nil)
 }
 
+// maxMalformedLogsPerBatch caps how many individual malformed grants are
+// logged per batch before collapsing to a single aggregate line, so a broken
+// connector emitting a run of bad grants can't flood the logs. The exact count
+// is always available via grantMalformedCounter.
+const maxMalformedLogsPerBatch = 10
+
+// grantHasValidIdentity reports whether a grant carries the identity a stored
+// row needs to be queryable. baseGrantRecord extracts these columns straight
+// off the proto with nil-safe getters, so a missing OR empty value would be
+// written as unqueryable junk (and an empty grant id additionally collides on
+// the (external_id, sync_id) unique index). The proto contract
+// (grant.proto) makes all of these required.
+func grantHasValidIdentity(g *v2.Grant) bool {
+	pid := g.GetPrincipal().GetId()
+	return g.GetId() != "" &&
+		pid.GetResourceType() != "" && pid.GetResource() != "" &&
+		g.GetEntitlement().GetId() != ""
+}
+
 // dropMalformedGrants returns the grants that are safe to persist, dropping
-// any with a nil principal or entitlement. Dropped grants are logged (with the
-// grant id) and counted via grantMalformedCounter. We drop-and-continue rather
-// than failing the batch so one buggy connector grant can't abort an entire
-// sync's grant write; the metric and log make the bad data visible.
+// any missing required identity (principal, entitlement, or grant id — nil or
+// empty). Dropped grants are logged (capped per batch) and counted via
+// grantMalformedCounter. We drop-and-continue rather than failing the batch so
+// one buggy connector grant can't abort an entire sync's grant write; the
+// metric and log make the bad data visible.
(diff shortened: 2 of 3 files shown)


arreyder:
## Update — review-driven additions (626ff10e)

Ran a multi-agent review (bugs / perf / test) over the diff. Core logic came back clean (fast-path correctness, replace-mode has no stale-row risk, nil-element safety, no caller depends on written count). Two substantive additions came out of it:

1. **Broadened the malformed check.** The original nil-only test let through grants with a non-nil but *empty* identity — e.g. a principal carrying an empty `ResourceId{}`, or an empty entitlement/grant id — which `baseGrantRecord` then persisted as the exact empty-column junk the drop exists to prevent (and an empty grant id collides on the `(external_id, sync_id)` unique index). New `grantHasValidIdentity` rejects nil **or** empty principal resource type/id, entitlement id, and grant id. Verified the old check let an empty-`ResourceId` principal through.

2. **Capped per-batch drop logging** to the first 10 grants + one aggregate line, so a broken connector emitting a run of malformed grants can't flood the logs. The exact count stays on `c1z_grant_malformed_dropped_total`.

**Test additions:** parameterized the drop test over full-blob **and** slim writers, added a `StoreExpandedGrants` (PreserveExpansion) drop test, added empty-identity fixtures + a direct `grantHasValidIdentity` boundary test, and fixed the fast-path test to assert slice identity (`require.Same`) rather than just contents.

**Consciously deferred:** asserting the `grantMalformedCounter` increment requires wiring an OTel manual-reader `MeterProvider` — moderate harness cost for modest value given the drop is well-tested. Happy to add if you'd like it.

The drop-and-continue vs. hard-fail policy question for @ggreer is unchanged.

kans:
I think we should reject these and error out when the connector surfaces them to the SDK?  @ggreer thoughts?

ennyjfrick (approved):

PR #922, opened 2026-06-04
feat: add AdditionalGrantsRevoked annotation for revoke responses

mstanbCO:
## Summary
- Adds a new `AdditionalGrantsRevoked` annotation (`proto/c1/connector/v2/annotation_additional_grants_revoked.proto`) that a connector can attach to a `GrantManagerServiceRevokeResponse`.
- It carries a repeated list of `revoked_grant_ids` — additional grants the connector had to revoke in order to service a single Revoke request (e.g. flipping a "has access" flag before removing a principal's last role).
- ConductorOne can use this to mark those grants revoked without dispatching redundant Revoke RPCs.
- Includes generated `pb` code and Go helpers `grant.NewAdditionalGrantsRevoked(...)` / `grant.AppendAdditionalGrantsRevoked(...)`, mirroring the existing `GrantReplaced` pattern.

## Usage
```go
annos := annotations.Annotations{}
annos.Append(grant.NewAdditionalGrantsRevoked(hasAccessGrant.GetId()))
return annos, nil
```

## Test plan
- [x] `buf generate` produces the pb files
- [x] `go build ./...`
- [x] `go test ./pkg/types/grant/` passes (new helper tests included)

Made with [Cursor](https://cursor.com)


kans:
@mstanbCO do we still want this?

PR #924, opened 2026-06-04
feat(synccompactor): select the c1z storage engine for compaction

btipling:
## What

Lets a caller choose the c1z storage engine for sync compaction. Today compaction is hardwired to
SQLite — the output is always a v1 SQLite `.c1z` and the merge always uses the ATTACH-database path —
so a Pebble-backed connector can never produce or read compacted syncs. This adds a wired-in option
so compaction can produce **SQLite (default, unchanged) or Pebble**.

## How

- **`synccompactor.WithEngine(dotc1z.Engine)`** — stores the engine as a strategy field on the
  `Compactor`. Unset is SQLite, byte-identical to before. The resolved engine is forced onto the
  output constructor last so an engine passed via `WithC1ZOptions` can't mislabel the artifact.
- **Branch in `doOneCompaction`/`Compact`:** the SQLite path (`NewC1ZFile` + the `attached` ATTACH
  merge) is untouched. The Pebble path opens inputs with the engine-aware `NewStore` (so v3 inputs
  decode), selects each input's latest finished compactable sync (diff syncs `partial_upserts`/
  `partial_deletions` excluded), and folds them all into the empty destination sync in a single pass.
- **Native record merge (`pkg/synccompactor/pebble.MergeInto`):** reads each input's primary records
  (resource types, resources, entitlements, grants), re-keys them to the destination sync, and writes
  them keep-newer — the strictly-greater `discovered_at` wins and an equal `discovered_at` keeps the
  earlier-applied record, matching the SQLite union exactly. Indexes are rebuilt from the surviving
  records by the engine's write path. **Assets are not copied**, matching the SQLite path (which folds
  only those four tables); copying them would give Pebble compacted syncs asset access SQLite does not
  have. The destination `sync_run` type and `ended_at` are set to the union / max across the inputs,
  and stats are recomputed, so downstream gating (e.g. grant expansion) behaves identically.
- **`pebble.AsEngine`** recovers the `*Engine` from the `NewStore` wrapper for the merge; the
  compactor registers the Pebble driver on demand (a caller may also register it explicitly).

## Tests

- `pkg/synccompactor/pebble/merge_test.go` — multi-source union with newer-wins; equal-`discovered_at`
  tie keeps the earlier-applied incumbent (matches SQLite strict `>`); sync-id re-keying.
- `pkg/synccompactor/compactor_pebble_test.go` — end-to-end: two v3 Pebble inputs compacted via
  `WithEngine(EnginePebble)` yield one merged Pebble sync with the unioned, deduped grants and the
  union sync type.
- The SQLite compaction path is exercised unchanged by the existing suite.

`go build ./...`, `go test ./pkg/synccompactor/... ./pkg/dotc1z/...`, `gofmt`, and `golangci-lint`
on the changed packages are green.

## Scope notes

Pebble compaction requires Pebble inputs (and SQLite compaction SQLite inputs); a mixed-engine input
set is not supported. The merge re-encodes each record to re-key the sync id, so it is record-paced
rather than a byte-level ingest — a future optimization could ingest unchanged buckets directly.

---

_🛰️ Built with stargate._

### DIFF
--- a/pkg/dotc1z/engine/pebble/cleanup.go
+++ b/pkg/dotc1z/engine/pebble/cleanup.go
@@ -31,6 +31,7 @@ func syncScopedRanges(syncIDBytes []byte) [][2][]byte {
 		{EntitlementByResourceSyncLowerBound(syncIDBytes), EntitlementByResourceSyncUpperBound(syncIDBytes)},
 		{encodeGrantPrefix(syncIDBytes), upperBoundOf(encodeGrantPrefix(syncIDBytes))},
 		{GrantByEntitlementSyncLowerBound(syncIDBytes), GrantByEntitlementSyncUpperBound(syncIDBytes)},
+		{GrantByEntitlementResourceSyncLowerBound(syncIDBytes), GrantByEntitlementResourceSyncUpperBound(syncIDBytes)},
 		{GrantByPrincipalSyncLowerBound(syncIDBytes), GrantByPrincipalSyncUpperBound(syncIDBytes)},
 		{GrantByPrincipalResourceTypeSyncLowerBound(syncIDBytes), GrantByPrincipalResourceTypeSyncUpperBound(syncIDBytes)},
 		{GrantByNeedsExpansionSyncLowerBound(syncIDBytes), GrantByNeedsExpansionSyncUpperBound(syncIDBytes)},
--- a/pkg/dotc1z/engine/pebble/cleanup_test.go
+++ b/pkg/dotc1z/engine/pebble/cleanup_test.go
@@ -1,6 +1,7 @@
 package pebble
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"os"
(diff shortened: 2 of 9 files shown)


kans:
Uncrrated AI Feedback:

```Please fix Pebble cleanup’s sync-scoped range list.

In pkg/dotc1z/engine/pebble/cleanup.go, syncScopedRanges() is missing the grant_by_entitlement_resource secondary index range. That index is written by grantIndexKeys()/writeGrantIndexes(), used by PaginateGrantsByEntitlementResource(), and included in CloneSync and synccompactor bucket plans, but DeleteSyncData()/CompactSyncRanges() do not currently delete/compact it.

Add:
  {GrantByEntitlementResourceSyncLowerBound(syncIDBytes), GrantByEntitlementResourceSyncUpperBound(syncIDBytes)}

near the other grant index ranges in syncScopedRanges().

Also add a focused test that:
1. creates a Pebble sync with a grant whose entitlement resource is populated,
2. verifies the grant_by_entitlement_resource range has at least one key,
3. calls DeleteSyncData(syncID),
4. verifies that same range is empty.
```

kans:
We should do some benchmarking too, to make sure we aren't doing something really wrong in pebble (its easy).   

edit: This adds a new data migration path that does decodes/encodes analogous to the old compaction path in sqlite.

kans:
```
 go test ./pkg/synccompactor -run '^$' -bench BenchmarkCompactorEngineComparison -benchtime=1x

by_sync_count, fixed 2.5k grants/sync
syncs=2:   sqlite 0.138s, pebble 0.417s  = 3.0x slower
syncs=4:   sqlite 0.186s, pebble 0.476s  = 2.6x slower
syncs=8:   sqlite 0.323s, pebble 0.657s  = 2.0x slower
syncs=16:  sqlite 0.679s, pebble 1.221s  = 1.8x slower


by_sync_size, fixed 6 syncs
small:   sqlite 0.158s, pebble 0.468s  = 3.0x slower
medium:  sqlite 0.777s, pebble 1.304s  = 1.7x slower
large:   sqlite 3.576s, pebble 5.292s  = 1.5x slower


```

kans (approved):
I'm merging this, and will make a follow up PR to make it ~4x faster.

PR #926, opened 2026-06-05
Add SkipSyncAnomalyDetection annotation for expected data drops

laurenleach:
Add SkipSyncAnomalyDetection annotation for expected data drops

Sync anomaly detection flags a sync when a resource type's count drops significantly between syncs. For some resource types this drop is legitimate (e.g. invitations that disappear once accepted). This lets a connector mark such a resource type so anomaly detection ignores drops for it only; all other resource types are still checked.

https://linear.app/ductone/issue/CE-748/ignore-resource-types-in-sync-anomaly-detection

### DIFF
--- a/pkg/connectorbuilder/connectorbuilder.go
+++ b/pkg/connectorbuilder/connectorbuilder.go
@@ -417,10 +417,11 @@ func (b *builder) GetCapabilities(ctx context.Context) (*v2.ConnectorCapabilitie
 		}
 
 		resourceTypeCapabilities = append(resourceTypeCapabilities, v2.ResourceTypeCapability_builder{
-			ResourceType:  rb.ResourceType(ctx),
-			Capabilities:  caps,
-			Permissions:   p,
-			OptInRequired: annos.Contains(&v2.OptInRequired{}),
+			ResourceType:             rb.ResourceType(ctx),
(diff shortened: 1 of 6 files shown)


pkg/connectorbuilder/connectorbuilder.go
```diff
@@ -417,10 +417,11 @@ func (b *builder) GetCapabilities(ctx context.Context) (*v2.ConnectorCapabilitie
...
 		resourceTypeCapabilities = append(resourceTypeCapabilities, v2.ResourceTypeCapability_builder{
-			ResourceType:  rb.ResourceType(ctx),
-			Capabilities:  caps,
-			Permissions:   p,
-			OptInRequired: annos.Contains(&v2.OptInRequired{}),
+			ResourceType:             rb.ResourceType(ctx),
+			Capabilities:             caps,
+			Permissions:              p,
+			OptInRequired:            annos.Contains(&v2.OptInRequired{}),
+			SkipSyncAnomalyDetection: annos.Contains(&v2.SkipSyncAnomalyDetection{}),
```
kans, line 424: Can we provide any more info on the shape or any means of verification?  For eg invites, are some patterns still suspect?

kans (approved):

PR #927, opened 2026-06-05
much much faster compaction for pebble only

kans:
Two benches: one with very small incrementals, and the other, with large ones.  We use the "overlay" algo if we can fit everything into memory, otherwise, we fall back to kway (which is much more disk intensive).   I expect to squeeze out another 20-30% on both pebble paths.

## Non-Skewed / Same-Size - middle 6 /8

| syncs | engine | avg time/op | avg bytes/op | avg allocs/op |
|---:|---|---:|---:|---:|
| 10 | sqlite | 391.1ms | 55.7MB | 30,972 |
| 10 | kway | 511.1ms | 109.1MB | 22,557 |
| 10 | overlay | 404.1ms | 75.2MB | 15,800 |
| 50 | sqlite | 2.09s | 239.1MB | 144,272 |
| 50 | kway | 824.9ms | 214.3MB | 66,099 |
| 50 | overlay | 784.8ms | 213.8MB | 55,658 |
| 500 | sqlite | 25.93s | 2.30GB | 1,419,838 |
| 500 | kway | 4.72s | 869.7MB | 579,807 |
| 500 | overlay | 4.76s | 844.6MB | 555,171 |

## Skewed - middle 6 / 8

| syncs | engine | avg time/op | avg bytes/op | avg allocs/op |
|---:|---|---:|---:|---:|
| 10 | sqlite | 163.1ms | 21.3MB | 30,241 |
| 10 | kway | 505.0ms | 96.7MB | 18,792 |
| 10 | overlay | 310.5ms | 41.4MB | 15,200 |
| 50 | sqlite | 456.4ms | 51.3MB | 140,314 |
| 50 | kway | 490.2ms | 113.3MB | 56,668 |
| 50 | overlay | 354.8ms | 57.3MB | 52,354 |
| 500 | sqlite | 3.98s | 390.0MB | 1,379,003 |
| 500 | kway | 1.10s | 355.0MB | 484,927 |
| 500 | overlay | 896.0ms | 233.4MB | 467,478 |

### DIFF
--- a/pkg/dotc1z/engine/equivalence/memory.go
+++ b/pkg/dotc1z/engine/equivalence/memory.go
@@ -47,7 +47,9 @@ func (m *MemoryRef) PutGrantRecord(_ context.Context, r *v3.GrantRecord) error {
 	}
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	sid := m.resolveSync(r.GetSyncId())
+	// GrantRecord values do not carry a sync_id in v3; writer paths bind
+	// grants to the current sync's keyspace.
+	sid := m.resolveSync("")
 	if sid == "" {
 		return fmt.Errorf("memoryref: no sync_id")
 	}
--- a/pkg/dotc1z/engine/pebble/grants.go
+++ b/pkg/dotc1z/engine/pebble/grants.go
@@ -80,14 +80,19 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
 		// in-sync calls already committed (e.g. paginated sources
 		// emitting an external_id on two pages).
 		skipGet := e.takeFreshGrantsEmpty()
+		// GrantRecord values do not carry sync_id in v3; records are
+		// written under the engine's currently bound sync.
+		idBytes, err := e.resolveSyncBytes("")
+		if err != nil {
+			return err
+		}
 
 		// Dedup pre-pass: keep only the LAST occurrence of each
-		// (sync_id, external_id). The map value is the records[]
+		// external_id. The map value is the records[]
 		// index — when we re-iterate, we process record i only if
-		// dedup[(sync,ext)] == i.
+		// dedup[ext] == i.
 		type dedupKey struct {
-			syncID string
-			extID  string
+			extID string
 		}
 		var dedup map[dedupKey]int
 		if len(records) > 1 {
@@ -96,42 +101,18 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
 				if r == nil {
 					continue
 				}
-				dedup[dedupKey{r.GetSyncId(), r.GetExternalId()}] = i
+				dedup[dedupKey{r.GetExternalId()}] = i
 			}
 		}
 
-		// Cache the resolved sync_id across records sharing the same
-		// string. The adapter's V2→V3 translator stamps every record
-		// with the engine's current sync, so the common case is one
-		// distinct sync_id per call.
-		var (
-			lastSyncIDStr string
-			lastIDBytes   []byte
-			haveLast      bool
-		)
 		for i, r := range records {
 			if r == nil {
 				continue
 			}
 			if dedup != nil {
-				if dedup[dedupKey{r.GetSyncId(), r.GetExternalId()}] != i {
+				if dedup[dedupKey{r.GetExternalId()}] != i {
 					continue
 				}
-			}
-			var (
-				idBytes []byte
-				err     error
-			)
-			if sid := r.GetSyncId(); haveLast && sid == lastSyncIDStr {
-				idBytes = lastIDBytes
-			} else {
-				idBytes, err = e.resolveSyncBytes(sid)
-				if err != nil {
-					return err
-				}
-				lastSyncIDStr = sid
-				lastIDBytes = idBytes
-				haveLast = true
 			}
 			key := encodeGrantKey(idBytes, r.GetExternalId())
 			val, err := marshalRecord(r)
@@ -142,12 +123,7 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
 				oldVal, closer, getErr := e.db.Get(key)
 				switch {
 				case getErr == nil:
-					old := &v3.GrantRecord{}
-					if err := unmarshalRecord(oldVal, old); err != nil {
-						closer.Close()
-						return fmt.Errorf("PutGrantRecords: unmarshal old %q: %w", r.GetExternalId(), err)
-					}
-					if err := e.deleteGrantIndexes(idxBatch, idBytes, old); err != nil {
+					if err := e.deleteGrantIndexesRaw(idxBatch, idBytes, r.GetExternalId(), oldVal); err != nil {
 						closer.Close()
 						return err
 					}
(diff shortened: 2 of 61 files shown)


pkg/dotc1z/engine/equivalence/memory.go
```diff
@@ -47,7 +47,7 @@ func (m *MemoryRef) PutGrantRecord(_ context.Context, r *v3.GrantRecord) error {
 	}
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	sid := m.resolveSync(r.GetSyncId())
+	sid := m.resolveSync("")
```
ggreer, line 52: Can't this resolve to the wrong sync ID?

pkg/dotc1z/format/v3/envelope.go
```diff
@@ -125,18 +127,61 @@ type Envelope struct {
...
+			e.zstdReader.Close()
+		}
 		e.zstdReader = nil
+		e.zstdReaderPooled = false
 	}
 	return nil
 }
 
+var zstdPoolDisabled = os.Getenv("BATON_ZSTD_POOL_DISABLE") == "1"
+var zstdDecoderPool = &sync.Pool{}
```
ggreer: Is this pool shared for the whole process? If that's the case, I think it might make temporal-sync & temporal-worker slower. We also do this for sqlite c1zs, so we should probably figure this out for both.

pkg/dotc1z/engine/pebble/grants.go
```diff
@@ -80,14 +80,17 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
 		// in-sync calls already committed (e.g. paginated sources
 		// emitting an external_id on two pages).
 		skipGet := e.takeFreshGrantsEmpty()
+		idBytes, err := e.resolveSyncBytes("")
```
ggreer, line 85: Also not sure this is resolving to the correct sync ID.

pkg/dotc1z/engine/pebble/raw_records.go
```diff
@@ -0,0 +1,407 @@
...
+)
+
+const (
+	resourceTypeDiscoveredAtField protowire.Number = 6
+	resourceDiscoveredAtField     protowire.Number = 8
+	entitlementDiscoveredAtField  protowire.Number = 8
+	grantDiscoveredAtField        protowire.Number = 5
+)
+
+var validateRawStringFields = os.Getenv("BATON_VALIDATE_RAW_STRING_FIELDS") == "1"
```
ggreer: Can we ditch this or is it important that we sometimes validate raw string fields?

pkg/dotc1z/format.go
```diff
@@ -72,18 +72,15 @@ const (
 	// engine's default" — TarZstd for Pebble.
 	PayloadEncodingUnspecified PayloadEncoding = 0
 
-	// 1 and 2 are reserved in the proto (formerly RAW + single-stream
-	// ZSTD). Don't reuse.
-
 	// PayloadEncodingTarZstd is the default Pebble v3 envelope
 	// encoding: tar of the Pebble directory, compressed with zstd.
-	PayloadEncodingTarZstd PayloadEncoding = 3
+	PayloadEncodingTarZstd PayloadEncoding = 1
```
ggreer: I agree with the bot. We should probably not change these values.

pkg/synccompactor/compactor_pebble.go
```diff
@@ -73,6 +76,74 @@ func unionV3SyncType(a, b v3.SyncType) v3.SyncType {
 	}
 }
 
+// manifestSourceSelection is the result of picking a compaction source
+// sync from the envelope manifest's sync-run projection.
+type manifestSourceSelection struct {
+	syncID   string
+	syncType v3.SyncType
+	endedAt  time.Time
+	stats    map[string]int64
```
ggreer: We have a SyncStats struct: https://github.com/ConductorOne/baton-sdk/blob/main/proto/c1/reader/v2/sync.proto#L21

I think we should try to use it as much as possible, though I don't think there's a public method in C1ZStore that returns it yet. We could probably change that and clean up some magic key names that are used in c1.

Could be done in a separate PR.

pkg/synccompactor/compactor_pebble.go
```diff
@@ -73,6 +76,74 @@ func unionV3SyncType(a, b v3.SyncType) v3.SyncType {
...
+	}
+	sel := manifestSourceSelection{
+		syncID:   best.GetSyncId(),
+		syncType: best.GetType(),
+		endedAt:  best.GetEndedAt().AsTime(),
+	}
+	// Same key shape as enginepkg.CachedSyncStats — consumed by the
+	// overlay's bucket planning (overlayEstimatedBucketKeys).
+	if s := best.GetStats(); s != nil {
+		sel.stats = map[string]int64{
```
ggreer: c1file.go has statsToMap(). Maybe it should be shared functionality so we don't accidentally have differing maps depending on the engine.

pkg/synccompactor/compactor_pebble.go
```diff
@@ -89,7 +160,9 @@ func (c *Compactor) compactPebble(ctx context.Context, newSyncId string) error {
 		return errors.New("compactPebble: compacted store is not a pebble engine")
 	}
 
-	sources := make([]mergepkg.SourceSync, 0, len(c.entries))
+	pebbleCompactorMode := os.Getenv("BATON_EXPERIMENTAL_PEBBLE_COMPACTOR")
```
ggreer: Another env var that we could probably replace with an option passed to NewStore(), or look at stats to figure out which compactor mode to use.

pkg/synccompactor/compactor_pebble.go
```diff
@@ -98,38 +171,81 @@ func (c *Compactor) compactPebble(ctx context.Context, newSyncId string) error {
...
-				l.Error("compactPebble: error closing input", zap.Error(cerr), zap.String("file", cs.FilePath))
-			}
-		}()
-
 		srcEng, ok := enginepkg.AsEngine(w)
 		if !ok {
+			_ = w.Close(ctx)
 			return fmt.Errorf("compactPebble: input %s is not a pebble c1z", cs.FilePath)
 		}
+		sourceStore := w
```
ggreer: Why alias this? It could be called w everywhere or sourceStore everywhere.

Also can you defer the close so that we don't have to make sure we close it before every return?

pkg/dotc1z/format/v3/envelope.go
kans: will pool for compaction

pkg/dotc1z/engine/equivalence/memory.go
kans: Yeah, it could!  Fixed.

pkg/synccompactor/compactor.go
```diff
@@ -204,6 +218,17 @@ func (c *Compactor) Compact(ctx context.Context) (*CompactableSync, error) {
 	fileName := fmt.Sprintf("compacted-%s.c1z", c.entries[0].SyncID)
 	destFilePath := path.Join(c.tmpDir, fileName)
 
+	if c.resolvedEngine() == dotc1z.EnginePebble {
```
ggreer, line 221: Not needed in this PR, but we should probably resolve the engine by looking at the c1z formats if no engine is specified in the compactor options. I ran into this bug playing around with compaction in c1.

ggreer (approved):

PR #928, opened 2026-06-08
Prevent formula injection in CSV/XLSX exports

pquerna:
## Summary
Add sanitization to exported CSV and XLSX files to prevent spreadsheet formula injection attacks. User-controlled data from upstream identity systems is now validated before export to prevent malicious formulas from being executed when files are opened in Excel, Google Sheets, LibreOffice, or other spreadsheet applications.

## Changes
- **New `sanitizeCell()` function** in `export.go`: Removes control characters (tabs, carriage returns, newlines) and prefixes cells starting with `=`, `+`, `-`, or `@` with a single quote to force literal text rendering
- **CSV export** (`csv.go`): Wrap all 12 cell values in the `csvRow.Row()` method with `sanitizeCell()`
- **XLSX export** (`xlsx.go`): Wrap all 12 cell values in the `excelRow.Row()` method with `sanitizeCell()`

## Implementation Details
The `sanitizeCell()` function:
1. Strips embedded control characters using `strings.Map()` with `unicode.IsControl()`
2. Checks if the resulting string begins with a formula indicator (`=`, `+`, `-`, `@`)
3. Prefixes with a single quote if needed, which spreadsheet applications interpret as a literal text marker
4. Returns the sanitized string

This approach ensures that attacker-controllable values (e.g., user display names, resource names) cannot be weaponized as formulas when exported data is opened in spreadsheet applications.

https://claude.ai/code/session_01Jxcdh6eSR6V8JX9G6iSefs

### DIFF
--- a/cmd/baton/csv.go
+++ b/cmd/baton/csv.go
@@ -220,18 +220,18 @@ type csvRow struct {
 
 func (c csvRow) Row() []string {
 	return []string{
-		c.rowType,
-		c.lastName,
-		c.firstName,
-		c.userID,
-		c.userStatus,
-		c.emailAddress,
-		c.entitlementDisplayName,
-		c.entitlement,
-		c.resourceType,
-		c.resourceName,
-		c.entitlementDescription,
-		c.entitlementSlug,
+		sanitizeCell(c.rowType),
+		sanitizeCell(c.lastName),
+		sanitizeCell(c.firstName),
+		sanitizeCell(c.userID),
+		sanitizeCell(c.userStatus),
+		sanitizeCell(c.emailAddress),
+		sanitizeCell(c.entitlementDisplayName),
+		sanitizeCell(c.entitlement),
+		sanitizeCell(c.resourceType),
+		sanitizeCell(c.resourceName),
+		sanitizeCell(c.entitlementDescription),
+		sanitizeCell(c.entitlementSlug),
 	}
 }
 
--- a/cmd/baton/export.go
+++ b/cmd/baton/export.go
@@ -1,8 +1,37 @@
 package main
 
 import (
+	"strings"
+	"unicode"
+
 	"github.com/spf13/cobra"
 )
+
+// sanitizeCell neutralizes spreadsheet formula injection in exported CSV/XLSX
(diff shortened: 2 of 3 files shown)


gontzess (commented):
**1. [issue] XLSX doesn't need the quote-prefix, and it corrupts data.** `excelRow.Row` writes values via `f.SetCellValue(sheet, cell, ...)`. For a Go string, excelize routes that to `SetCellStr`, which writes a **shared-string cell** (type `s`) and removes any formula — it never creates a formula. Excel renders shared strings as literal text, so the XLSX path was not vulnerable to formula injection. Because excelize stores the value verbatim and never sets the `quotePrefix` style, the leading `'` becomes part of the cell's data: `@handle` → `'@handle`, `-5` → `'-5`. Net result is altered data with no security gain. Suggest dropping `sanitizeCell` on the XLSX path (string cells are already safe), or, if you want defensive hardening, set the cell's `quotePrefix` *style* rather than mutating the string.

**2. [suggestion] No unit test for `sanitizeCell`.** It's a pure function — a small table test would lock the behavior: each prefix (`= + - @`) gets quoted, control chars stripped, `"\t=1+1"` → `"'=1+1"`, empty string, and a multibyte/no-op leading char.

**3. [suggestion] Full-width Unicode prefixes aren't caught.** The first-byte check only matches ASCII. OWASP's CSV-injection guidance also lists full-width variants (`＝ ＋ － ＠`, U+FF1D etc.) as test cases; a value starting with one slips through. Whether that's exploitable depends on the target app's Unicode normalization — worth a conscious decision rather than leaving it uncovered.

**4. [nit] Control-char stripping removes more than needed.** `unicode.IsControl` strips every control rune. Go's `csv.Writer` already quote-escapes embedded newlines/tabs, so stripping isn't required for CSV correctness — it just drops legitimate whitespace from names/descriptions. Fine as defense-in-depth, but a conscious tradeoff.

CSV approach is otherwise sound: control-char stripping covers tab/CR/LF, the `= + - @` set matches OWASP, and `csv.Writer` prevents separator breakout.

ggreer (dismissed):

PR #929, opened 2026-06-08
Add capabilities connector factory

MarcusGoldschmidt:


gontzess:
overall looks good to me. maybe we could add a command-level test proving the factory path skips config-field validation, invokes the factory only for capabilities, falls back to GetMetadata, and closes the returned connector?current tests cover the helpers well, but not the whole command path.

gontzess (approved):
could we add a description and a command level test if possible? otherwise lgtm!

PR #931, opened 2026-06-09
c1zsanitize: replace per-annotation log spam with structured progress logging

btipling:
## Summary

The c1zsanitize transform emitted a `c1zsanitize: dropping unknown annotation` Debug log on every dropped annotation inside `transformAnnotations`. On a real sanitize run that iterates tens of millions of records (the motivating case was a ~7 GB c1z with ~57M grants), that produced 3M+ log lines and gave an operator zero progress signal — which phase was running, how far it had gotten, or how much longer it had to go.

This PR removes the per-annotation log lines and replaces them with two pieces of structured logging that match the existing zap+ctxzap conventions in the package.

## Changes

- **Aggregate the dropped-annotation signal**: `transformAnnotations` now increments two `uint64` counters on the sanitizer struct (`droppedUnknownAnnotations`, `passedUnknownAnnotations`) plus a per-type-URL breakdown map. No information is lost — only the volume.
- **Per-phase progress logging**: each `copy*` function (`resource_types`, `resources`, `entitlements`, `grants`, `assets`) opens a `phaseProgress` at start, calls `page(n)` per page boundary, and emits a done summary on return. Periodic progress logs are throttled to one per 10-second tick — a multi-hour run emits dozens of lines, not millions.
- **Real N-of-M / percent / ETA when totals are known**: a new optional `syncStatsReader` interface lets the sanitizer ask the source for per-record-type totals before each phase. `*dotc1z.C1File` already implements `Stats()`, which reads the cached `sync_runs` row in O(1). When the source exposes it, progress logs carry `total`, `remaining`, `percent_complete`, `items_per_sec`, and `eta`. When the source doesn't (test fakes), the logger falls back to a running count + rate so an operator can still extrapolate.
- **copyAssets** has the total known up front from `refs.drain()` length, so N-of-M progress is always available there.
- **End-of-run summary**: `Sanitize()` emits one `c1zsanitize: complete` log line with the wall-clock elapsed time, the cumulative dropped / passed counts, and the per-type-URL breakdown.

No change to sanitize semantics; logging only.

## Volume impact

- **Before**: one Debug line per dropped annotation. 57M grants × N annotations each = millions of log lines per run, no progress.
- **After**: 1 start + 1 done per phase (5 phases × 1+ syncs) + 1 throttled progress line per 10 seconds while a phase is running + 1 end-of-run summary. A multi-hour run is dozens of log lines.

## Tests

- `TestSanitizeAggregatesDroppedAnnotations` — locks the regression: zero per-annotation log lines emitted, and the end-of-run summary carries the exact aggregate count.
- `TestSanitizeProgressIncludesTotalsWhenSourceHasStats` — locks that when the source is a `*dotc1z.C1File` the `grants` phase-start log carries a non-zero `total` field, so an operator can answer "how much longer?"
- Existing tests (cross-reference integrity, graph integrity, drop-unknown-by-default, etc.) pass unchanged.

## Test plan

- [x] `go build ./...` — passes
- [x] `go test ./pkg/c1zsanitize/...` — passes including the two new tests
- [x] `go test -short ./pkg/dotc1z/...` — passes (used by c1zsanitize via the optional `Stats()` capability)
- [x] `golangci-lint run ./pkg/c1zsanitize/...` — 0 issues

## Vendor note

`go.uber.org/zap/zaptest/observer` is now vendored — the new tests inspect emitted log lines via the observed-logger pattern. The repo is `-mod=vendor`, so `go mod vendor` was re-run to pull it in.

## Downstream

ductone/c1 vendors this package. A follow-up `go get github.com/conductorone/baton-sdk@<new>` + `go mod vendor` bump in c1 will pick this up after merge.

---

_(satellite) Built with stargate._ 🛰

### DIFF
--- a/pkg/c1zsanitize/annotations.go
+++ b/pkg/c1zsanitize/annotations.go
@@ -38,9 +38,16 @@ func typeURL(m proto.Message) string {
 }
 
 // transformAnnotations walks the slice once, dispatching each entry
-// on its Any type URL. Unknown types are dropped by default (with a
-// log line naming the URL) or passed through unchanged if the
-// operator opted in via Options.AllowUnknownAnnotations=true.
+// on its Any type URL. Unknown types are dropped by default or passed
+// through unchanged if the operator opted in via
+// Options.AllowUnknownAnnotations=true.
+//
+// The dropped / passed counts and a per-type-URL breakdown are
+// accumulated on the sanitizer struct and surfaced in the per-phase
+// progress + end-of-run summary log lines. Per-annotation logging is
+// intentionally absent: a real sanitize run iterates tens of millions
+// of records each carrying multiple annotations, so a one-line-per-
+// annotation log produced millions of entries with no progress signal.
 func (s *sanitizer) transformAnnotations(in []*anypb.Any, refs *assetRefSet) []*anypb.Any {
 	if len(in) == 0 {
 		return nil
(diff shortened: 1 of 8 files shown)


btipling:
Closing per decision — the c1zsanitize annotation-log handling is being folded into #932 (bt/dotc1z-sanitize-bulkload-perf) so the sanitize perf + logging changes ship as one PR. The per-annotation 'dropping unknown annotation' logging is being removed in #932.

PR #932, opened 2026-06-09
perf(c1zsanitize/dotc1z): bulk-load deferred indexes, embedded-transform memoization, graph-annotation handlers

btipling:
## Summary

Dramatic performance work for `c1z` sanitization, plus a correctness fix to the annotation handling. Two layers:

### dotc1z bulk-load write path (`pkg/dotc1z`)
The destination of a sanitize run bulk-loads tens of millions of grants into a freshly-created file. Building all secondary indexes up front means every insert maintains them — and four of the grants indexes are keyed on sanitized (HMAC, i.e. uniformly random) columns, so each insert is a random B-tree page touch whose cost climbs once the table+indexes outgrow the page cache. That is the accelerating, "looks like O(n²)" slowdown.

- New opt-in **bulk-load mode** (`WithBulkLoad` / `WithC1FBulkLoad`), scoped to a net-new single-writer destination:
  - drops the base **non-unique** secondary indexes right after table creation (instant on the empty table), keeping the rowid PK and the `(external_id, sync_id)` unique index the upsert path needs — so the load maintains two B-trees instead of six-plus;
  - rebuilds the deferred indexes in **one pass at `Close`** (before flush) by re-running the canonical `Schema()` DDL, so the final index set is **byte-identical** to the normal path;
  - identifies deferrable indexes generically via `PRAGMA index_list` (`unique=0, origin=c`), captured **before migrations** so partial migration indexes are never disturbed.
- Normal connector syncs don't pass the option and are **unchanged**.
- The sanitize CLI opens its destination with bulk-load + `cache_size=-1048576` (1 GiB), `mmap_size=8 GiB`, `temp_store=MEMORY` (alongside the existing `journal_mode=OFF`/`synchronous=OFF` for the disposable output). All scoped to that writer instance — blast radius is one file.

### c1zsanitize package (`pkg/c1zsanitize`)
- **Correctness — graph annotations preserved, not dropped.** Non-trait grant/entitlement annotations carry the topology that makes a sanitized c1z representative. They are now **preserved-and-sanitized** (fail-closed handlers), not dropped. `AllowUnknownAnnotations` stays `false`. Handler set:
  - `GrantExpandable` — entitlement ids → `transformID`; resource-type tokens follow the known-type rule; `shallow` preserved.
  - `GrantImmutable` / `EntitlementImmutable` — `source_id` → `transformID`; metadata struct string leaves HMAC'd.
  - `ExternalLink` — URL redacted to `https://redacted.example.com`.
  - `ETag` — `entitlement_id` → `transformID`; `value` HMAC'd (fail-closed).
  - `ChildResourceType` — resource-type token follows the known-type rule.
- **Memoize embedded Entitlement/Principal transforms** per sync. Grants are emitted grouped by entitlement, so the cache collapses the expensive per-grant annotation re-derivation to map hits. Cached protos are shared (safe — nothing mutates a message after `.Build()`). Bounded principal cache; off-by-default `proto.Equal` verify guard.
- **Email hot path** through the sanitizer's reused HMAC (not the per-call-keyed reference impl); domain HMAC computed outside the lock.
- **Stop draining asset payloads** that are only discarded (close without read).
- **Skip the timestamp-shift allocation** when the anchor delta is zero.
- **Freeze `knownResourceTypes`** after the resource-types phase so `transformID`'s type-token decision is order-independent (a determinism fix).
- **Permanent per-page `read`/`transform`/`put` timing log** — the regression canary.

## Scoping / blast radius
Bulk-load and the aggressive pragmas are opt-in per writer instance and only wired at the sanitize CLI call site. dotc1z's shared write path for normal connector syncs is untouched.

## Correctness verification
- The deferred-index rebuild re-runs the **same** `Schema()` DDL as the normal path → identical final index set (names, columns, uniqueness).
- New tests: graph-annotation handler coverage (each type preserved + sanitized correctly), annotation determinism, and a **cache-equivalence** test (cached vs uncached transform output identical, shared-pointer reuse on hits). Existing cross-reference / graph-integrity / id-alignment suites still pass.
- `go build ./...`, `go test ./pkg/dotc1z/... ./pkg/c1zsanitize/...`, and `go test -race ./pkg/c1zsanitize/...` all green.

## Benchmark (`BenchmarkTransformGrant`, 5000 grants, G/E=1000)
| | ns/op | allocs/op |
|---|---|---|
| uncached | 29,746,445 | 345,058 |
| cached | 12,105,741 | 144,198 |

~2.46× faster, ~2.4× fewer allocations on the transform stage. The ratio grows with embedded-transform cost at whale scale; full pipeline gains compound with the dotc1z write-path change above.

## Deferred (follow-up)
Per-page parallel transform (fan-out the CPU-bound transform across cores with per-worker HMAC state + sharded caches) is **not** in this PR — it is the highest-risk piece for determinism/races and is additive on top of the memoization here. Best as its own reviewed change. The `knownResourceTypes` freeze in this PR is the prerequisite it needs.

The per-annotation log-dedup (P1-1) is intentionally left to the in-flight log-spam PR to avoid conflicting on `annotations.go`; adding handlers here already removes most unknown-annotation drops for the graph types.

---

_(satellite) Built with stargate._ 🛰

### DIFF
--- a/cmd/baton/sanitize.go
+++ b/cmd/baton/sanitize.go
@@ -113,6 +113,10 @@ func runSanitize(cmd *cobra.Command, args []string) error {
 		dotc1z.WithPragma("mmap_size", "8589934592"),
 		dotc1z.WithPragma("temp_store", "MEMORY"),
 		dotc1z.WithBulkLoad(true),
+		// bulkLoad already implies skip-cleanup; skip VACUUM too — vacuuming
+		// before the deferred indexes are rebuilt at Close is wasted work on a
+		// throwaway artifact.
+		dotc1z.WithSkipVacuum(true),
 	)
 	if err != nil {
 		return fmt.Errorf("open dst c1z: %w", err)
--- a/cmd/baton/sanitize.go
+++ b/cmd/baton/sanitize.go
@@ -93,14 +93,26 @@ func runSanitize(cmd *cobra.Command, args []string) error {
 	}
 	defer src.Close(ctx)
 
-	// The dst is a net-new intermediate that is discarded on any
-	// failure, so durability pragmas only cost throughput: skip the
-	// journal and fsync entirely, and give SQLite a 64MB page cache to
-	// cut index-maintenance misses on large (multi-million-grant) syncs.
+	// The dst is a net-new, single-writer intermediate that is discarded
+	// on any failure, so durability and index-maintenance costs are pure
+	// throughput overhead here:
+	//   - journal_mode=OFF + synchronous=OFF: no rollback journal, no fsync
+	//     (the output is rebuilt from source on any failure, not recovered).
+	//   - cache_size=-1048576 (1 GiB) + mmap_size=8 GiB + temp_store=MEMORY:
+	//     keep index pages resident and run the deferred index build in
+	//     memory, cutting page-cache misses on large (multi-million-grant)
+	//     syncs.
+	//   - WithBulkLoad: defer secondary-index creation until after the load
+	//     so per-row random-key B-tree maintenance does not dominate.
+	// These pragmas are scoped to THIS writer instance; normal connector
+	// syncs open their own C1File and are unaffected.
 	dst, err := dotc1z.NewC1ZFile(ctx, outPath,
 		dotc1z.WithPragma("journal_mode", "OFF"),
 		dotc1z.WithPragma("synchronous", "OFF"),
-		dotc1z.WithPragma("cache_size", "-65536"),
+		dotc1z.WithPragma("cache_size", "-1048576"),
+		dotc1z.WithPragma("mmap_size", "8589934592"),
+		dotc1z.WithPragma("temp_store", "MEMORY"),
+		dotc1z.WithBulkLoad(true),
 	)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


cmd/baton/sanitize.go
```diff
@@ -93,14 +93,30 @@ func runSanitize(cmd *cobra.Command, args []string) error {
...
 		dotc1z.WithPragma("synchronous", "OFF"),
-		dotc1z.WithPragma("cache_size", "-65536"),
+		dotc1z.WithPragma("cache_size", "-1048576"),
+		dotc1z.WithPragma("mmap_size", "8589934592"),
+		dotc1z.WithPragma("temp_store", "MEMORY"),
+		dotc1z.WithBulkLoad(true),
+		// bulkLoad already implies skip-cleanup; skip VACUUM too — vacuuming
+		// before the deferred indexes are rebuilt at Close is wasted work on a
+		// throwaway artifact.
+		dotc1z.WithSkipVacuum(true),
```
btipling, line 119: This change doesn't really matter since we don't use binary

kans (approved):

PR #936, opened 2026-06-09
perf(c1zsanitize): parallel per-page transform + resumable sanitization

btipling:
## Summary

Two changes to `pkg/c1zsanitize`, building on the memoization + bulk-load work in v0.12.7:

1. **Parallelize the per-page transform stage.** Each fetched page's record transforms fan out across `GOMAXPROCS` workers; reads and writes stay sequential (SQLite is a single writer). Only the CPU-bound HMAC + proto marshal/unmarshal work parallelizes.
2. **Resumable sanitization (checkpointing).** An interrupted run resumes from its last committed position instead of restarting.

## Parallel transform

- Each page allocates a pre-sized output slice; workers pull indices off a shared atomic counter and write `out[i]`, so order is preserved with no channels. After the workers join, the single `Put*` runs as before.
- **Per-worker HMAC:** `s.id` now borrows a pre-keyed HMAC-SHA256 hasher from a `sync.Pool` rather than sharing one `hash.Hash`. Each call Resets a hasher it exclusively owns, so output depends only on `(secret, input)` — never on scheduling.
- **Concurrency-safe caches:** the grant sub-cache gets a mutex with a double-checked insert (the expensive transform on a miss runs *outside* the lock — entitlement cardinality is orders of magnitude below grant count, so the map is read-mostly after warmup and a plain `Mutex` suffices). The per-run drop counters and the undeclared-type dedup set are guarded by a stats mutex (tiny critical sections). The domain map and asset-ref set were already mutex-guarded. `knownResourceTypes` is populated only in the resource-types pre-pass and read-only afterward, so the concurrent phases read it lock-free.

## C-2 fix (behavior fix, not just a parallelism prerequisite)

Resource types are registered **only** during their own buffering pre-pass, and the known-type set is read-only for the rest of the sync. Previously an embedded resource type (e.g. in a grant's `GrantableTo`) could become "known" mid-run, so the id transform would HMAC that token before the encounter and preserve it after — the same token, two representations, dependent on row order, producing order-dependent output and broken cross-references. Freezing the set after the pre-pass makes the transform's known-type decision deterministic. This is strictly a correctness improvement (it also removes a data race the fan-out would otherwise expose).

## Checkpoint design

- **Storage:** each destination sync's existing `sync_token` (via `CheckpointSync` / `CurrentSyncStep`). It lives inside the destination c1z — no sidecar file, no schema change — and is the same mechanism connector syncs already use for resumption. It is persisted with the rest of the c1z, so resume only finds progress a prior run actually committed to disk.
- **Granularity:** per source sync, per phase, and **per page within the grant phase**. The token records `{fingerprint, source sync id, next phase, grant page token}`. `resource_types` always re-runs (cheap, and it repopulates the known-type set the id transform consults); `resources` and `entitlements` are skipped once complete; the grant phase resumes from its last committed page (keyset pagination on the source rowid makes that token stable across runs).
- **Resume semantics:** on a `Resumable` start, existing destination syncs are scanned; a token whose fingerprint matches this run yields a resume point. A fully-`EndSync`'d source sync is skipped entirely (its dst id is kept for parent linkage); an unfinished one is re-entered via `SetCurrentSync` and continued from its phase/page.
- **Mismatch handling:** the fingerprint is `HMAC-SHA256(secret, "…" + anchor)` (the secret is never stored). A destination carrying a checkpoint whose fingerprint does not match the current run's secret/anchor is **rejected with an error** rather than appended to — fail-closed, never mixing transforms from two secrets. A destination with no tokens (fresh, or written by a non-resumable run) yields a clean full run.
- **Idempotence / cleanup:** all `Put*` calls upsert, so re-writing a boundary page or the always-rerun `resource_types` phase is idempotent and the resumed output is record-identical to an uninterrupted run. Completed source syncs are `EndSync`'d; their tokens mark them done, so a resumable re-run against a finished destination skips rather than duplicating. For a genuinely clean run, use a fresh destination path.
- **Lifecycle note (documented tradeoff):** `Sanitize` receives a `connectorstore.Writer` and does not own the destination file's open/close, so it cannot itself force a mid-run flush. Resume therefore recovers whatever the caller persisted (the caller closes/saves the c1z at its own cadence). This is why the granularity is "last committed page of the persisted destination," not "every page regardless of persistence."

## Guardrails preserved

Determinism (same `(secret, input)` → same output regardless of scheduling/cache state); fail-closed `transformID`; `AllowUnknownAnnotations` default false; public API signatures unchanged (`Sanitize`, `Options`, `SanitizeID`, `MinSecretBytes`, `LoadOrGenerateSecret` — `Options` only gains the additive `Resumable` field); no new dependencies.

## Tests

- Determinism golden + cache-equivalence (existing) still pass.
- `TestSanitizeParallelMultiSync` — multi-sync fan-out, run under `-race`.
- `TestSanitizeCheckpointResume` — interrupt a resumable run at the grant phase, persist, resume; asserts the resumed output is record-identical to an uninterrupted run with the same secret/anchor.
- `BenchmarkSanitize` — end-to-end over a synthetic c1z (the tracked headline is `G=1M`; the committed `genGrants` is modest so the benchmark is CI-runnable).

`go build ./...`, `go test ./pkg/c1zsanitize/... -race -count=1`, `go test ./pkg/dotc1z/ -count=1`, `gofmt`, and `make lint` are all green on the touched files.

---

_🛰️ Built with stargate._

### DIFF
--- a/pkg/c1zsanitize/annotations.go
+++ b/pkg/c1zsanitize/annotations.go
@@ -72,28 +72,25 @@ func (s *sanitizer) transformAnnotations(in []*anypb.Any, refs *assetRefSet) []*
 			// could leak un-sanitized customer data. Pass-through is opt-in
 			// via Options.AllowUnknownAnnotations, for development only.
 			if s.dropUnknownAnnotations {
-				if s.droppedAnnotations[a.GetTypeUrl()] == 0 {
+				if s.recordAnnotation(s.droppedAnnotations, a.GetTypeUrl()) {
 					s.log.Warn("c1zsanitize: dropping unknown annotation type (first occurrence; counted thereafter)",
 						zap.String("type_url", a.GetTypeUrl()))
 				}
-				s.droppedAnnotations[a.GetTypeUrl()]++
 				continue
 			}
-			if s.passedAnnotations[a.GetTypeUrl()] == 0 {
+			if s.recordAnnotation(s.passedAnnotations, a.GetTypeUrl()) {
 				s.log.Warn("c1zsanitize: passing unknown annotation through unchanged (first occurrence; counted thereafter)",
 					zap.String("type_url", a.GetTypeUrl()))
 			}
-			s.passedAnnotations[a.GetTypeUrl()]++
 			out = append(out, a)
 			continue
 		}
 		msg, err := a.UnmarshalNew()
 		if err != nil {
-			if s.failedAnnotations[a.GetTypeUrl()] == 0 {
+			if s.recordAnnotation(s.failedAnnotations, a.GetTypeUrl()) {
 				s.log.Warn("c1zsanitize: annotation transform failed (first occurrence; counted thereafter)",
 					zap.String("type_url", a.GetTypeUrl()), zap.Error(err))
 			}
-			s.failedAnnotations[a.GetTypeUrl()]++
 			continue
 		}
 		sanitized := handler(s, msg, refs)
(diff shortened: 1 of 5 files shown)


btipling:
Superseded by #938 (combined c1zsanitize: parallel + resumable + expansion-topology preservation, with the resumable-half blockers fixed). Branch retained.

PR #937, opened 2026-06-10
feat(dotc1z): preserve grant-expansion topology when streaming grants

btipling:
## Problem

The SQLite engine strips a grant's `GrantExpandable` annotation into a side `expansion` column on write (`extractAndStripExpansion`). `C1File.StreamGrants` reads only the `data` blob, so it yields grants with their **expansion topology missing**. The Pebble engine reconstructs the annotation from its record (`V3GrantToV2` → `expansionRecordToV2`), so the two engines disagree — and the SQLite→Pebble engine copy (`to_pebble` `copyGrants`) silently drops the expansion edges.

## Change

Add an opt-in `StreamGrantsOptions.IncludeExpansion` (default `false`, so the existing data-only stream is byte-for-byte unchanged). When set, the SQLite stream selects the `expansion` column alongside `data` and re-attaches the deserialized `GrantExpandable` annotation onto each yielded grant, restoring the exact expansion edges. The cross-engine copy opts in so copied grants keep their topology; the Pebble stream already includes expansion, so the option is a no-op there. A debug line reports per-stream `grants` / `expansion_reattached` counts.

## Steering mechanism — why `StreamGrantsOptions`

Of the three options:
- **`StreamGrantsOptions` field (chosen)** — purely additive, opt-in, default-off, changes no existing behavior, and `StreamGrants`/the Pebble adapter already branch on `opts` fields. Least-invasive.
- **interface-widening** (a new `StreamGrantsWithExpansion` method on `StreamingReader`) — forces every implementer (SQLite `C1File`, Pebble `Adapter`) to add a method; more surface for no benefit.
- **request-annotation** — the unary `ListGrants` path uses request annotations, but `StreamGrants` is not request-driven, so it does not fit.

## Observability

`streamGrantsWithExpansion` logs once per completed scan at debug: `grants` (total streamed) and `expansion_reattached` (count that carried a stored `GrantExpandable`).

## Tests

`TestStreamGrantsIncludeExpansionRoundTrip` (rollback/replay round-trip guard): write a grant carrying a `GrantExpandable` (which is stripped into the side column), then stream it back — `IncludeExpansion:false` yields it **without** the annotation (the documented default), and `IncludeExpansion:true` restores the **exact** `GrantExpandable` (entitlement ids, shallow, resource-type ids). A non-expandable grant is unchanged either way.

`go build ./...`, `go vet`, `gofmt`, and `go test ./pkg/dotc1z/ ./pkg/connectorstore/...` are green.

## go.mod pin-delta note (for the be-temporal-sync reviewer)

This is a baton-sdk-only change. A grant consumer such as **be-temporal-sync** that wants the expansion topology preserved must (1) bump its `github.com/conductorone/baton-sdk` pin to the release containing this commit (`5297629`), and (2) pass `connectorstore.StreamGrantsOptions{IncludeExpansion: true}` at its `StreamGrants` call site. No be-temporal-sync code is changed here; the default-off option means existing consumers are unaffected until they opt in.

---

_🛰️ Built with stargate._

### DIFF
--- a/pkg/connectorstore/streaming.go
+++ b/pkg/connectorstore/streaming.go
@@ -38,10 +38,19 @@ type StreamingReader interface {
 
 // StreamGrantsOptions narrows the stream to a single entitlement
 // and/or principal. Empty fields mean no filter.
+//
+// IncludeExpansion re-attaches each grant's GrantExpandable annotation when
+// set. The SQLite engine strips GrantExpandable into a side column on write,
+// so without this the SQLite stream yields grants with their expansion
+// topology missing — a faithful round-trip (cross-engine copy, rollback /
+// replay, or an external grant consumer) must set it. Default false preserves
+// the existing data-only stream. The Pebble engine reconstructs the
+// annotation from its record either way, so the option is a no-op there.
 type StreamGrantsOptions struct {
 	EntitlementID         string
 	PrincipalResourceType string
 	PrincipalResourceID   string
+	IncludeExpansion      bool
 }
 
 // StreamResourcesOptions narrows the stream by resource_type.
--- a/pkg/dotc1z/streaming.go
+++ b/pkg/dotc1z/streaming.go
@@ -6,9 +6,12 @@ import (
 	"iter"
 
 	"github.com/doug-martin/goqu/v9"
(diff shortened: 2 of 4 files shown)


btipling:
Superseded by #938 (combined c1zsanitize PR). The grant-expansion-topology work is carried into #938. Branch retained.

PR #946, opened 2026-06-11
feat(dotc1z): add SnapshotTo to materialize a partial .c1z from a live handle

btipling:
## What

Adds `(*C1File).SnapshotTo(ctx, outPath, opts...)` to `pkg/dotc1z`: a generic dotc1z method that writes a valid, openable `.c1z` of the destination's **current committed on-disk state** to `outPath` **without closing the live handle** and **without re-decompressing** the source. The live handle remains fully usable for reads and writes afterward.

Unlike `CloneSync`, which copies exactly one *ended* sync filtered by `sync_id`, `SnapshotTo`:

- copies **every** `sync_runs` row — ended **and** un-ended alike — and every row of every object table (resources, entitlements, grants, assets, ...) wholesale; and
- preserves each row's `sync_token` bytes **exactly** (a plain table copy, no parse/re-encode).

That byte-for-byte `sync_token` preservation is the load-bearing property: a consumer that checkpoints resume progress into `sync_token` (e.g. `c1zsanitize` with `Options.Resumable`) can reopen the snapshot with `NewC1ZFile` and resume from the checkpoint that was current when the snapshot was taken.

Also adds `c1zsanitize.ResumePhase(syncToken)` — a read-only, no-I/O decoder that returns the resume phase, grant page token, and source sync id from a destination sync's raw `sync_token`, so a consumer can read resume state without coupling to the unexported checkpoint JSON shape.

## Why

A consumer runs `c1zsanitize.Sanitize` over whale-scale `.c1z` files inside a long-running activity. Resumable sanitize already checkpoints progress into each destination sync's `sync_token`, but there was no way to materialize a valid partial `.c1z` from a **live, mid-run** destination handle: `NewC1ZFile` decompresses into a hidden temp sqlite db, and the `.c1z` at the caller's path is only written by `Close`. So a partial destination could not be staged without closing the file, and reopening a whale `.c1z` pays a long single-threaded decompress. `SnapshotTo` closes that gap: stage a resumable partial at checkpoint time, then continue (or resume later) with no full restart.

This is an additive, backward-compatible change — a new method and a new exported function, no signature changes.

## How

- Refactors `CloneSync`'s body into a shared private `cloneCopy(ctx, outPath, syncID, selectAll, opts...)` helper: schema-init → `ATTACH` → per-table `INSERT...SELECT` → `DETACH` → compress. `selectAll=false` copies `WHERE sync_id=?` (CloneSync); `selectAll=true` copies all rows including the full `sync_runs` table (SnapshotTo).
- `CloneSync` becomes a thin wrapper that resolves the sync id, enforces its existing `NotFound` / `FailedPrecondition` (not-ended) preconditions, then calls `cloneCopy(..., selectAll=false)`. Its exported signature, godoc, preconditions, and observable behavior are unchanged.
- `SnapshotTo` is the other wrapper: no sync resolution, no not-ended precondition, prepends the same default opts (`WithC1FEncoderConcurrency(0)`, `WithC1FSkipCleanup(true)`), and calls `cloneCopy(..., selectAll=true)`. It rejects the Pebble engine and read-only handles with a clear error.

### Concurrency and atomicity

The row copy runs over the file's single SQLite connection (`SetMaxOpenConns(1)`), so it serializes against the live writer via that one connection rather than racing it — the SELECTs read committed rows on the writer's own connection. The compress half runs on a separate temp db and does **not** hold the live connection, so only the row-copy window blocks the writer; there is no live-sync cancellation and no phase redo. The write is atomic: a `.tmp` file is built and renamed into place only on full success, so a failed snapshot leaves no partial file at `outPath` (same guarantee as `Close`/`saveC1z`). `cloneCopy` never touches the live `C1File`'s db handle or sync state, so the handle is left exactly as it was found.

For whale files the godoc recommends callers also pass `WithC1FBulkLoad(true)` + `WithC1FSkipVacuum(true)` so the snapshot-side write defers secondary-index maintenance to a single rebuild pass.

## Files changed

- `pkg/dotc1z/clone_sync.go` — `cloneCopy` extraction, `cloneTableQueryAll`, `SnapshotTo`; `CloneSync` reduced to a wrapper (behavior unchanged).
- `pkg/c1zsanitize/resume_phase.go` — new `ResumePhase` decoder.
- `pkg/dotc1z/snapshot_test.go` — un-ended-sync inclusion, atomicity-on-failure, live-handle-still-writable, engine/read-only/outPath guards, and a CloneSync not-ended-rejection regression pin.
- `pkg/c1zsanitize/snapshot_resume_test.go` — headline resume-from-snapshot record-identity (resources/entitlements boundary and mid-grants with a non-empty grant page token) and `sync_token` byte-for-byte preservation.

## Tests

`gofmt` clean; `go build ./...` clean; `golangci-lint` clean for the changed files. `go test ./pkg/dotc1z/... ./pkg/c1zsanitize/...` is green:

```
ok  github.com/conductorone/baton-sdk/pkg/dotc1z                          40.3s
ok  github.com/conductorone/baton-sdk/pkg/dotc1z/c1zstore                  0.005s
ok  github.com/conductorone/baton-sdk/pkg/dotc1z/engine/equivalence        0.066s
ok  github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble             0.598s
ok  github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble/codec       0.009s
ok  github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble/microtests  0.014s
ok  github.com/conductorone/baton-sdk/pkg/dotc1z/format/v3                 0.049s
ok  github.com/conductorone/baton-sdk/pkg/c1zsanitize                      2.746s
```

The existing `clone_sync_test.go` passes unchanged (the shared-helper refactor's regression gate).

---

_(🛰) Built with stargate._

### DIFF
--- a/pkg/dotc1z/clone_sync.go
+++ b/pkg/dotc1z/clone_sync.go
@@ -99,6 +99,13 @@ func cloneTableQueryAll(tableName string, columns []string) string {
 	)
 }
 
+// errOutPathExists builds the single canonical error returned when a clone or
+// snapshot output path is already occupied. errPrefix namespaces it to the
+// caller ("clone-sync" / "snapshot-to") while keeping one message shape.
+func errOutPathExists(errPrefix, outPath string) error {
+	return fmt.Errorf("%s: output path (%s) must not exist", errPrefix, outPath)
+}
+
 // CloneSync uses sqlite hackery to directly copy the pertinent rows into a new database.
 // 1. Create a new empty sqlite database in a temp file
 // 2. Open the c1z that we are cloning to get a db handle
@@ -116,11 +123,11 @@ func (c *C1File) CloneSync(ctx context.Context, outPath string, syncID string, o
 	}
 	opts = append(defaultOpts, opts...)
 
-	// Be sure that the output path is empty else return an error. Checked here,
-	// ahead of the sync preconditions, so an occupied output path is reported
-	// before the target sync is inspected — the order callers have always seen.
+	// Check the output path ahead of the sync preconditions so an occupied
+	// output path is reported before the target sync is inspected — the order
+	// callers have always seen. cloneCopy re-checks it as the shared chokepoint.
 	if _, statErr := os.Stat(outPath); statErr == nil || !errors.Is(statErr, fs.ErrNotExist) {
-		return fmt.Errorf("clone-sync: output path (%s) must not exist for cloning to proceed", outPath)
+		return errOutPathExists("clone-sync", outPath)
 	}
 
 	if syncID == "" {
@@ -224,15 +231,20 @@ func (c *C1File) SnapshotTo(ctx context.Context, outPath string, opts ...C1FOpti
 // The compress runs on the separate temp db and does not hold the live
 // connection. cloneCopy reads c.rawDb to take that one connection but never
 // mutates c.rawDb/c.db/c.currentSyncID/c.closed, so the live handle is left
-// exactly as it was found (plus the row-copy stall). Callers guard a closed
-// handle before reaching here, so c.rawDb is non-nil.
+// exactly as it was found (plus the row-copy stall). Both entry points already
+// reject a closed handle; the nil-rawDb guard below is defensive against a
+// future caller that does not.
 // errPrefix is the caller's error-message namespace ("clone-sync" /
 // "snapshot-to") so each entry point keeps its own observable error strings.
 func (c *C1File) cloneCopy(ctx context.Context, outPath string, syncID string, selectAll bool, errPrefix string, opts ...C1FOption) error {
+	if c.rawDb == nil {
+		return ErrDbNotOpen
+	}
+
 	// Be sure that the output path is empty else return an error
 	_, err := os.Stat(outPath)
 	if err == nil || !errors.Is(err, fs.ErrNotExist) {
-		return fmt.Errorf("%s: output path (%s) must not exist", errPrefix, outPath)
+		return errOutPathExists(errPrefix, outPath)
 	}
 
 	tmpDir, err := os.MkdirTemp(c.tempDir, "c1zclone")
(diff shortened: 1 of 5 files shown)


pkg/dotc1z/clone_sync.go
```diff
@@ -100,10 +123,128 @@ func (c *C1File) CloneSync(ctx context.Context, outPath string, syncID string, o
...
+func (c *C1File) SnapshotTo(ctx context.Context, outPath string, opts ...C1FOption) error {
+	ctx, span := tracer.Start(ctx, "C1File.SnapshotTo")
+	var err error
+	defer func() { uotel.EndSpanWithError(span, err) }()
+
+	if err = c.validateDb(ctx); err != nil {
+		return err
+	}
+
+	if c.engine == EnginePebble {
```
btipling, line 198: This is a follow up

pkg/dotc1z/clone_sync.go
```diff
@@ -136,66 +282,77 @@ func (c *C1File) CloneSync(ctx context.Context, outPath string, syncID string, o
 	initFile.rawDb = nil
 	initFile.db = nil
 
-	if syncID == "" {
```
btipling, line 139: This block looks removed but it's still above on line 133 in a wrapper that wraps cloneCopy

pkg/dotc1z/clone_sync.go
```diff
@@ -83,6 +83,29 @@ func cloneTableQuery(tableName string, columns []string) string {
 	)
 }
 
+// cloneTableQueryAll is the all-rows sibling of cloneTableQuery: it omits the
```
btipling, line 86: The changes in this file look hard, but it adds a wrapper, existing behavior is preserved

kans (approved):

PR #947, opened 2026-06-11
Add storage-engine option to specify which storage engine to use.

ggreer:


pkg/connectorrunner/runner.go
```diff
@@ -665,6 +667,13 @@ func WithWorkerCount(workerCount int) Option {
 	}
 }
 
+func WithStorageEngine(engine dotc1z.Engine) Option {
+	return func(ctx context.Context, cfg *runnerConfig) error {
+		cfg.storageEngine = engine
```
kans, line 672: What happens if you open a c1z with the other format?

kans (approved):

pkg/connectorrunner/runner.go
ggreer: If it's sqlite and the engine is pebble, it gets converted to pebble and we keep on truckin'. If it's pebble and the engine is sqlite, we error.

PR #953, opened 2026-06-12
Real interface for list grants by principal. 

mj-palanker:
Exposing the by principal listing will enable performant new user grant uplift.

### DIFF
--- a/pb/c1/reader/v2/grant.pb.go
+++ b/pb/c1/reader/v2/grant.pb.go
@@ -752,6 +752,214 @@ func (b0 GrantsReaderServiceListGrantsForEntitlementsResponse_builder) Build() *
 	return m0
 }
 
+type GrantsReaderServiceListGrantsForPrincipalRequest struct {
+	state       protoimpl.MessageState `protogen:"hybrid.v1"`
+	PrincipalId *v2.ResourceId         `protobuf:"bytes,1,opt,name=principal_id,json=principalId,proto3" json:"principal_id,omitempty"`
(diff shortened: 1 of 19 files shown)


kans:
From Fable:

Finding (confirmed): The Pebble primary grant key is exactly (entitlement RT, entitlement resource, flag, tail, principal RT, principal ID) (keys.go:109-121), so entitlement + principal fully identifies one grant. PaginateGrantsByEntitlementPrincipal (paginate.go:327-350) does a single db.Get via getGrantByIdentity — a true O(1) point lookup returning at most one record with no cursor. ListGrantsForEntitlement already routes to it when principal_id is set (adapter_reader.go:245-247).

The branch's gap (confirmed): On the branch, Adapter.ListGrantsForPrincipal always walks the by_principal index page-by-page and drops non-matching entitlements with continue after fetching, while the next-page token comes from the unfiltered scan — so a principal with many grants on other entitlements yields empty pages with a token, and O(K) work for what should be one read.

Fix: In Adapter.ListGrantsForPrincipal, when req.Entitlement is set, resolve it via entitlementIdentityForRequest and return PaginateGrantsByEntitlementPrincipal directly; keep the by_principal scan for the principal-only case. ~15 lines, no new index, no proto change.

mj-palanker:
Fixed the above issue in [3f047e8](https://github.com/ConductorOne/baton-sdk/pull/953/commits/3f047e84f044f5f9ac7387d5bd0ade81f28cc321)

kans (approved):

PR #959, opened 2026-06-12
Don't insert grants with nil principals.

ggreer:


ggreer:
Actually I don't think this is needed because we have validation in the proto, so connectors will always return grants with principals. That means the only way to insert nil principals is if baton-sdk messes something up.

PR #962, opened 2026-06-15
cmd/baton: remove Hidden flag from rollback-expansion command

btipling:
The `rollback-expansion` subcommand was registered with `Hidden: true` in
`cmd/baton/rollback_expansion.go`, which suppressed it from appearing in
`baton --help` output. The command is compiled in and fully functional;
hiding it makes it undiscoverable to users.

This change removes the `Hidden: true` field so the command appears in
the standard help listing alongside other subcommands.

---

Built with stargate.


ggreer (approved):
Maybe in the help, mention that this should only be used if you know what you're doing. Is there any use case where a customer would want to undo grant expansion on a c1z?

kans (approved):

PR #963, opened 2026-06-15
make expansion go fasterer

kans:
image


ggreer:
Closing because #965 gives us more speedup.

PR #965, opened 2026-06-16
 Kans/provenance semirings 20x++ grant expansion speed

kans:
## Summary
This PR adds a projection-backed topological grant expansion path for Pebble c1z files

The new path changes expansion from repeated source-page fanout / destination probing into a destination-first topological reduce. It builds a temporary Pebble projection DB for planned source entitlement streams, ingests the projection via SST, and then merges parent contribution streams by principal before writing dirty grants once per destination entitlement.  This is probably at least 25x faster than main, possibly double that or more.

image

Key pieces:

- Adds EntitlementGraphPlan, stored on EntitlementGraph, with per-node strategy metadata and selective ProjectionSources.
- Adds RunTopologicalMergeProjection, plus supporting topological/streaming merge implementations.
- Adds benchmark-only dense graph harnesses for current, topological, projection, and SST shuffle variants.
- Adds projection-specific benchmark metrics:
    - projection rows built
    - nodes reduced
    - destination entitlements reduced
    - dirty grants written

Why
Large high-fan-in graphs make the existing expansion loop repeatedly probe and rewrite the same destination grant sets. 

The main improvement is algorithmic: build compact source projections once, then reuse sorted source streams across destination reductions instead of repeatedly hydrating source grants and probing destination principals.


pkg/sync/expand/topological_merge_projection.go
```diff
@@ -0,0 +1,522 @@
...
+	}
+	return s.iter.Close()
+}
+
+// decodeProjectionPrincipal extracts the principal (resource_type, resource)
+// from a projection key without allocating. The returned slices alias key and
+// are only valid until the iterator that produced key advances. ok mirrors the
+// validation of the encoder's rt\0resource\0externalID layout: two separators
+// and a non-empty external id.
+func decodeProjectionPrincipal(key, prefix []byte) (rt, resource []byte, ok bool) {
```
btipling: **[bug]** New code added by the PR violates the project's golangci-lint config (nonamedreturns linter). make lint fails with: "named return "rt" with type "[]byte" found (nonamedreturns)" on decodeProjectionPrincipal. Prior snapshot passed lint; this head does not.

**Suggestion:** Rename the return or disable per the linter rule for this helper (or rewrite to avoid named returns). Re-run make lint as part of any change touching new expand code.

pkg/sync/expand/expansion_plan.go
```diff
@@ -0,0 +1,187 @@
...
+import (
+	"context"
+	"sort"
+)
+
+const expansionPlanVersion = 1
+
+// ExpansionPlanStrategy names the expansion algorithm selected for a node.
+// The current default expander does not yet execute mixed strategies; this
+// plan is the persisted decision artifact that a hybrid executor will consume.
```
btipling: **[suggestion]** Comment still states "this plan is the persisted decision artifact that a hybrid executor will consume." BuildExpansionPlan / ensureExpansionPlan only ever assign to the in-memory EntitlementGraph.ExpansionPlan (with json tags for debug?); there is no Write/Checkpoint/protobuf/c1z persistence in this change (or prior). The plan is recomputed on demand inside topo Run* and discarded after the step.

**Suggestion:** Update the comment to describe the actual current role (in-memory shape-driven policy attached to the graph for the current topo step). Remove or qualify "persisted". If future persistence is planned, use a TODO with ticket rather than stating as accomplished fact.

pkg/sync/expand/topological_merge_streaming.go
```diff
@@ -0,0 +1,473 @@
...
+// key. Every input stream MUST yield groups in strictly increasing principal
+// key order (see principalGroupStream / projectionContributionStream); the heap
+// surfaces the minimum key and collects all streams positioned at it.
+func mergeContributionGroupStreams(
+	ctx context.Context,
+	destEntitlement *v2.Entitlement,
+	streams []contributionGroupStream,
+) ([]*v2.Grant, error) {
+	for _, stream := range streams {
+		defer stream.close()
```
btipling, line 311: **[nit]** defer stream.close() is written inside a range loop over the streams slice inside mergeContributionGroupStreams. (Classic defer-in-range pattern, even if 1.22+ iteration semantics make the var capture safe here.)

**Suggestion:** Hoist with explicit capture (for i := range streams { s := streams[i]; defer s.close() }) or a helper that closes the slice at the end of the func (after the early error returns). Same pattern noted on ancestor.

pkg/sync/expand/expander.go
```diff
@@ -109,6 +125,28 @@ func (e *Expander) Run(ctx context.Context) error {
 // Returns true when the graph is fully expanded, false if more work is needed.
 // This matches the syncer's step-by-step execution model.
 func (e *Expander) RunSingleStep(ctx context.Context) error {
+	switch os.Getenv("BATON_GRAPH_EXPAND_ALGORITHM") {
```
btipling: **[suggestion]** The dispatch on os.Getenv("BATON_GRAPH_EXPAND_ALGORITHM") hardcodes the string literals ("topological_projection", "projection", "topological_streaming", "streaming", "topological") in RunSingleStep (and duplicated in the added benchmark files and test helpers). No central consts, documented matrix, or single source of truth for supported values vs. benchmark-only.

**Suggestion:** Introduce unexported consts (e.g. algoProjection = "topological_projection") or a small map/dispatcher table. Document which values are production-intended vs. debug/bench. Call sites that read the env can share the table.

pkg/sync/expand/topological_merge.go
```diff
@@ -0,0 +1,506 @@
...
+type topoContribution struct {
+	sources        map[string]bool
+	principal      *v2.Resource
+	principalBytes []byte
+}
+
+func (c *topoContribution) add(sourceEntitlementID string, isDirect bool, principal *v2.Resource) {
+	c.addSource(sourceEntitlementID, isDirect)
+	if c.principal == nil && principal != nil {
+		c.principal = proto.Clone(principal).(*v2.Resource)
```
btipling, line 433: **[nit]** Unchecked proto.Clone casts remain in the new topo code: c.principal = proto.Clone(principal).(*v2.Resource), byID[id] = proto.Clone(grant).(*v2.Grant) (404), grant := proto.Clone(baseGrant).(*v2.Grant) (446). A store returning an unexpected concrete type (or future proto change) will panic rather than a typed error. Same sites existed on ancestor; now on hot topo paths.

**Suggestion:** Introduce a small safeClone helper in the expand package (or reuse existing) that does the , ok := ... check and returns error. Apply at these three sites (and audit pre-existing clones in the same file if they are now on expansion hot paths).

pkg/sync/expand/topological_merge_projection.go
```diff
@@ -0,0 +1,522 @@
...
+		return nil
+	}
+
+	tempDir, err := os.MkdirTemp("", "baton-expand-projection-*")
+	if err != nil {
+		return fmt.Errorf("topological projection: temp dir: %w", err)
+	}
+	defer os.RemoveAll(tempDir)
+
+	projDB, err := pebble.Open(filepath.Join(tempDir, "db"), &pebble.Options{})
```
btipling, line 48: **[suggestion]** projDB is opened with &pebble.Options{} (engine defaults, no memtable/WAL/cache tuning) inside a plain os.MkdirTemp("baton-expand-projection-*"). buildProjectionDB does full materialization of projectionSources into kvs then one Ingest; later addProjectionRows use NoSync batches. No explicit Flush between phases beyond what Ingest provides, no size accounting or budget, and defer projDB.Close() + RemoveAll is the only cleanup (correct on happy path but the temp pebble is a whole second engine instance with its own invariants).

**Suggestion:** Document the memory envelope (same order as the chosen projectionSources grants). Consider explicit Options (DisableWAL: true, etc.) or a bounded cache. Add a comment confirming Ingest + later batch visibility to iters on the same handle. Ensure iterator closes on all error paths inside build/add (current defers help). Consider whether the temp DB abstraction could be narrower than a full *pebble.DB.

pkg/sync/expand/topological_merge_projection.go
```diff
@@ -0,0 +1,522 @@
...
+	if err != nil {
+		return time.Time{}, false, fmt.Errorf("topological projection: parse BATON_BENCH_BUDGET: %w", err)
+	}
+	if d <= 0 {
+		return time.Time{}, false, nil
+	}
+	return time.Now().Add(d), true, nil
+}
+
+func projectionKey(entitlementID string, principal topoPrincipalKey, externalID string) []byte {
```
btipling, line 341: **[nit]** projectionKey / projectionPrefix / decodeProjectionPrincipal / projectionUpperBound / encodeProjectionKV use a hand-rolled 0-byte-separated layout and manual byte scanning (bytes.IndexByte on tails) instead of the shared codec.AppendTuple* or the main grant index builders in pebble/keys.go. Duplication of separator/scan logic, even if the temp DB lifetime is isolated.

**Suggestion:** Add an explicit comment at the top of the key helpers: "custom 0-separated encoding for the transient projection DB only; must stay in sync with decode/upperBound and the call sites in build/add". If more temp structures appear, extract a tiny shared helper; for now the comment + tests (topological_merge_sqlite_parity_test etc.) are the guard.

pkg/sync/expand/expander.go
btipling: **[suggestion]** BuildExpansionPlan / ensureExpansionPlan / the EntitlementGraphPlan (with per-node strategies, ProjectionSources, NodeIDsByStrategy) are reached only when a topological* algorithm is selected via env var (RunTopological* call ensure; the default/legacy RunSingleStep path and production absent the env never build or consult the plan). The plan types therefore have test/bench-only reach today, and the computed per-node strategies are not used to dispatch mixed modes inside the current Run* impls (they run full streaming or full projection for the step).

**Suggestion:** Either (a) call ensure unconditionally in RunSingleStep (it is cheap) and simply ignore the plan for the legacy batch path, or (b) clearly mark the plan construction + strategy types as "topological debug / future hybrid only". Remove dead code paths or add build tags/docs. (See chooseExpansionPlanStrategy and the switch in Build that populates Summary counts.)

btipling (commented):
Deep-arbiter review auto-submitted to make draft comments live for editing. Inlines had empty bodies due to extraction bug in the review tool; being PATCHed to full text in the next steps of this run. Verdict: NEEDS-WORK (WIP). Full details in the (updated) comment bodies and review body.

pkg/dotc1z/engine/pebble/grants.go
```diff
@@ -148,6 +149,132 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
...
+			idxScratch, err = e.writeGrantIndexesScratch(idxBatch, r, idxScratch)
+			if err != nil {
+				return err
+			}
+		}
+
+		if err := priBatch.Commit(pebble.NoSync); err != nil {
+			return err
+		}
+		return idxBatch.Commit(pebble.NoSync)
```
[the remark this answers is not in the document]
kans: thats the point

pkg/sync/expand/topological_merge_streaming.go
```diff
@@ -0,0 +1,397 @@
...
+// key order (see principalGroupStream / projectionContributionStream); the heap
+// surfaces the minimum key and collects all streams positioned at it.
+func mergeContributionGroupStreams(
+	ctx context.Context,
+	destEntitlement *v2.Entitlement,
+	streams []contributionGroupStream,
+	sink destinationSink,
+) error {
+	for _, stream := range streams {
+		defer stream.close()
```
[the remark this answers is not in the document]
kans: meh

ggreer (approved):

PR #972, opened 2026-06-17
[CE-717] Add is hidden field to config

MarcusGoldschmidt:


proto/c1/config/v1/config.proto
```diff
@@ -54,6 +54,7 @@ message Field {
   bool is_required = 5;
   bool is_ops = 6;
   bool is_secret = 7;
+  bool is_hidden = 8;
```
ggreer, line 57: ~What's the difference between secret and hidden?~ Ah, secret is "field is visible, value is hidden".

laurenleach (changes requested):
The linear issue says 
> For fields with ExportTargetOps + WithHidden(true) (e.g. several fields in pkg/field/defaults.go lines 243–255), the hidden flag is silently ignored and those fields appear in the C1 Ops UI when they shouldn't.

But `ExportTargetOps` does hide from c1 UI, I think we can close this out 

https://github.com/ductone/c1/blob/119d291b34593baadecbce2840cbbd5490667a9a/pkg/api/integration/rpc_connector_catalog.go

PR #977, opened 2026-06-19
Grant Hash Digest Index + leaf only tree

mj-palanker:
Introduces grant content hashing.

Adds a new grant index where the key sorts grants into buckets based on the hash of the principal ID such that each entitlement's grants are now sorted into roughly equal sized buckets. The value is the hash of the _content_ of the grant, such that it genuinely should describe it's state.

each "bucket" is actually a set number of bits into the 2 byte hash s.t. we can dynamically "bucket" at higher / lower resolution without re-indexing.

any two c1zs could compare any given bucket for grant equality by just scanning this index.

We also add new objects which stores an XOR rollup of the hashes on an entitlement.

for a given bit count k, we store 2^k leaf nodes where the value is the XOR of all hashes in their bucket.
we also store a root note which is the XOR of ALL grants on the entitlement (equivalent to the rollup of the leaf nodes)


pkg/dotc1z/engine/pebble/options.go
```diff
@@ -65,6 +73,20 @@ func WithDurability(d Durability) Option { return func(o *Options) { o.durabilit
...
 func WithReadOnly(readOnly bool) Option { return func(o *Options) { o.readOnly = readOnly } }
 
+// WithGrantDigestIndex toggles maintenance of the
+// by_entitlement_principal_hash index and the per-entitlement grant
+// digests. Default true (preserves existing behavior).
+//
+// Set false on files that will never be grant-diffed (e.g. local CLI
+// syncs, connector development) to skip the per-grant index write and
+// the seal-time digest fold. NOTE: a file written with this off has no
+// hash index, so a future cross-file diff MUST detect its absence
```
btipling: `grantDigestIndex` is in memory only, so how does a "future cross-file diff" detect its absence? What is written to the disk that can be detected?

pkg/dotc1z/engine/pebble/index_migrations.go
```diff
@@ -63,7 +64,92 @@ type indexMigration struct {
...
+// committed before the digests are built because BuildAllGrantDigests
+// folds over the committed index.
+//
+// No-op when the digest index is disabled (WithGrantDigestIndex(false)):
+// the caller opted out of the index, so we neither write it nor build
+// digests. The migration is still marked applied — re-enabling the
+// index on an existing file requires a fresh sync or explicit rebuild,
+// not an Open-time backfill.
+func (e *Engine) backfillGrantHashIndexAndDigests(ctx context.Context) error {
+	if !e.opts.grantDigestIndex {
```
btipling: when indexMigrations are applied and they return nil we still record the index version with writeAppliedIndexVersion in applyIndexMigrations, so in the case here where we return nil when grantDigestInex is nil, I'm not sure if it is possible to back fill later if grantDigestIndex is enabled

btipling (approved):

kans (changes requested):
Before merge:

1. We should profile this. There is possibly a way to do the work in a single pass or way more efficiently.
2. We need to run the benchmarks for expansion and compaction to see what it does to perf.
3. We need to run the benchmark for the whale C1z to see what happens to perf, there.

Then, we can decide if this tradeoff is a net savings.

pkg/dotc1z/engine/pebble/index_migrations.go
```diff
@@ -63,7 +64,92 @@ type indexMigration struct {
...
+		// buckets (width in bits, chosen per entitlement), leaf keys
+		// carrying a 2-byte left-aligned bucket index; node keys carry
+		// the digested index's id after the sync_id (see
+		// encodeDigestNodeKey). Index entries are unchanged; the bump
+		// forces a digest rebuild, whose per-partition range-clear
+		// removes the v1 nodes.
+		Name:    "grant_by_entitlement_principal_hash",
+		Version: 2,
+		Apply: func(ctx context.Context, e *Engine) error {
+			return e.backfillGrantHashIndexAndDigests(ctx)
```
kans: Historically, grant migrations cause incidents.  For pebble, do we record if we have run a migration?  If not, we probably should.  That info could potentally go in the c1z header, too.  A differetn strategy would be to not run the migration at all, and opt into the new behavior on the c1 side if the c1z supports it.

pkg/dotc1z/engine/pebble/bulk_import.go
```diff
@@ -307,6 +309,24 @@ func (s *BulkGrantShard) AddGrants(ctx context.Context, grants ...*v2.Grant) err
...
 			}
 		}
+		// by_entitlement_principal_hash is the one grant index that
+		// carries a VALUE (the grant content hash), so it's written
+		// separately from the nil-valued grantIndexKeys above. The grant
+		// digests that fold over it are not built here — a reader falls
+		// back to an on-demand index fold, and the on-Open migration
+		// backfills the stored digests when the file is next opened
+		// writable. Skipped entirely when the digest index is disabled.
+		if s.b.e.opts.grantDigestIndex {
```
kans: Stepping through 4 objects violates Code Complete.

pkg/dotc1z/engine/pebble/adapter.go
```diff
@@ -313,6 +314,21 @@ func (a *Adapter) EndSync(ctx context.Context) error {
...
 		)
 	}
+	// Build the per-entitlement grant digests at seal time, after all
+	// grants are written and while still on the fresh-sync NoSync path
+	// (the EndFreshSync flush below hardens the nodes). Non-fatal, like
+	// the stats sidecar: a missing/stale digest only forces a diff
+	// consumer onto the on-demand index fold, and the on-Open migration
+	// backfills it next time the file opens writable. Skipped when the
+	// digest index is disabled — there is no hash index to fold over.
+	if a.engine.GrantDigestIndexEnabled() {
```
kans, line 337: This will load every single grant after we sync, but we already walked through all of them in the process of syncing.  Can we do this work incrementally?

pkg/dotc1z/engine/pebble/grants.go
```diff
@@ -124,6 +132,21 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
 						closer.Close()
 						return err
 					}
+					if mm != nil {
+						// The digest removal needs the old record's content
+						// hash, which folds fields the raw index scan does
+						// not extract (sources) — unmarshal only on this
+						// non-fresh path.
+						old := &v3.GrantRecord{}
+						if err := unmarshalRecord(oldVal, old); err != nil {
```
kans: This should be a shallow unmarshal for the fields we need (probably just the digest), so that we don't allocate a ton of pointers on the heap on the hot path.   Can we get the other fields out of new record or whatever index  is handy?

pkg/dotc1z/engine/pebble/grants.go
```diff
@@ -124,6 +132,21 @@ func (e *Engine) PutGrantRecords(ctx context.Context, records ...*v3.GrantRecord
...
+							closer.Close()
+							return fmt.Errorf("PutGrantRecords: unmarshal old %q: %w", r.GetExternalId(), err)
+						}
+						if err := mm.removeGrant(old); err != nil {
```
kans: mm or removeGrant should have a more descriptive name.

pkg/dotc1z/engine/pebble/digest.go
```diff
@@ -0,0 +1,914 @@
...
+			continue
+		}
+		if err := m.applyPartition(batch, mp); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (m *digestMutator) applyPartition(batch *pebble.Batch, mp *mutatorPartition) error {
```
kans: This function looks like it is doing too much work.

pkg/dotc1z/engine/pebble/digest.go
```diff
@@ -0,0 +1,914 @@
...
+	}
+	for _, n := range mp.nodes {
+		if n.countDelta == 0 && n.xor == zeroDigest {
+			continue
+		}
+		var (
+			curCount  int64
+			curDigest [hashLen]byte
+		)
+		val, closer, err := m.e.db.Get(n.key)
```
kans: This is doing a random get in a loop per grant.  This is probably bad.

pkg/dotc1z/engine/pebble/if_newer.go
```diff
@@ -62,6 +66,18 @@ func (e *Engine) PutGrantRecordsIfNewer(ctx context.Context, records ...*v3.Gran
 					closer.Close()
 					return err
 				}
+				// The digest removal needs the old record's content hash,
+				// which folds fields the raw index scan does not extract
+				// (sources).
+				old := &v3.GrantRecord{}
+				if err := unmarshalRecord(oldVal, old); err != nil {
```
kans: Same story here, we try very hard in this function to unmarshal any protos.

mj-palanker:
Benchmarks run with and without the Feature, adds ~1 second at 10M grants (from 7s to 8s)
```
goos: darwin
goarch: arm64
pkg: github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble
cpu: Apple M4
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        33246971041 ns/op             8262 ms/EndSync        19694 ms/PutGrants 17276227616 B/op        70217760 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        32720522541 ns/op             8237 ms/EndSync        19435 ms/PutGrants 17157200984 B/op        70210882 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        34873743666 ns/op             8348 ms/EndSync        21436 ms/PutGrants 17137948992 B/op        70210371 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        35549664292 ns/op             8205 ms/EndSync        22288 ms/PutGrants 17154485096 B/op        70210820 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        34943085750 ns/op             8195 ms/EndSync        21765 ms/PutGrants 17244106352 B/op        70213758 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        34301921666 ns/op             8199 ms/EndSync        20812 ms/PutGrants 16943747552 B/op        70205628 allocs/op
PASS
ok      github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble      233.487s
goos: darwin
goarch: arm64
pkg: github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble
cpu: Apple M4
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        33889880458 ns/op             7135 ms/EndSync        21917 ms/PutGrants 16823701896 B/op        70195957 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        33256215584 ns/op             7072 ms/EndSync        21228 ms/PutGrants 16844674776 B/op        70191977 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        34725127125 ns/op             7107 ms/EndSync        22654 ms/PutGrants 16831746496 B/op        70192288 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        34117817333 ns/op             7078 ms/EndSync        22076 ms/PutGrants 16914976696 B/op        70193591 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        33318628875 ns/op             7190 ms/EndSync        21272 ms/PutGrants 16770682784 B/op        70190711 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        33588608750 ns/op             7199 ms/EndSync        21690 ms/PutGrants 16993501960 B/op        70196186 allocs/op
PASS
ok      github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble      228.640s
```

kans:
I just merged my other branch.  Apologies for the conflicts.  Fable and I talked about this branch a bit, and this is what it suggests another agent do:  

# Task: Port the grant-digest system from `mjp/grant-digest` onto main (post-injective-keys)

## Situation

Branch `mjp/grant-digest` (6 commits, ~3,000 added lines) implements per-entitlement
grant digests for the Pebble c1z engine: bucketed XOR set-digests that answer "does
this entitlement hold exactly the same grants as in some other sync/file?" with a
single key read, and when the answer is no, localize the difference to principal-hash
buckets so a diff driver loads only the differing grants.

The branch was cut from v0.15.6. Since then, main merged the large "injective keys"
branch (PR #993, commit 393077d9), which rewrote the substrate the digest branch sits
on. A naive merge conflicts in 6 files, and the textual conflicts understate the
breakage:

1. **Grant identity changed.** The branch keys and hashes grants by `external_id`.
   On main, grant primary keys are the injective identity tuple
   `v3|typeGrant|0x00|ent_rt|ent_rid|ent_flag|ent_tail|p_rt|p_id` (see
   `pkg/dotc1z/engine/pebble/keys.go`, `identity.go`).
2. **The write paths the branch hooks were rewritten.** `PutGrantRecords`,
   `BulkGrantShard.AddGrants`, `UnsafePutUniqueGrantRecords` — `grants.go` changed
   ~980 lines on main.
3. **EndSync already has the pass we need.** Main's `Adapter.EndSync` runs one sealed
   O(grants) scan — `BuildDeferredGrantIndexes` in `deferred_index.go` — that builds
   the by_principal index via spill sorters → SST → IngestAndExcise, and fuses sync
   stats. The branch's `BuildAllGrantDigests` is a second full pass; we don't want it.

## Decision (already made — do not relitigate)

**Keep the hash index (`idxGrantByEntitlementPrincipalHash`, 0x08), but build it ONLY
in the deferred EndSync pass. No inline maintenance anywhere.**

Rationale: primary grant keys sort by meaningful identity, not hash, so digest buckets
are NOT contiguous in the primary keyspace. The index is what makes "scan one dirty
bucket" O(bucket) instead of O(entitlement) — essential for the motivating case
(whale entitlement, tiny diff). What we drop from the branch is the inline write-path
emission, not the index itself.

## Index redesign (verified against branch encoder + main's machinery)

New key layout — same shape as the branch's `encodeGrantByEntPrincHashIndexKey`, with
the entitlement identity substituted for `entitlement_id` and `external_id` dropped:

    key   = v3 | typeIndex | idxGrantByEntitlementPrincipalHash | 0x00 |
            ent_rt | 0x00 | ent_rid | 0x00 | ent_flag | 0x00 | ent_tail | 0x00 |
            <raw bucket hash: digestBucketHashLen = 2 bytes> |
            p_rt | 0x00 | p_id
    value = 8-byte grant content hash (xxHash64)

Verified properties:

- The `digestIndexSpec` shape contract (partition prefix | raw fixed-width hash |
  tail) is preserved, so `bucketBounds`'s byte arithmetic works unchanged. Tuple
  escaping (no bare 0x00 in segments) + the load-bearing trailing separator keep
  partition prefixes prefix-free, so whole-key byte order = (entitlement, hash,
  principal). Keep the branch's positional-decoder discipline: raw hash bytes can be
  0x00, so generic tuple walkers must never cross the hash region (see the branch's
  `decodeEntPrincHashTail` and its doc comment).
- Uniqueness: (ent identity, principal identity) IS the grant primary identity, so
  one index row per grant by construction — no duplicate-key failures in the merge
  (`mergeSortedSpillChunksToSST` treats duplicates as corruption).
- The tail carries the full grant identity, so cross-file bucket comparison is a pure
  merge scan of two index ranges with ZERO point Gets (an improvement over the
  branch, which needed a Get per index entry). The primary key for a differing grant
  is reconstructable from the index key by removing the hash region — one Get per
  actually-changed grant, done by byte splice, not decode+re-encode.
- Keep the branch's 2-byte stored bucket hash (commit 19a15988) —
  `digestBucketHashLen = 2` covers `digestMaxWidthBits = 16`.

## Build: fuse into the deferred pass

In `buildDeferredGrantIndexesLocked` (`deferred_index.go`):

1. Add a second spill-sorter family for the hash index, alongside the existing
   by_principal sorter. The machinery already supports valued entries
   (`spillSorter.add(key, val)`; `kvView.valEnd`; the chunk reader and
   `bulkSSTWriter.add` round-trip values) — the by_principal family just happens to
   use nil values.
2. Per scanned grant, compute:
   - the index key by SPLICING segments out of the primary key (same trick as
     `appendGrantByPrincipalKeyFromPrimary` — segments are escaped, raw byte splice
     is byte-identical to decode+re-encode);
   - the bucket hash from the spliced principal segments;
   - the content hash = xxHash64(primary-key tail ‖ sorted source-entitlement ids).
     Source ids come from a raw protobuf value scan (sources are field 9; follow the
     `raw_records.go` helper pattern, e.g. `scanGrantEntitlementResourceTypeRaw`).
     No proto unmarshal anywhere on this path.
3. Ingest the merged index SST with its own IngestAndExcise call, excise span =
   the branch's `GrantByEntPrincHashLowerBound/UpperBound`.
4. **Fold the digests during the sorter's final merge.** Entries emit in
   (entitlement identity, bucket hash) byte order. Accumulate (xor, count) at max
   width — 2^16-bucket scratch, ~1.5MB — and fold down to `chooseDigestWidth`'s pick
   at each entitlement boundary (the total count isn't known until then; XOR
   split-independence makes the fold-down exact).
5. **Digest nodes do NOT ride the index SST.** typeDigest (0x0A) is separated from
   the index keyspace by typeCounter (0x08) and typeSession (0x09); a single excise
   span covering both would wipe counters and sessions. Write digest nodes via
   normal sealed-path batch Sets during the merge (they're tiny: one root + sparse
   leaves per entitlement), or a second SST with its own excise span limited to the
   digest keyspace.
6. Zero-grant entitlements must still get a `{count: 0}` root node (the branch's
   "empty vs. never built" distinction). After the grant work, do one pass over the
   entitlement primary keyspace and write roots for entitlements not seen.
   Entitlements are by far the smallest record type — don't bother fusing this.

## What ports from the branch vs. what to write fresh

Ports nearly unchanged:
- `digest.go` + `digest_test.go` (~900 + ~1,200 lines): the generic digest core —
  node framing, `chooseDigestWidth`, `dirtyPartitionBuckets` cross-width merge
  comparison, `digestMutator`. One generalization needed:
  `digestIndexSpec.partitionPrefix` and `encodeDigestPartitionPrefix` take the
  partition as a `string` (was entitlement_id). Re-key digest partitions by the
  entitlement's encoded primary-key tail bytes (spliced, not re-encoded).
- The `typeDigest` (0x0A) keyspace + digest node key encoders from the branch's
  `keys.go`; `DigestLowerBound/UpperBound`; the synccompactor `bucket_plans.go`
  additions (digest keyspace AND the hash-index keyspace must be in the
  cleanup/clone/compaction plans).
- `pkg/connectorstore/connectorstore.go`: `GrantDigest`, `GrantDigestNode`,
  `GrantDigestBucket`, `EntitlementGrantDigestReader`. Note its methods take
  `entitlementID string` — align with how main's reader addresses entitlements
  under the identity scheme.
- `c1file.go` option + `engine_registry.go` plumbing (`WithGrantDigestIndex`).
- Reader plumbing from the branch's `adapter_reader.go` additions; the bucket grant
  loader now scans the index range and reconstructs primary keys by splice (no
  per-entry point Get except for materializing changed grants).

Write fresh:
- `grant_digest.go`: content hash, bucket hash, index-key splice helpers (above).
- The deferred-pass integration (above).

## Post-seal mutations: v1 = invalidate

Do NOT port the incremental `digestMutator` wiring into the rewritten Put/Delete
paths. On any post-seal grant mutation, DeleteRange the affected entitlement's digest
partition (`encodeDigestPartitionPrefix` — the branch already has this
inconsistency-drop path) AND its hash-index range (the index is stale too). Readers
treat a missing digest root as "fall back to on-demand fold over the primary
keyspace". Leave the mutator code in digest.go for a follow-up.

## Explicitly drop from the branch

- All inline write-path emission of the index: the grants.go, bulk_import.go
  (`BulkGrantShard.AddGrants`), and if_newer.go hooks.
- `BuildAllGrantDigests` (the second EndSync pass) and its `Adapter.EndSync` hook.
- The branch's `index_migrations.go` backfill.
- external_id anywhere in the index key or content hash.

## ABI / migration notes

- Main's index-migration registry is currently EMPTY ("none yet, because we have no
  existing data"), so the content-hash definition can be set freely now. Once
  shipped it is on-disk ABI: document the field set ("the same grant" = identity
  tuple + sorted source ids; excludes sync-relative/transient state — discovered_at,
  needs_expansion, expansion, annotations) with a comment block like the branch's,
  and note that changing it requires a migration bump.
- Digest node framing, `digestBucketHashLen`, `digestLeafPrefixLen`, the XOR
  combiner, and leaf left-alignment are ABI — keep as the branch defines them.
- The digest/index build must be non-fatal at EndSync (warn + continue), matching
  the stats sidecar's treatment of seal-time extras.

## Verification

- Port `digest_test.go`; keep the cross-width comparison and split-independence pins;
  adapt grant-instantiation tests to identity keys.
- Pin test: content hash / index key computed from raw key+value bytes equals the
  same computed from a decoded record (mirrors
  `TestAppendGrantByPrincipalKeyFromPrimary`).
- End-to-end: two syncs with a known grant delta → digest roots differ, dirty buckets
  localize the delta, bucket scan yields exactly the changed grants; identical
  syncs → roots equal, `dirtyPartitionBuckets` returns nil.
- Post-seal mutation → digest partition gone → reader falls back to on-demand fold
  and still answers correctly.
- Whale-shaped bench: deferred-pass runtime with vs. without the second sorter
  family. It roughly doubles sorter volume, and the 128MiB
  `deferredIndexSpillChunkBytes` sizing rationale assumed ONE family — revisit the
  memory-budget comment (two sorters share the sort semaphore and freelist sizing).
- `make lint` and `go test ./pkg/dotc1z/...` clean.

## Key references

- Branch `mjp/grant-digest`: `digest.go`, `grant_digest.go`, `digest_test.go`,
  `keys.go` additions (incl. `encodeGrantByEntPrincHashIndexKey` and its raw-hash
  doc comment), `connectorstore.go` API.
- Main: `pkg/dotc1z/engine/pebble/deferred_index.go` (the pass to fuse into;
  `appendGrantByPrincipalKeyFromPrimary` for the splice pattern; the
  finalize → `mergeSortedSpillChunksToSST` → `IngestAndExcise` sequence),
  `bulk_import.go` (spillSorter, valued entries, merge), `keys.go` (identity
  layout), `raw_records.go` (raw value scans), `sync_stats_sidecar.go` (precedent
  for fused, non-fatal seal-time work).

kans:
@mj-palanker Can we close this PR?

PR #982, opened 2026-06-24
Entitlement grant count stats

MarcusGoldschmidt:


cmd/baton/stats_info.go
```diff
@@ -0,0 +1,165 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func renderSyncRunInfo(sync *v2.SyncRun) error {
+	startedAt := sync.GetStartedAt().AsTime()
+	endedAt := sync.GetEndedAt().AsTime()
+
```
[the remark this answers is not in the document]
MarcusGoldschmidt: We check nil before, it's good

cmd/baton/stats_info.go
```diff
@@ -0,0 +1,165 @@
+package main
```
ggreer: Why not replace the old stats command with this?

proto/c1/reader/v2/sync.proto
```diff
@@ -25,6 +25,7 @@ message SyncStats {
   int64 grants = 4;
   map<string, int64> resources_by_resource_type = 5;
   map<string, int64> grants_by_resource_type = 6;
+  map<string, int64> entitlements_by_resource_type = 7;
```
ggreer, line 28: In hindsight, we probably should have done something like this:

```proto

message SyncStats {
  int64 resource_types = 1;
  int64 resources = 2;
  int64 entitlements = 3;
  int64 grants = 4;
  map<string, ResourceStats> resource_stats = 5;
}

message ResourceStats {
  int64 resources = 1;
  int64 entitlements = 2;
  int64 grants = 3;
}
```

I don't think it's worth making that change right now though.

ggreer (commented):
I think we should have just one stats command, but otherwise the code looks good.

cmd/baton/stats_info.go
MarcusGoldschmidt: Replaced

ggreer (approved):

PR #984, opened 2026-06-25
Add generic support for pebble to "external resource c1z" 

mj-palanker:
we were specifically using sqlite only methods in this external resource use case
switch to engine for pebble support


ggreer:
Closing because https://github.com/ConductorOne/baton-sdk/pull/983 accomplishes the same thing.

PR #990, opened 2026-06-29
multi-engine-compaction support

kans:
The new compaction policy is that if pebble is the target, or if any c1z is pebble, we convert to pebble.  Or, if sqlite is the target and any sync is pebble, we error out.  If the base sync is not pebble, we can not choose Fold.

### DIFF
--- a/pkg/synccompactor/compactor_pebble.go
+++ b/pkg/synccompactor/compactor_pebble.go
@@ -483,37 +483,58 @@ func (c *Compactor) compactPebbleFold(ctx context.Context) (string, error) {
 	// base at entries[0]); strictly-newer-wins makes earlier
 	// applications take precedence on ties.
 	var foldStats mergepkg.FoldStats
+	// SQLite/v1 partials are converted to Pebble in the tmp dir before being
+	// folded in; their converted copies are removed when this run completes.
+	var convertedInputs []string
+	defer func() {
+		for _, path := range convertedInputs {
+			_ = os.Remove(path)
+		}
+	}()
 	for i := len(c.entries) - 1; i >= 1; i-- {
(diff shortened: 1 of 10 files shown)


pkg/synccompactor/compactor_pebble.go
```diff
@@ -645,6 +666,111 @@ func copyFileForFold(src, dst string) error {
...
+			SyncType: string(st),
+		}.Build())
+		if err != nil {
+			return "", err
+		}
+		sync := resp.GetSync()
+		if sync == nil {
+			continue
+		}
+		if best == nil || sync.GetEndedAt().AsTime().After(best.GetEndedAt().AsTime()) {
```
gontzess: Low severity / High confidence — `resolveSQLiteCompactionSyncID` (`pkg/synccompactor/compactor_pebble.go:~700`)
- It queries Full / ResourcesOnly / Partial separately and replaces `best` only when a candidate's `ended_at` is strictly greater. On an `ended_at` tie, slice order wins (Full beats a later Partial).
- This diverges from the repo's canonical latest-finished ordering `ended_at DESC, sync_id DESC` (`pkg/dotc1z/sync_runs.go:256-260`) and the Pebble manifest selector's sync_id tiebreak (`compactor_pebble.go:414-419`).
- Impact: with an empty `CompactableSync.SyncID` and a timestamp tie, the wrong source sync is selected → the converted Pebble input can drop the newer sync's changes.
- Fix direction: add the sync_id tiebreak; add a regression test with two compactable sync types sharing `ended_at `and empty `SyncID`.

gontzess (approved):
left a small note but otherwise looks good to me

PR #991, opened 2026-06-30
[Draft] ~50% faster grant expansion

kans:
## Summary

Speeds up Pebble grant expansion on large tenants by attacking the two biggest
write-path costs, both behind off-by-default env flags so merging changes no
behavior until explicitly enabled (remove those)

1. **Deferred index build** (`BATON_PEBBLE_DEFER_EXPANSION_INDEXES`) — during
   expansion, skip the three grant index families that are *not* read by the
   merge (`by_entitlement_resource`, `by_principal`, `by_principal_resource_type`)
   and rebuild them once at `EndSync` from a single `by_entitlement` scan →
   spill-sort → one sorted SST ingest per family. Trades millions of
   out-of-order inline index writes (the dominant compaction-churn source) for
   one external sort. `by_entitlement` + `needs_expansion` stay inline (the merge
   reads `by_entitlement`).

2. **Synthesized-grant skip-Get** (`BATON_EXPAND_SKIP_GET`) — expanded grants the
   merge knows are brand-new skip the per-record read-before-write `Get` (and
   stale-index cleanup) that only exists to merge a prior record. Routed via a
   new `StoreNewExpandedGrants` fast path (`PutSynthesizedGrantRecords`).

3. **Write-path instrumentation** — `expandWriteStats` attributes expansion
   write time across `rbw-get` / marshal / index-delete / primary-commit /
   index-commit and breaks index volume out per family, plus a persisted whale
   harness test for end-to-end measurement.

TODO: 

- [ ] Fix soundness issue with colon delimited indexes.
- [ ] Remove the stats thing (probably)

## Performance (whale: ~3.68M base → 54M expanded grants)

| config | expansion wall | `rbw-get` | `idx-commit` |
|---|---|---|---|
| baseline (inline) | 9m14s | 3m12s (54M no-op Gets) | 2m08s |
| defer + skip-Get | **3m39s** | **8ms** (1.5K real Gets) | **11.6s** |

~2.5× faster expansion; compaction write dropped ~100GB → ~53GB, estimated debt
17GB → ~6GB. The deferred build itself is ~1m17s for all three families
(single amortized scan + parallel merge).

## Safety

- Both paths are **off by default** (env-gated); no behavior change on merge.
- `skip-Get` depends on the deterministic grant-id assumption and is **not yet
  safe to enable in production** — it must wait on the injective grant-id fix
  (a non-injective id could let two distinct `(entitlement, principal)` tuples
  collide). Shipped here gated-off for measurement only.
- The deferred build is a pure function of durable, checkpoint-captured
  `by_entitlement` state (resumable) and produces **byte-identical** index
  keyspaces to inline writes (`TestDeferredGrantIndexesMatchInline`).

### DIFF
--- a/pkg/dotc1z/engine/pebble/adapter.go
+++ b/pkg/dotc1z/engine/pebble/adapter.go
@@ -322,6 +322,15 @@ func (a *Adapter) EndSync(ctx context.Context) error {
 			zap.Error(err),
 		)
 	}
+	// If the expansion write path deferred the principal/resource index
+	// families (BATON_PEBBLE_DEFER_EXPANSION_INDEXES), rebuild them now in one
+	// sorted ingest, before the durability flush so EndFreshSync hardens them.
+	if a.engine.deferGrantIndexes && a.engine.deferredIdxPending.Load() {
+		if err := a.engine.BuildDeferredGrantIndexes(ctx); err != nil {
+			return fmt.Errorf("EndSync: build deferred grant indexes: %w", err)
+		}
+		a.engine.deferredIdxPending.Store(false)
+	}
 	// Single flush + WAL fsync at sync end. This is the durability
 	// boundary — counterpart to MarkFreshSync at StartNewSync. After
 	// this returns, all writes from the sync are on disk.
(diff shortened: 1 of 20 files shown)


kans:
superseded by https://github.com/ConductorOne/baton-sdk/pull/993

PR #992, opened 2026-06-30
Windows: Add config option to log to event log.

ggreer:
Add option to output logs to the windows event log. Default behavior (logging to a file) is unchanged.


kans:
This is a breakign change.  Should we do the opposite?

ggreer:
I'll flip the default logic, but I thing long term, we should default to logging to the windows event log, not a file on disk.

kans (approved):

PR #999, opened 2026-07-08
Entitlement Grant Hash Index & Entitlement Grant Digest Tree

mj-palanker:
Adds a new "hash index" which stores a descriptive hash of the content of a grant on a new index which is shaped exactly like the normal grant key but with a new 2 bytes of the hash of the principal keys. This distributes all grants for an entitlement randomly across the keyspace when considering only the hash. A "bucket" is just a prefix scan of those 2 bytes, meaning you can chop the grant space into as many as  2^16 even buckets. 

Also introduces the concept of a grant digest which is the XOR of all grant hashes for a given bucket (plus the # of grants in said bucket). By default we always create a root digest storing info on all the entitlement's grants, plus a set of "leaf nodes" for some desired "depth" into the tree.

This desired depth is calculates such that (if distributed evenly) we have no more than 512 grants per bucket.
16 bits lets this hold until we hit 33.5 Million Grants on a single entitlement.

since this desired depth calculated value isn't assumed ever, we could change this in the sdk without breaking anything

Any stored depth can be rolled up into any less granular depth simply by XOR-ing the hashes of nodes, and adding the counts, meaning you can trivially compare grant digests stored at different resolutions. We may want to leverage this to store a fixed max resolution in c1.

Derived from https://github.com/ConductorOne/baton-sdk/pull/977 via robot onto main

perf tests show we _doubled_ the End Sync time, but that's from ~25ms to ~50ms 
```
goos: darwin
goarch: arm64
pkg: github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble
cpu: Apple M4
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        2549355708 ns/op                41.00 ms/EndSync              2345 ms/PutGrants 5171366088 B/op    46525 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        2218060458 ns/op                54.00 ms/EndSync              2027 ms/PutGrants 5170921376 B/op    42772 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        2210106875 ns/op                68.00 ms/EndSync              2005 ms/PutGrants 5170957704 B/op    42766 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        2272597000 ns/op                48.00 ms/EndSync              2084 ms/PutGrants 5170922776 B/op    42777 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        2172931042 ns/op                50.00 ms/EndSync              1986 ms/PutGrants 5170920936 B/op    42764 allocs/op
BenchmarkRegisteredPebbleWritePack/grants=10000000-10                  1        2144806791 ns/op                46.00 ms/EndSync              1958 ms/PutGrants 5170960672 B/op    42774 allocs/op
PASS
ok      github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble      41.450s
goos: darwin
goarch: arm64
pkg: github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble
cpu: Apple M4
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        3560330167 ns/op                28.00 ms/EndSync              3365 ms/PutGrants 5031893760 B/op    45705 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        2149813833 ns/op                27.00 ms/EndSync              1985 ms/PutGrants 5031500976 B/op    42041 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        2136727709 ns/op                27.00 ms/EndSync              1977 ms/PutGrants 5031444792 B/op    42030 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        2142267459 ns/op                26.00 ms/EndSync              1979 ms/PutGrants 5031456872 B/op    42053 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        2156140416 ns/op                26.00 ms/EndSync              1992 ms/PutGrants 5031494040 B/op    42037 allocs/op
BenchmarkRegisteredPebbleWritePack_NoDigestIndex/grants=10000000-10                    1        2119374666 ns/op                26.00 ms/EndSync              1956 ms/PutGrants 5031441216 B/op    42027 allocs/op
PASS
ok      github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble      41.383s
```


kans:
Final brief: 

@mj-palanker do we really want to exclude sources from grant hashes?  Thats user facing data, and maybe goes into ticket processing?  

@alan-lee-12 you possibly know best ^

# Brief: grant-digest review fixes (mjp/grant-digest-v2)

Findings from a correctness/error-recovery review of the branch. One
must-fix soundness hole, a few small hardening items, one decision to
sign off. No changes to the digest core design or the hash ABI.

## Fix 1 (must-fix): standalone build crash window breaks "root present ⇒ index present"

`buildGrantDigestsFromSpill` commits digest nodes during the merge
(`closePartition` flushes batches every 4 MiB) and again at
`fold.finish()` — including the global root — **before** the hash-index
`IngestAndExcise`. The fused seal pass is protected by the durable
`deferredIdxPending` marker (a crash reruns the full build); the
standalone `BuildGrantDigests` path has no equivalent.

Failure sequence: inline-only sync (deferred marker never armed) →
EndSync → `BuildGrantDigests` → process killed mid-merge or between
`fold.finish()` and the ingest. Committed WAL writes survive a process
crash, so partial (or all) digest roots — possibly the global root —
are durable while the index ingest never ran. On resume,
`RepairMissingGrantDigests` either fast-paths on the surviving global
root or repairs only partitions *missing a root*, trusting every
crashed-build root that committed. The file seals with correct-looking
digests over an **empty hash index**: `ScanEntitlementGrantBucket`
yields zero grants for dirty buckets and
`GetEntitlementGrantDigestNodes` at a finer-than-stored level scans the
empty index and reports "no grants". That is the silent
skip-entitlements-at-uplift failure mode.

Fix options (pick one):

- **Preferred — durable build-pending marker**, mirroring
  `deferredIdxPending`: arm (fsync'd) before the standalone build's
  first node commit, clear after `writeMissingEntitlementDigestRoots`
  completes. `Open` and `RepairMissingGrantDigests` seeing the marker →
  `dropAllGrantDigestState` first. Converts every crash into "digests
  absent", which the design already treats as safe.
- Alternative — reorder publication: stream digest nodes to their own
  SST (they are emitted in ascending key order) and ingest it *after*
  the index SST, so a crash can only leave index-without-digests
  (reads as missing → safe).

Verification: a test that kills the build between the digest-node
commit and the index ingest (inject a failpoint or split the function),
reopens, reruns EndSync, and asserts either no digest state at all or
digest state whose hash index matches a from-scratch build byte-for-
byte. Cover both the mid-merge and post-`finish()` windows.

## Fix 2: `writeGrantIndexes` swallows identity errors — DONE

Fixed on `origin/mjp/grant-digest-v2` ("fixup for error swallow"):
`writeGrantIndexes` now propagates the `grantIdentityFromRecord` error
instead of silently skipping invalidation. Nothing further to do once
that commit is in your local stack.

## Fix 3: repair path drops malformed grant keys silently

`repairOneGrantDigestPartitionLocked` skips keys that fail
`splitGrantPrimaryKey` with a bare `continue`. The build paths count and
`Error`-log the same condition ("rows NOT represented in the digests").
Add the same counter + log to the repair path.

## Fix 4: doc contradiction on the `ComputeEntitlementBucketDigest` fallback

`GetEntitlementDigestRoot` / `getPartitionDigestRoot` doc comments
invite falling back to `ComputeEntitlementBucketDigest` when no root
exists — but the index only exists alongside built digests, so on a
never-built file that fold returns `{0, 0}` ("zero grants"): the exact
trap `dirtyPartitionBuckets`' own comment warns against. No in-repo
caller misuses it; fix the comments so the exported API doesn't
recommend the unsafe pattern.

## Fix 5 (scale, not correctness): bound repair memory

`repairOneGrantDigestPartitionLocked` buffers one partition's full row
set in memory, one allocation per row. A fold repair touching an
"everyone"-style entitlement (millions of grants) is multi-GB. Add a
row-count/byte threshold above which the repair falls back to the
spill-sorter machinery (or degrade to `BuildGrantDigests`).

## Decision (sign-off, no code): content-hash field exclusions

`grantContentHash64` deliberately excludes grant annotations,
source-record VALUES, external_id, and expansion flags. Any change in
only those fields fingerprints as "no change" and gets skipped by a
digest-driven diff. Confirm with the uplift consumer that nothing it
needs rides on grant annotations (e.g. immutability/system markers).
If something does, that is an ABI change — bump
`grantDigestABIVersion` — and far cheaper to decide now than after
files ship.

## Noted, no action required

- Bucket-plan `Compactor.Compact` (synccompactor/pebble) replaces
  grants and digest buckets in separate ingests; a crash mid-loop
  leaves new grants under old, present digests. Recovery contract is
  "re-run Compact" and there is no in-repo production caller — but if
  the API is live in c1, consider dropping the digest/hash buckets
  first and replacing them last.
- Orphan partitions (grants without an entitlement record) invalidated
  post-seal are never repaired and are excluded from the recomputed
  global root. All consequences are conservative (false-dirty, never
  false-clean); documented in `RepairMissingGrantDigests`.

kans (approved):

PR #1000, opened 2026-07-08
feat(lambda): binary wire frames for annotation-lossless transport

cxcollins:
The lambda transport encodes frames with protojson, which must resolve every google.protobuf.Any against the process type registry. Connector- specific annotation types (e.g. baton-jira's CustomField on ticket schema custom fields) are not linked into every runtime, so the outbound marshal hard-failed the whole sync while the inbound filter silently dropped them — custom fields never survived the axiomatic runtime (CE-918).

Add a v2 wire encoding that carries the binary proto bytes of the transport frame, base64-embedded in the JSON payload. Binary proto copies Any payloads verbatim with no type resolution, so unknown annotation types cross the transport byte-identical to what a native connector emits, and registry-aware consumers decode them losslessly.

Version skew needs no negotiation state:

- Requests dual-encode: legacy protojson fields and the frame share one JSON object. Old connectors discard the unknown fields; new connectors prefer the frame. Oversized payloads drop the frame.
- Responses are frame-only exactly when the request carried a frame, which proves the invoker reads them. Legacy requests get legacy responses.
- When protojson cannot produce the legacy view at all (an Any the sending process does not link), the request goes out frame-only and the invoker logs why. The send side never prunes data it happens not to resolve: the sender's registry is no authority on what the receiver understands or needs, so a payload either arrives intact or fails loudly — the pre-frame behavior — rather than arriving silently degraded. The receive-side filter is unchanged: dropping what this process cannot use is a local decision.

Mixed-version pairings therefore behave exactly as they do today (loud failure on payloads the legacy encoding cannot represent); the fix turns on when both ends are frame-capable. Deploy ordering note: bump c1's baton-sdk and rebuild the axiomatic runtime promptly after release — until both land, exotic-annotation payloads keep failing exactly as they do today.

My quick handwritten hitting summary:
- This code is the send-side equivalent of this [PR](https://github.com/ConductorOne/baton-sdk/pull/958/changes)
- It also adds functionality to not lose the custom fields annotations (the above PR was just ignoring them), which is why they were being lost in the JC schema

**Here is a visual summary of how the changes will impact the shape of the protobuf**

1. Old v1 - pure protojson (what every request was before the PR)

{
  "method": "/c1.connector.v2.TicketsService/CreateTicket",
  "req": {
    "@type": "type.googleapis.com/c1.connector.v2.TicketsServiceCreateTicketRequest",
    "schema": {
      "id": "10001",
      "annotations": [
        {
          "@type": "type.googleapis.com/c1.connector.jira.v1.CustomField",
          "fieldId": "customfield_10042"
        }
      ]
    }
  },
  "headers": { "x-baton-connector-config-version": "12" }
}

(If the sending process does not link CustomField, this marshal hard-fails and the call never leaves the process — the CE-918 crash.)

2. New v1 - dual-encoded (what a new invoker sends when the legacy view is representable)

{
  "method": "/c1.connector.v2.TicketsService/CreateTicket",
  "req": {
    "@type": "type.googleapis.com/c1.connector.v2.TicketsServiceCreateTicketRequest",
    "schema": { "id": "10001", "annotations": [ ...same annotations... ] }
  },
  "headers": { "x-baton-connector-config-version": "12" },

  "v": 2,
  "frame": "CiwvYzEuY29ubmVjdG9yLnYyLlRpY2tldHNTZXJ2aWNlL0NyZWF0ZVRpY2tldBKk..."
}

3. New v2 - frame only

{
  "v": 2,
  "frame": "EqgBCkV0eXBlLmdvb2dsZWFwaXMuY29tL2MxLmNvbm5lY3Rvci52Mi5UaWNrZXRz..."
}

Used for (a) every response to a frame-bearing request, and (b) requests whose legacy view protojson cannot represent — an Any the sender does not link. In case (b) the invoker logs the reason; a pre-frame connector fails on such a payload, exactly as the failed marshal would have pre-frames.

https://linear.app/ductone/issue/CE-918

### DIFF
--- a/pkg/lambda/grpc/client.go
+++ b/pkg/lambda/grpc/client.go
@@ -11,6 +11,8 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/lambda"
 	"github.com/aws/aws-sdk-go-v2/service/lambda/types"
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
@@ -24,9 +26,17 @@ type lambdaTransport struct {
 }
 
 func (l *lambdaTransport) RoundTrip(ctx context.Context, req *Request) (*Response, error) {
-	payload, err := req.MarshalJSON()
+	payload, frameOnly, err := req.marshalPayload()
 	if err != nil {
 		return nil, fmt.Errorf("lambda_transport: failed to marshal frame: %w", err)
+	}
+	if frameOnly != nil {
+		ctxzap.Extract(ctx).Warn(
+			"lambda_transport: request has no legacy encoding, sending v2 frame only; a connector on a pre-frame SDK cannot process this call",
+			zap.String("method", req.Method()),
+			zap.String("function_name", l.functionName),
+			zap.NamedError("legacy_encoding_error", frameOnly),
+		)
 	}
 
 	input := &lambda.InvokeInput{
--- a/pkg/lambda/grpc/server.go
+++ b/pkg/lambda/grpc/server.go
@@ -236,7 +236,18 @@ func TimeoutForRequest(req *Request) (time.Duration, bool, error) {
 	return 0, false, nil
 }
 
+// Handler serves one transport request. The response echoes the request's
+// wire version so v2 invokers get lossless frames and legacy invokers get
+// protojson (see Response.MarshalJSON).
 func (s *Server) Handler(ctx context.Context, req *Request) (*Response, error) {
+	resp, err := s.handle(ctx, req)
+	if resp != nil {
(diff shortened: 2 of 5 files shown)


kans (changes requested):
The directionality is wrong for the authority (the proto registry) to make the call to prune stuff.  Its different if we strip the data that we don't understand for our own usage. We couldn't have done anything with that data in the first place because _we don't understand what it is_.  Stripping data we don't understand and giving it to someone else is incorrect, because we don't know what they understand and what they need.
 
1. keep the dual encoding and handshake bits
2. entirely drop prune.go (I think), and replace marshalTolerantProtoJSON wiht an ordinary proto marshal.
3. Update the connectors to stop sending crazy annotations and use the general blob

cxcollins:
Updating based on your comments to basically get rid of the silent failures that the original changes could have caused.

1. Kept the new formatting and versioning
2. Got rid of prune (I think you're right) and replaced marshalTolerantProtoJSON with a regular protojson.Marshal
3. The connectors and any annotations themselves aren't causing the issue, it's the in between (which this PR fixes)

kans (approved):

PR #1003, opened 2026-07-09
Make uhttp transport timeouts configurable and add connection retry-with-reconnect

johnallers:
## Summary

The `uhttp` transport hardcodes its connection timeouts — idle-conn (30s), response-header (60s), and the HTTP/2 PING keepalive (15s+15s = 30s). Those defaults were deliberately tuned for chatty API connectors behind an egress proxy, but two gaps fall out of them:

1. **A single transient connection blip can abort a whole sync.** When a proxy silently drops a pooled/tunneled connection, it surfaces as `http2: client connection lost` or `connection reset`, which `wrapTransientNetworkError` maps to `codes.Unavailable`. Nothing in `uhttp` retries it, so the error propagates up and the sync workflow discards its in-progress `c1z` and restarts. On a long, per-entity (N+1) walk through a flaky proxy, a run may rarely get an uninterrupted window to completion.
2. **No way to tolerate slow synchronous endpoints.** An endpoint that blocks before sending response headers (e.g. a report-generation call) hits the fixed 60s `ResponseHeaderTimeout` — which, because the transport uses `http2.ConfigureTransports`, applies over HTTP/2 too — with no knob to raise it.

## What changed

**Configurable timeouts** (`pkg/uhttp`): new options `WithIdleConnTimeout`, `WithResponseHeaderTimeout`, `WithHTTP2KeepAlive`, and `WithMaxConnectionRetries`. The tuning fields become nil-able pointers: a `nil` option preserves the existing default (via the new `default*` constants), so an explicit `0` can still mean "disabled/unlimited" per net/http semantics. **Timeout defaults are unchanged.**

**Retry-with-reconnect** (`pkg/uhttp`): a new `retryRoundTripper` sits below the error-wrapping layer and transparently replays a request on reconnectable transport errors (new `IsReconnectableError` helper — `http2 client connection lost`, `ECONNRESET`, `EPIPE`, `net.ErrClosed`, unexpected/EOF; **timeouts/`DeadlineExceeded` are deliberately excluded**). When the HTTP/2 health check tears down a dead `ClientConn` it is evicted from the pool, so the retried `RoundTrip` dials fresh — retry here also reconnects. Only **idempotent** requests (safe methods, or a request carrying an `Idempotency-Key`) with a **rewindable** body are retried; jittered exponential backoff is bounded by the request context.

**Config plumbing** (`pkg/field`, `pkg/cli`): a new `http-response-header-timeout-seconds` field (default 60, `0` disables) is injected into the run context and consumed by the transport. An explicit `WithResponseHeaderTimeout` option takes precedence over the config value, mirroring the existing `http-timeout-seconds` behavior.

## ⚠️ One behavior change

`WithMaxConnectionRetries` **defaults to 3** — previously there was no retry. This is the intended fix, and it's scoped tightly (idempotent + rewindable + reconnectable-only, context-bounded). To opt out, pass `WithMaxConnectionRetries(0)`. Happy to flip the default to opt-in (0) if reviewers prefer.

## Note

The retry sits at the transport layer, below `BaseHttpClient.Do`'s rate limiter, so a retried attempt does not take a fresh rate-limit token. For rare connection-death retries this is negligible; called out for visibility.

## Testing

`go test ./pkg/uhttp/... ./pkg/field/...` and `golangci-lint` pass. New tests cover error classification, retry-until-success, stop-at-max, non-reconnectable/non-idempotent/non-rewindable guards, body rewind on retry, context cancellation, and option-vs-config precedence.

### DIFF
--- a/pkg/cli/commands.go
+++ b/pkg/cli/commands.go
@@ -433,6 +433,13 @@ func MakeMainCommand[T field.Configurable](
 		}
 		runCtx = context.WithValue(runCtx, uhttp.ContextHTTPTimeoutKey, time.Duration(httpTimeout)*time.Second)
 
+		responseHeaderTimeout := v.GetInt(field.HttpResponseHeaderTimeoutField.GetName())
+		responseHeaderTimeoutField := field.HttpResponseHeaderTimeoutField
+		if _, err := field.ValidateField(&responseHeaderTimeoutField, responseHeaderTimeout); err != nil {
+			return err
+		}
+		runCtx = context.WithValue(runCtx, uhttp.ContextHTTPResponseHeaderTimeoutKey, time.Duration(responseHeaderTimeout)*time.Second)
+
 		storageEngine := v.GetString(field.StorageEngineField.GetName())
 		storageEngineField := field.StorageEngineField
 		if _, err := field.ValidateField(&storageEngineField, storageEngine); err != nil {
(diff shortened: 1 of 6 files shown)


kans:
Sol 5.6 says:

Cancellation during retry backoff returns the prior network error instead of context.Canceled/DeadlineExceeded, misclassifying canceled requests as retryable Unavailable errors. Fix both cancellation paths in pkg/uhttp/transport.go.

ggreer (approved):

PR #1005, opened 2026-07-11
uhttp: drop stale pooled connections and add a bounded, safe retry

jugonzalez12:
Hosted connectors run on Lambda behind an egress proxy. Lambda freezes pause the monotonic clock, so IdleConnTimeout never observes the gap while the proxy and origins time out pooled connections in real time. The first requests after a thaw (or after a proxy instance replacement) then fail with "connection reset by peer", "unexpected EOF", or "http2: client connection lost", burning task-level retries.

Three changes to the uhttp Transport:

- Wall-clock gap detection: track last-activity time in wall-clock nanos (the wall clock is resynced on thaw; time.Since would measure the frozen monotonic clock and report no gap). On the first request after a 60s+ gap, close idle connections on both the HTTP/1.1 and HTTP/2 pools so requests dial fresh instead of reusing presumed-dead connections.

- cycle() now closes the pool it abandons when rebuilding the round tripper, instead of leaving connections to linger until their idle timers fire.

- A single bounded retry in RoundTrip for two provably-safe classes: dial-phase failures (dial/proxyconnect ops — no request bytes were written, safe for every method), and stale-connection errors (ECONNRESET, unexpected EOF, h2 connection lost) for requests that declare idempotence via a safe method or an Idempotency-Key header. net/http already retries the second class on HTTP/1.1 but never on HTTP/2, where its retry gate requires a reused persistConn and h2 connections are always wrapped in fresh ones. Retries require a rewindable body, and the idle pool is dropped before a stale-connection retry so the second attempt cannot draw another dead connection.

Includes connection-counting integration tests for the gap drop, the cycle close, and the pool drop before retry, plus unit tests for the retry gating (dial-phase any-method, declared-idempotent POST, GET, undeclared POST refused, non-rewindable body refused).

### DIFF
--- a/pkg/uhttp/errors.go
+++ b/pkg/uhttp/errors.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"net/url"
 	"strings"
 	"syscall"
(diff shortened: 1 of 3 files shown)


jugonzalez12:
Running a sync of the baton GCP connector locally through squid with this change (and interrupting it during resource and grant sync) behaves nicely.

kans:
Fable says:

```
Swap publishes a fresh lastActivityNs before the pool is cleared. After a freeze-sized gap, concurrent callers behave like this:

Goroutine A: Swap sees a large gap → will close idles.
Goroutine B: Swap sees A’s timestamp → gap looks small → skips close.
B reaches RoundTrip and can take a still-pooled dead connection before A’s CloseIdleConnections runs.
Consequence: the wall-clock drop does not actually protect parallel first-use-after-gap traffic. Idempotent calls may still recover via the one retry; undeclared POST/PATCH/etc. fail with no retry (TestTransportDoesNotRetryStaleConnForUndeclaredPOST encodes that policy).

The intended “only one closer” pattern needs the close (or a generation/lock) to happen before other goroutines are allowed to treat the pool as live—not an optimistic Swap that suppresses peer closes.

Retry path (read as correct)
requestNeverSent only matches net.OpError ops dial / proxyconnect (stdlib uses these only while establishing the connection/CONNECT). Safe for any method.
Stale retries require declaredIdempotent (safe methods or idempotency headers), matching Request.isReplayable.
Body replay requires GetBody after the transport closes the body on failure; otherwise no retry.
Classification runs on the raw transport error; wrapTransientNetworkError runs only after the retry attempt.
At most one retry; context cancellation/deadline suppresses retry via req.Context().Err().
Narrow follow-on of the same race
On stale retry, closeIdleConnections() closes t.baseTransport, while the retry uses the rt returned earlier from cycle(). If another goroutine cycles the transport in between, those can diverge (close the new pool, retry on the old). Same concurrency window as above; uncommon (5‑minute cycle) but the same structural issue.

Bottom line: Retry safety looks sound. The idle-drop fix is not sound under concurrent RoundTrip after a gap—the Swap-then-close ordering is the defect.
```

kans (approved):

PR #1007, opened 2026-07-14
uhttp: drop pooled proxy tunnels after a wall-clock jump (Lambda freeze)

arreyder:
## Problem

Hosted connectors run as Lambdas whose outbound HTTPS rides CONNECT tunnels through the egress proxy (HTTPS_PROXY → Squid behind an internal NLB, :3128). Between invocations Lambda **freezes the execution environment and the monotonic clock stops**. `http.Transport.IdleConnTimeout` ages pooled connections on the monotonic clock, so a tunnel that sat frozen for ten real-world minutes looks *zero seconds old* to the pool — while Squid, the NLB's fixed 350s idle timeout, or the origin has already closed the socket.

On thaw the first request reuses the dead tunnel and either burns a task attempt on `connection reset by peer` / `http2: client connection lost`, or stalls until the h2 ping (15s) / `ResponseHeaderTimeout` (60s) notices. The existing `IdleConnTimeout: 30s` and the 5-minute transport re-`cycle()` can't help — both measure monotonic time, which is exactly what stops during a freeze.

## Evidence (prod-usw2, 2026-07-14)

- Proxy NLB `net/proxy-nlb-ecdfd6e`, 24h: **3.34M new flows vs 851k target-side TCP RSTs (25.5%)**; NLB-generated RSTs only 266. Reset rate is flat around the clock — a client-behavior signature, not overload (Squid fleet at 2.3% CPU, 0 unhealthy targets).
- `be-temporal-sync` 48h: ~4k `http2: timeout awaiting response headers`, 1.5k `request timeout`, 1.3k `context canceled`; smoking gun `run-sync: error getting connector metadata: … http2 client connection lost` — the *first* request of a sync on a reused dead tunnel.

Root-cause analysis: ductone CE-837 + [Slack thread](https://conductorone.slack.com/archives/C09LELZDTNJ/p1783711823108819).

## Fix

Track last-request activity on the wall clock (monotonic stripped via `Round(0)`, since `time.Sub` silently prefers the monotonic reading, which reads ~0 across a freeze — `CLOCK_REALTIME` is resynced on thaw). When the gap exceeds 15s, `CloseIdleConnections()` on the base transport before dispatching. This subsumes "flush the pool at invocation start": the first request of a thawed invocation always sees the gap. A false positive costs one CONNECT+TLS handshake against syncs that make thousands of calls.

## Test

`TestTransportDropsPooledConnsAfterWallClockJump` drives a real `httptest` server, confirms the connection is pooled/reused with no gap, then simulates a freeze by backdating the wall-clock marker and asserts a new connection is dialed.

## Verification after rollout

`aws.networkelb.tcptarget_reset_count{loadbalancer:net/proxy-nlb-ecdfd6e*}` should fall from ~35k/hr, and `@error:*awaiting*` counts in `be-temporal-sync` for the top affected tenants should drop sharply.

### DIFF
--- a/pkg/uhttp/transport.go
+++ b/pkg/uhttp/transport.go
@@ -36,6 +36,14 @@ var loggedResponseHeaders = []string{
 	"Retry-After",                // Often returned with 429
 }
 
+// wallClockGapThreshold is the wall-clock jump between requests treated as
+// evidence the process was suspended (e.g. a Lambda freeze between
+// invocations): the proxy/NLB has likely idle-closed our pooled connections
+// while IdleConnTimeout's monotonic clock was stopped, so drop the pool
+// rather than reuse a dead socket. A false positive only costs one
+// CONNECT+TLS handshake.
+const wallClockGapThreshold = 15 * time.Second
+
 // NewTransport creates a new Transport, applies the options, and then cycles the transport.
 func NewTransport(ctx context.Context, options ...Option) (*Transport, error) {
 	t := newTransport()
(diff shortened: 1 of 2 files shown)


kans:
The fix landed in b54ed673 (#1005).

PR #1012, opened 2026-07-16
[CXP-778] lambda: default the v2 wire frame to off, gate it behind opt-in

luisina-santos:
## Summary
- `pkg/lambda/grpc/wire.go`'s v2 wire frame (#1000, shipped in v0.18.2) dual-encodes every outbound lambda request by splicing top-level `v`/`frame` JSON fields onto the legacy protojson payload, assuming every peer discards unknown fields.
- That assumption doesn't hold for connectors still vendoring an SDK from before `DiscardUnknown` was added (#958, pre `cbcf7dcc`): their `Request.UnmarshalJSON` does a strict `protojson.Unmarshal` and fails outright (`unknown field "v"`) on the very first RPC of every sync (`GetMetadata`).
- This surfaced in production when `ductone/c1`'s baton-sdk bump (PR #21237) crossed v0.18.2 for the first time and started sending the extra fields to every lambda-hosted connector, including ones on pre-#958 SDKs.

## Fix
- Added `WithWireFrames(enabled bool)` as an explicit opt-in `LambdaTransportOption` on `NewLambdaClientTransport`; it defaults to `false`.
- `RoundTrip` now marshals through the new `Request.marshalLegacy()` by default — plain `protojson.Marshal`, no `v`/`frame` fields — which is byte-for-byte the pre-frame wire shape every deployed connector can decode.
- The dual-encoded/annotation-lossless frame path is unchanged and still available, but only sent once a caller explicitly opts in for connectors confirmed to run an SDK new enough to tolerate it.
- This restores compatibility purely by bumping the SDK dependency — no code change required on the invoker side (`ductone/c1`), since it doesn't currently pass `WithWireFrames`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/lambda/...` (74 passed)
- [x] Added `TestNewLambdaClientTransport_WireFramesDefaultOff` — confirms the transport defaults to frames disabled and `WithWireFrames(true)` flips it.
- [x] Added `TestRequest_MarshalLegacy_NoFrameFields` — confirms the default payload has no `v`/`frame` fields and round-trips through strict (non-`DiscardUnknown`) `protojson.Unmarshal`, simulating a pre-#958 connector.

### DIFF
--- a/pkg/lambda/grpc/client.go
+++ b/pkg/lambda/grpc/client.go
@@ -23,20 +23,51 @@ import (
 type lambdaTransport struct {
 	lambdaClient *lambda.Client
 	functionName string
+
+	// wireFramesEnabled gates the v2 dual-encoded wire frame (see wire.go).
+	// It defaults to false: some deployed lambda connectors predate the
+	// DiscardUnknown behavior the frame relies on and unmarshal transport
+	// JSON with strict protojson, so an unrecognized top-level "v"/"frame"
+	// pair fails their request decode outright rather than being ignored.
+	// Callers that know every target connector is on an SDK new enough to
+	// tolerate (or read) the frame can opt in with WithWireFrames.
+	wireFramesEnabled bool
+}
+
+// LambdaTransportOption configures a lambdaTransport constructed by
+// NewLambdaClientTransport.
+type LambdaTransportOption func(*lambdaTransport)
+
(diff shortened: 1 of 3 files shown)


luisina-santos:
we don't want to patch this issue. Already fixed by updating connectors versions

PR #1014, opened 2026-07-16
filter safe dangling grants

kans:
## Summary
During full syncs, drop entitlements and grants that reference resource types not present in the current sync’s store (never listed / not selected via WithSyncResourceTypes). That matches what uplift already does: unknown principal/entitlement types are Debug-skipped and never become platform AEUB/AERB rows, so keeping them in the c1z is dead data.

Probe scheduled types via store GetResourceType (cached across workers); resume-safe because the store keyspace already reflects the run’s selection.
- Full syncs only; partial syncs are unchanged.
- External-match grants keep placeholder principals (rewritten after the external phase copies types/resources).
- InsertResourceGrants: still insert grant-discovered resources of scheduled types even when the discovering grant is dropped; do not insert resources (or keep grants) for unscheduled types — uplift can’t use either.
- Narrow GrantExpandable.ResourceTypeIds to scheduled types; remove the annotation if the intersection is empty (empty list would mean unfiltered expansion).
- Run the filter before exclusion-group validation so dropped entitlements can’t fail the sync.
- One aggregate warn at end of sync with drop counters.

### DIFF
--- a/pkg/sync/ingest_filter.go
+++ b/pkg/sync/ingest_filter.go
@@ -1,0 +1,237 @@
+package sync //nolint:revive,nolintlint // we can't change the package name for backwards compatibility
+
+// Fresh connector-page filtering for references into resource types that were
+// not scheduled for the current full sync. This stays in the sync engine:
+// hosted connector calls are stateless, while the engine has already stored
+// the selected resource-type set before entitlement and grant collection.
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


ggreer (approved):

pkg/sync/ingest_filter.go
```diff
@@ -0,0 +1,292 @@
...
+// are transformed by the later external-resource matching phase.
+func (s *syncer) filterGrantExpansionTypes(
+	ctx context.Context,
+	grant *v2.Grant,
+	annos annotations.Annotations,
+) (*v2.Grant, error) {
+	if grant == nil || hasExternalResourceMatch(annos) {
+		return grant, nil
+	}
+	expandable := &v2.GrantExpandable{}
```
ggreer, line 154: ~You can use the hasGrantExpandable function here.~ Actually it's slightly faster to do it this way because you don't have to pick the annotation twice.

PR #1015, opened 2026-07-17
PHASE 1: Fault injection harness for pebble

kans:
## Summary

First slice of the sync-replay stack: fault-injection infrastructure for the Pebble engine, landed ahead of the upcoming write-path refactors so every later change merges under an IO-failure test gate.

### `WithVFS(vfs.FS)` engine option

The engine (and the `pebble.DB` under it) can now run over a caller-supplied filesystem. This required threading the FS through every path that stages SST files the DB later reads via `Ingest`/`IngestAndExcise` — the deferred index build, digest build, bulk import, synth-grant layer, id-index migration, all spill-chunk merges — plus staging-dir setup/teardown, the read-only checkpoint clone, and checkpoint WAL truncation. Spill-chunk scratch deliberately stays on plain OS IO (only engine code ever reads it back).

**Production callers are unchanged**: nil FS leaves `pebble.Options.FS` untouched (pebble's `EnsureDefaults` applies its usual disk-health-wrapped `vfs.Default`), and engine staging uses `vfs.Default` exactly as before. No per-row or per-batch path changed; the added work is a handful of syscalls per staging build / checkpoint.

### Fault-injection sweep (`errorfs` + crashable MemFS)

The contract under test: **after any surfaced write/fsync failure followed by a hard discard (crash image keeps only fsynced state; no clean close), the store must reopen either UNFINISHED-and-resume-to-completion, or FINISHED-and-content-complete.** Anything else — a finished-but-incomplete artifact, an unfinished-but-unresumable store — is a durability-ordering bug.

- `TestErrorFSEndSyncWindowSweep` (default CI, ~1s): exhaustive — fails the k-th write op **and every op after it** across the EndSync window, crash-clones at the failure point (injector still armed, so nothing can heal the image), and verifies the reopen dichotomy. Self-terminating when an armed run injects nothing, which proves full window coverage. The sweep also **self-asserts** its coverage: every k must inject, and both dichotomy arms must be populated (currently 34 points: 32 land unfinished-and-resume, 3 land finished-complete).
- `TestErrorFSWholeSyncRandomSweepSoak` (`BATON_SOAK`-gated): seeded random failure points across open + pages, with page-replay resume.
- The content oracle asserts **exact row-set equality**, not counts: the precise grant external-id set through the primary keyspace, and per principal the precise entitlement set through the `by_principal` index, with each index-served record's refs checked against the queried principal.
- **Validated by mutation testing**: seeding a stamp-before-index-build ordering bug fails the sweep immediately with the lying-artifact signature (`finished image: by_principal must serve alice`).

Harness design notes: pebble escalates WAL-commit/MANIFEST failures to `Logger.Fatalf` assuming process exit, so injected engines get a test-only logger that parks the goroutine and signals the harness (the horked engine is leaked, like a crashed process); non-injected engines get a fail-fast panicking logger. One documented coverage gap: IO failures inside an EndSync whose pages are still memtable-resident ride pebble's flushable-ingest path, where failures are raw panics by design — in production that's simply a process crash, whose recovery path *is* covered; modeling the failure itself would need subprocess isolation.

### Regression guards for split-brain IO

Review of this PR found three instances of one bug class: a file created through one filesystem and read/renamed/truncated through the other. Two guards now pin the class:

- `os_io_allowlist_test.go`: AST meta-test — every direct `os` file-IO call in the package must be registered with a reason; unregistered calls, stale registry entries, aliased imports, and dot-imports all fail.
- `memfs_lifecycle_test.go`: the full engine lifecycle (sync with deferred-index staging → EndSync → writable checkpoint → checkpoint reopen → primary reopen) over a **pure** in-memory FS, which fails loudly on any host-IO regression the DB can observe. This test reproduces all three review findings if their fixes are reverted.

### Notable fix that outlived review

`truncateCheckpointWALs` now builds each zero-byte WAL replacement at a side name and atomically renames it over the original — `vfs.Default.Create` is remove-then-recreate (not `O_TRUNC`), so creating the WAL path directly could have *deleted* a WAL on a failed recreate, the exact discovered-file-set hazard the truncate-not-delete policy exists to avoid.

## Review

Three adversarial review passes across two model families; five findings, all fixed and re-verified (disarm-before-clone image healing, checkpoint WAL truncation on the wrong FS, id-index SST rename on the wrong FS, hang-instead-of-fail on verify engines, allowlist import-alias bypass). Final targeted pass on the last delta came back clean.

## Test plan

- [x] Full suite green
- [x] Engine package under `-race`
- [x] Both sweeps incl. `BATON_SOAK=1` (60 seeds)
- [x] Mutation validation (seeded ordering bug caught at k=2, reverted)
- [x] `golangci-lint` clean; `GOOS=windows go vet` clean (MemFS-backed tests skip on Windows; meta-test enforces there)

### DIFF
--- a/pkg/dotc1z/engine/pebble/engine.go
+++ b/pkg/dotc1z/engine/pebble/engine.go
@@ -821,10 +821,16 @@ func copyReadOnlyDBDir(pfs vfs.FS, srcDir, destDir string) error {
 //
 // pfs is the engine FS (Engine.fs()) — db.Checkpoint wrote the
 // checkpoint through it, so the WALs to truncate live there, not
-// necessarily on the host filesystem. vfs has no Truncate, so the
-// truncation is a Create (same open-existing-with-O_TRUNC semantics
-// as os.Create on the default FS; permissions of an existing file are
-// untouched by O_CREATE|O_TRUNC).
+// necessarily on the host filesystem. vfs has no Truncate, and
+// vfs.Default.Create is REMOVE-then-recreate (not O_TRUNC — see its
+// hard-link rationale), so "Create the WAL path directly" could
+// unlink a WAL and then fail the recreate, which is exactly the
+// delete-changes-the-discovered-file-set hazard the truncate-not-
+// delete policy above exists to avoid. Instead the zero-byte
+// replacement is built at a side name and Rename'd over the WAL:
+// the original survives every failure before the rename, and the
+// rename replaces the path atomically on both the default FS and
+// MemFS.
 func truncateCheckpointWALs(pfs vfs.FS, destDir string) error {
 	names, err := pfs.List(destDir)
 	if err != nil {
@@ -841,11 +847,17 @@ func truncateCheckpointWALs(pfs vfs.FS, destDir string) error {
 			}
 			continue
 		}
-		f, err := pfs.Create(path, vfs.WriteCategoryUnspecified)
+		tmp := path + ".trunc"
+		f, err := pfs.Create(tmp, vfs.WriteCategoryUnspecified)
 		if err != nil {
 			return err
 		}
 		if err := f.Close(); err != nil {
+			_ = pfs.Remove(tmp)
+			return err
+		}
+		if err := pfs.Rename(tmp, path); err != nil {
+			_ = pfs.Remove(tmp)
 			return err
 		}
 	}
--- a/pkg/dotc1z/engine/pebble/engine.go
+++ b/pkg/dotc1z/engine/pebble/engine.go
@@ -670,6 +670,41 @@ func (e *Engine) InvalidateBareIDLookups() { e.noteEntitlementKeyspaceWrite() }
 // migration (see migratedOnOpen).
 func (e *Engine) MigratedOnOpen() bool { return e.migratedOnOpen }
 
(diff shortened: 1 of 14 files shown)


pkg/dotc1z/engine/pebble/engine.go
```diff
@@ -670,6 +670,46 @@ func (e *Engine) InvalidateBareIDLookups() { e.noteEntitlementKeyspaceWrite() }
...
+// created through e.fs() (the pebble.DB resolves ingest paths through
+// that FS), while spill-chunk scratch is plain OS IO — so the directory
+// must exist on both. On the default FS the MkdirAll is a no-op.
+//
+// Portability note: mirroring a host temp path onto a MemFS assumes
+// "/" separators (MemFS only splits on "/"). WithVFS with a MemFS is a
+// test-only configuration and the tests are unix-only; a Windows port
+// would need to stage under fs.PathJoin'd relative paths instead.
+func (e *Engine) prepareStagingDir(tmpDir, pattern string) (string, error) {
+	dir, err := os.MkdirTemp(tmpDir, pattern)
```
ggreer, line 695: It would be nice if we could use the same vfs for everything, but I understand why the scratch stuff shouldn't go in the pebble dir.

ggreer (approved):

PR #1018, opened 2026-07-17
deturducken pebble engine

kans:
## Simplify the pebble storage stack: fewer layers, one sync-lifecycle state, same guarantees

Refactor only — no behavior change (two small, deliberate exceptions listed below).

### Problem

Writing a record traversed five layers, with glue between them that existed for
structural reasons rather than functional ones:

```
pebbleStore → Adapter → Engine → rawdb → pebble
```

- `internal/rawdb` (the write choke point) couldn't import its parent package, so the
  engine injected nine derivation closures into it at Open, plus an adapter struct to
  carry them.
- The compactor reached the DB through a separate handle type (`Engine.DB()` returning a
  narrowed view struct, with type aliases so other packages could name it).
- The `Adapter` type existed as a separate layer even though only `pebbleStore`
  constructed one, and it kept its own copy of "which sync is open" (id, type, step
  token, …) shadowing the engine's binding — with rehydration logic to keep the two in
  agreement, and real bugs when they drifted (a stale cached step token silently
  restarted the syncer's FSM from scratch).

### Change

```
pebbleStore → Engine → rawdb → pebble
```

- **rawdb owns the key formats it enforces.** The keyspace ABI (key splices, value
  scanners, digest-invalidation key builders, type/index discriminator bytes) moved into
  rawdb (`keyspace.go`). Its typed staging ops now derive their obligations with direct
  function calls; the injected-closure machinery (`RecordDerivers`, `SetRecordDerivers`,
  `recordStager`) is deleted.
- **rawdb owns the write-side crash state.** The deferred-index marker (in-memory flag +
  durable key, with the arm/clear ordering contracts) and the digests-present gate moved
  onto `rawdb.DB`, next to the staging ops that consume them.
- **The compactor handle is gone.** The operations the compactor uses are `Engine`
  methods now (`merge_surface.go`), each checking engine lifecycle per call — using a
  retained reference after Close yields a clean `ErrEngineClosing` (or an explicit,
  messaged panic for the two ops with no error path) instead of a nil dereference.
  A meta-test fences these ops' production use to the compactor package
  (mutation-verified: planting a call in engine production code fails CI).
- **The Adapter layer is dissolved into the Engine.** Same files, same method bodies —
  the receiver changed. `type Adapter = Engine` and `NewAdapter` remain as aliases so
  existing code and tests compile. `pebbleStore` embeds `*pebble.Engine` directly.
- **One sync-lifecycle state.** The engine's sync binding is the only in-memory
  lifecycle state; everything else (step token, type, parent) is read from the durable
  `SyncRunRecord` on demand — the same model the SQLite engine has always used, which is
  why it never needed the cache-rehydration logic this deletes.

### What did not change

The compile-time fence around the raw `*pebble.DB`, the typed staging API and its
obligation derivation, write barriers and seal semantics, all on-disk key formats
(byte-identical), and performance (allocation counts identical before/after; the deleted
indirection was compile-time-inlined all along).

### Deliberate behavior deltas

- `CurrentDBSizeBytes` walks the engine's VFS instead of the host filesystem (required
  for VFS-injected engines), tolerates files deleted mid-walk by compaction, and does
  not follow symlinks.
- `Engine.SetCurrentSync` gained a `ctx` parameter (the validating store-facing form);
  the bare binding is now unexported.

### Testing

Full engine/store/compactor/equivalence suites green (sqlite paths included); fault-
injection sweeps and soak green; `-race` battery green across every touched package
(the one failure, `TestC1ZConcurrentClose`, is a pre-existing sqlite race that
reproduces on main and is tracked separately); lint clean. Two independent adversarial
model reviews on the rawdb half confirmed key-format byte-fidelity and invalidation
equivalence; their findings (symlink traversal, a weakened meta-test, closed-engine nil
panics) are fixed in this branch.

### DIFF
--- a/pkg/dotc1z/engine/pebble/adapter.go
+++ b/pkg/dotc1z/engine/pebble/adapter.go
@@ -21,6 +21,7 @@ import (
 	v3 "github.com/conductorone/baton-sdk/pb/c1/storage/v3"
 	"github.com/conductorone/baton-sdk/pkg/connectorstore"
 	"github.com/conductorone/baton-sdk/pkg/dotc1z/c1zstore"
+	"github.com/conductorone/baton-sdk/pkg/dotc1z/engine/pebble/internal/keys"
 )
 
 // Adapter wraps an *Engine and implements connectorstore.Writer
@@ -810,7 +811,7 @@ func (a *Adapter) ListResources(ctx context.Context, req *v2.ResourcesServiceLis
 	// them on the next call.
 	cursorFor := func(rec *v3.ResourceRecord) string {
 		if useParent {
-			return encodeCursor(encodeResourceByParentIndexKey(
+			return encodeCursor(keys.EncodeResourceByParentIndexKey(
 				parent.GetResourceType(), parent.GetResource(),
 				rec.GetResourceTypeId(), rec.GetResourceId(),
 			))
--- a/pkg/dotc1z/engine/pebble/adapter.go
+++ b/pkg/dotc1z/engine/pebble/adapter.go
@@ -36,8 +36,20 @@ import (
 // reads). Adapter survives as a type alias for compatibility.
 
 // Adapter is the Engine, by its historical name. The wrapping type
-// dissolved in PR 2.6; the alias keeps the many existing constructors
-// and fixtures compiling. New code should use *Engine directly.
+// dissolved when the layer collapsed; the alias keeps existing
+// constructors and fixtures compiling. New code should use *Engine
+// directly.
+//
+// DELIBERATE BREAK vs the old wrapper: a bare Adapter/Engine is NOT a
+// connectorstore.Writer anymore. The old type carried Close(ctx); the
+// Engine's teardown is Close() (no ctx), and a Go alias cannot carry
+// both signatures. The supported Writer — before and after — is the
+// pkg/dotc1z store (dotc1z.NewStore with the Pebble engine), whose
+// Close(ctx) owns the envelope save the bare engine never performed;
+// a bare handle "closing" a c1z without saving it was a trap, not a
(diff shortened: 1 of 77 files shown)


pkg/dotc1z/engine/pebble/adapter.go
```diff
@@ -21,97 +21,81 @@ import (
...
+type Adapter = Engine
 
-// NewAdapter wraps an Engine. The engine must remain alive for the
-// lifetime of the adapter.
-func NewAdapter(e *Engine) *Adapter {
-	return &Adapter{engine: e}
-}
+// NewAdapter is a compatibility constructor from the dissolved-layer
+// era: the "adapter" IS the engine now.
+func NewAdapter(e *Engine) *Adapter { return e }
```
ggreer, line 57: I'm guessing a subsequent PR will remove this.

pkg/dotc1z/engine/pebble/adapter.go
```diff
@@ -202,77 +170,76 @@ func (a *Adapter) ResumeSync(ctx context.Context, syncType connectorstore.SyncTy
...
 	}
-	return nil
-}
-
-// CurrentSyncStep returns the current sync's step string, or "".
-func (a *Adapter) CurrentSyncStep(ctx context.Context) (string, error) {
-	a.mu.Lock()
-	defer a.mu.Unlock()
-	return a.current.step, nil
+	return rec.GetSyncToken(), nil
```
ggreer, line 231: I forgot that CurrentSyncStep() is really CurrentSyncToken(). :(

ggreer (approved):
Feels like this should delete more code, but at least it's less struct nesting.

PR #1019, opened 2026-07-18
sync: omit the entitlement graph from checkpoint tokens

arreyder:
## Problem

`state.Marshal` serializes the **entire entitlement graph** into every checkpoint token (every ≥10s during a sync, including once per page while the graph is being *loaded*). For large tenants this is what has been OOM-killing `be-temporal-sync` in prod — twice in the last day, first at a 24 G limit (2026-07-17 23:06Z) and again at 30 G after the limit was raised (2026-07-18 02:04Z, image 121645).

The last pre-OOM heap profile (22s before the second kill) shows the marshal path owning roughly **11 GB of a 13.5 GB heap** for a live graph of only **~0.9 GB**:

```
state.Marshal                     6.4 GB cum (47%)
 ├ json mapEncoder.encode         5.8 GB   (the graph's maps; key-sorting churn)
 └ bytes.growSlice                5.3 GB   (the output buffer + doubling history)
snappy.Encode                     2.0 GB   (the blob compressed into the store)
pebble manual.New / rawalloc      3.5 GB   (the blob as a pebble large batch →
                                            its own flushable memtable, bypassing
                                            MemTableSize entirely)
EntitlementGraph.AddEdge          0.9 GB   (the actual live graph)
```

Every copy is O(graph) and several are live simultaneously, so one sufficiently large tenant checkpointing is enough to kill a worker — and since the OOM lands *mid-checkpoint*, the sync crash-loops: the checkpoint meant to preserve progress is the thing preventing completion.

## Change

Checkpoints no longer carry the graph. It is a projection of data already in the store — `loadEntitlementGraph` rebuilds it from `PendingExpansionPage` with **no connector calls**, and `PrepareExpansionReplayToken` already relies on whole expansions being re-runnable from the store on finished syncs. This extends that same replay property to interim checkpoints: a crash mid-expansion re-runs the load + expansion phases on resume instead of resuming them from the token.

Checkpoint memory becomes O(actions + stats) — effectively KBs — independent of tenant size.

### Escape hatch: `WithEntitlementGraphInCheckpoints`

Per review, the omission is configurable. `WithEntitlementGraphInCheckpoints(true)` restores inline-graph checkpoints, so if a tenant turns out to need cross-restart expansion progress (see Trade-off) that's a config change rather than a new SDK release. **Default is off** — the omission above is the default behavior.

When enabled, the graph and the expansion `PageToken` are serialized **together**. They have to move as a unit: the page token indexes a pagination the graph was accumulating, so keeping one without the other silently drops the edges from earlier pages. Enabled output is therefore byte-compatible with what pre-omission SDKs wrote.

The knob is a `stateOpt` on the concrete `state`, deliberately *not* part of the `State` interface — it's a process-level setting, not token content, so it can't survive an `Unmarshal` of a token written by some other process.

### Resume-safety detail

A resumed reader gets no graph, so an in-flight `SyncGrantExpansionOp` action's `PageToken` is blanked **in the serialized copy** (live state is untouched): resuming that pagination against a fresh graph would silently drop the edges from earlier pages. Doing this at *marshal* time keeps graph-less tokens safe for **older readers too** (version rollback mid-sync ⇒ the old SDK sees a fresh graph + empty page token ⇒ clean full reload, not silent corruption). `Unmarshal` additionally normalizes orphan page tokens defensively.

Note the load phase leaves the final page token on the action even after `Loaded=true`, so the blanking applies uniformly across phases.

### Compatibility

| writer → reader | behavior |
|---|---|
| old token (graph inline) → new SDK | decodes + resumes exactly as before (field kept for decode; covered by test) |
| new token → new SDK | fresh graph, load restarts from page one |
| new token → old SDK (rollback) | same clean restart — this is why blanking happens at marshal time |
| opt-in token (graph inline) → either SDK | resumes the load where it left off; identical to pre-omission output |

Nothing outside `pkg/sync` consumes the token's graph: `ClearEntitlementGraph` already stripped it from every *finished* sync's token, and c1 passes `SyncToken` through opaquely.

## Trade-off

Crash/restart mid-expansion now redoes the graph load + expansion (store-local reads and idempotent `StoreExpandedGrants` upserts, no connector traffic) — minutes of local recompute in the rare-restart case, versus a guaranteed OOM-loop for large tenants today. Small tenants redo little; large tenants are exactly the ones that couldn't get through a checkpoint at all.

Stated plainly: **by default, cross-restart expansion progress is zero.** A tenant whose *entire* expansion exceeds one worker/activity lifetime would livelock on repeated restarts, where before it could inch forward page by page (though for exactly those tenants the checkpoint itself was what OOM'd the worker, so the old "progress" was moot in practice).

Note the two failure modes trade off directly against each other: the tenants big enough to need cross-restart progress are the same ones whose graph is expensive enough to encode that checkpointing may OOM. So `WithEntitlementGraphInCheckpoints` is an escape hatch for diagnosis and recovery, not a general answer — if the livelock is ever observed, the real follow-up is still a durable expansion progress marker persisted in the store, not re-inflating the token.

## Testing

- `TestSyncerTokenOmitsEntitlementGraph` — token carries neither graph nor in-flight expansion pagination; live state unmutated; resumed state restarts the load; non-expansion actions keep their pagination and no map entries are dropped
- `TestSyncerTokenLegacyInlineGraphStillDecodes` — old-style tokens restore the graph and keep their pagination; the next checkpoint after a legacy resume writes the new format (graph omitted, expansion pagination blanked) while the live state is unaffected
- `TestSyncerTokenIncludesEntitlementGraphWhenEnabled` — with the option on, graph *and* expansion pagination are both serialized, a resume continues the load with graph contents intact, and `newState()` still defaults to off
- `TestWithEntitlementGraphInCheckpointsReachesState` — pins the `SyncOpt` → syncer field → state plumbing so the knob can't silently go inert
- `TestSyncerTokenUnmarshalBlanksOrphanExpansionPageToken` — defensive normalization (mixed action map)
- `TestSyncerTokenV0OrphanExpansionPageTokenBlanked` — the defensive blanking also covers tokens decoded through the V0 fallback
- Full `./pkg/sync/...` suite green (incl. expand's topological-merge interrupt/resume tests, pebble fastpath, sqlite parity); token tests pass with `-race`

A syncer-level "interrupt exactly mid-expansion, resume, assert grant parity" e2e would need a new test hook (the existing interruption harness is connector-failure-based, and expansion makes no connector calls) — happy to add one as a follow-up if you want the hook.


ggreer:
This one makes me nervous because if for some reason grant expansion does take a while (or we reduce how long activities run for in temporal), we'll never make progress. As a baby step we could make this configurable via a syncer option. That way if we do run into issues, we don't have to tag a new baton-sdk release to fix it.

kans:
I think the right approach is to checkpoint an expansion cursor, not the whole graph.  The graph itself is fully deterministic.  We can calculate it once, and commit that (once), or we can recalculate it when we need it.  We shouldn't lose progress in expansion though.

edit: its cheap to reproduce the graph.  This is better than the altertnative.

kans (approved):

PR #1020, opened 2026-07-20
fix(pebble): hydrate entitlement resource annotations on read

mstanbCO:
## Summary

The Pebble c1z storage engine drops **resource-level annotations from the resource embedded in an entitlement** on read. Anything that reads connector metadata off `entitlement.Resource` — most importantly the `RawId` annotation — silently breaks when a connector moves from the SQLite engine to Pebble. This PR rehydrates the embedded resource in all four entitlement reader paths so the two engines return identical shapes.

## Background: how the two engines differ

Connectors emit an `Entitlement` carrying its **full** `Resource` (id **+ annotations** + profile). For example baton-okta builds the group member entitlement from the fully-annotated group resource, which carries `RawId{Id: <group id>}`.

- **SQLite engine**: stores the entire `v2.Entitlement` proto as a blob and unmarshals it verbatim on read → `entitlement.Resource.Annotations` (incl. `RawId`) is intact.
- **Pebble engine**: normalizes storage. On write (`V2EntitlementToV3`) the embedded resource is reduced to an id-only `v3.ResourceRef{ResourceTypeId, ResourceId}`; on read (`V3EntitlementToV2`) it is rebuilt as an **identity-only stub** — id present, **annotations gone**. This is called out in `V3EntitlementToV2`'s own doc comment ("The Resource side is hydrated as a stub (identity only); callers that need the full v2 Resource must hydrate via the engine's GetResourceRecord.").

Both types implement the same `reader_v2` service interface (`type Adapter = Engine`), so consumers written against the SQLite behavior — which never required separate hydration — get a silently degraded object from Pebble.

## Impact (how this surfaced)

In ConductorOne's uplift, the entitlement's embedded-resource `RawId` is the source for two things:

1. **`raw_baton_id`** on connector-sourced entitlements — derived from `source.Entitlement.GetResource().GetAnnotations()`.
2. **`match_baton_id` placeholder merging** — a manually-created "waiting to be merged" entitlement is matched to its connector entitlement by comparing `match_baton_id` against the same embedded-resource `RawId`.

When a tenant's connector was switched to the Pebble engine (rollout gated by a feature flag), that `RawId` disappeared from entitlements. Observed symptoms:

- `raw_baton_id` empty on newly-synced connector entitlements.
- `match_baton_id` placeholders never merged — every sync created a **duplicate** connector-sourced entitlement alongside the orphaned placeholder.
- **Resource-level** `match_baton_id` merges kept working, because standalone `ResourceRecord`s retain their annotations (`V3ResourceToV2` copies `Annotations`) — which is exactly the asymmetry that made this confusing to diagnose.
- `entitlement.Resource.GetId()` still resolved (id survives), so the entitlement was still processed — it just missed the merge and got created fresh.

## Fix

Add `(*Engine).hydrateEntitlementResource`, which replaces the identity-only stub with the full resource loaded from the resource record (via `GetResourceRecord` + `V3ResourceToV2`), and call it from every reader path that returns entitlements:

- `GetEntitlement`
- `ListEntitlements`
- `ListEntitlementsByIds`
- `StreamEntitlements`

Behavior:

- A resource **not present** under the sync (targeted/partial syncs can reference resources they didn't write) leaves the id-only stub in place — a missing resource must never turn into a failed entitlement read.
- The list/stream paths pass a per-call cache so entitlements sharing a resource incur a single lookup.

This restores parity with the SQLite engine: readers again see the full embedded resource (annotations included).

## Alternative considered

Persist the resource's annotations on the `EntitlementRecord` at write time (`V2EntitlementToV3`) so reads stay self-contained with no per-read lookup. That avoids the read-side `GetResourceRecord` cost but requires a storage-proto change and duplicates data. Read-side hydration was chosen as the smaller, engine-local fix that exactly matches SQLite semantics; happy to switch to the write-side approach if maintainers prefer it for the hot list/stream paths.

## Testing

- New `TestEntitlementResourceAnnotationsHydrated` writes a resource carrying a `RawId` annotation plus an entitlement on it, then asserts the `RawId` is present on the embedded resource returned by all four reader paths. (Without the fix, all four fail.)
- `go test ./pkg/dotc1z/engine/pebble/ -count=1` passes (incl. the SQLite-parity/differential suites).
- `go build ./...` clean.

## Note for downstream (c1)

A defense-in-depth guard on the consumer side is also worth doing: derive `RawId`/`raw_baton_id` and the match key from the standalone app resource that uplift already resolves, rather than the entitlement's embedded copy — so the merge is robust regardless of storage engine.

### DIFF
--- a/pkg/dotc1z/engine/pebble/adapter.go
+++ b/pkg/dotc1z/engine/pebble/adapter.go
@@ -888,8 +888,11 @@ func (e *Engine) ListEntitlements(ctx context.Context, req *v2.EntitlementsServi
 		return nil, c1zstore.AdaptNotFound(err, pebble.ErrNotFound)
 	}
 	out := make([]*v2.Entitlement, 0, len(records))
+	resCache := make(map[string]*v2.Resource)
 	for _, rec := range records {
-		out = append(out, V3EntitlementToV2(rec))
+		ent := V3EntitlementToV2(rec)
+		e.hydrateEntitlementResource(ctx, ent, resCache)
+		out = append(out, ent)
 	}
 	return v2.EntitlementsServiceListEntitlementsResponse_builder{
 		List:          out,
--- a/pkg/dotc1z/engine/pebble/adapter_reader.go
+++ b/pkg/dotc1z/engine/pebble/adapter_reader.go
@@ -30,6 +30,60 @@ var _ reader_v2.GrantsReaderServiceServer = (*Adapter)(nil)
 // the engine, then returns a v2-shaped response (via the translation
 // layer in translate_v2.go).
 
+// hydrateEntitlementResource swaps the identity-only Resource stub that
+// V3EntitlementToV2 produces for the full resource stored under the same
+// sync, so the Pebble reader returns entitlements shaped exactly like the
+// SQLite engine does.
+//
+// Connectors emit an entitlement carrying its full Resource (id +
+// annotations + profile). The SQLite engine round-tripped that whole proto,
+// so readers saw entitlement.Resource.Annotations intact. The Pebble engine
+// normalizes the entitlement's resource to an id-only ref on write
+// (V2EntitlementToV3) and rebuilds only the id on read, so downstream
+// consumers that read connector metadata off entitlement.Resource — most
+// importantly the connector RawId annotation, which uplift uses to populate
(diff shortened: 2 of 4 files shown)


kans:
We fixed this on the caller side.  We should drop the old v2 proto interfaces to ensure this doesn't happen again.

PR #1025, opened 2026-07-21
feat(dotc1z): surface discovered_at through reader

manojacs:
## What & why

c1z entries store a per-entry `discovered_at` — the time the connector actually observed a grant. It's persisted in the storage layer but **stripped when storage records convert to `c1.connector.v2` protos**: `pebble.V3GrantToV2` never copies `r.GetDiscoveredAt()`, and `v2.Grant` has no field to carry it. So c1 can't read it.

c1 needs it: its uplift deletes optimistic "overlay" grant records by comparing `overlay.created_at` against a single coarse per-sync timestamp, which can delete an overlay before the connector has actually re-observed the grant (→ a user loses access). The real fix compares against the per-entry `discovered_at`. This PR stops dropping it — it surfaces it through the reader.

## Mechanism chosen: **A — additive optional reader capability**

New interface `connectorstore.GrantDiscoveredAtReader`, in the **exact mold** of the existing `EntitlementGrantDigestReader` / `ExpansionGrantLister`:

```go
type GrantDiscoveredAtReader interface {
    GetGrantDiscoveredAt(ctx context.Context, grantID string) (discoveredAt *timestamppb.Timestamp, found bool, err error)
}
```

- **Type-asserted, Pebble-implemented.** Only the Pebble (v3) engine implements it; callers type-assert and fall back on failure.
- **Addressed by the same `grantID` `GetGrant` accepts** (stored external id, or the legacy reconstructed concat). The Pebble impl reuses `GetGrantRecord` and returns `rec.GetDiscoveredAt()`.
- **`v2.Grant` is untouched** → the `c1.connector.v2` API contract is unchanged, keeping the review surface narrow.

**Why A over B (adding a field to `v2.Grant`):** B is simpler for c1 to consume but mutates the connector API contract (proto field on a wire type shared by every connector) — a much wider blast radius for a value only c1's uplift needs. A is fully additive and mirrors how every other engine-specific read capability in this package is already exposed.

### Semantics
- Value is the **STORED** `discovered_at`, stamped once at first write and carried forward across re-syncs / expander rewrites — **never a fresh `time.Now()`**.
- `found=false` (nil error) = "no such grant"; an **ambiguous** concat id stays an error (never a guess), matching `GetGrant`.
- `found=true` with a **nil** timestamp = a grant stored without a stamp (legacy) → caller treats as "unknown", not epoch zero.

### Scope
- **Grants only** (the must-have). Entitlements also carry `discovered_at` in storage — a parallel `GetEntitlementDiscoveredAt` is a cheap follow-up if the "absence/coverage" signal ever needs it, deliberately left out here to keep the surface minimal.
- **SQLite (deprecated) is intentionally not implemented** — a failed type assertion is the caller's "unavailable, fall back" signal, exactly like `EntitlementGrantDigestReader`.

## Files touched
- `pkg/connectorstore/connectorstore.go` — new `GrantDiscoveredAtReader` interface + doc.
- `pkg/dotc1z/engine/pebble/adapter_reader.go` — Pebble `GetGrantDiscoveredAt` impl.
- `pkg/dotc1z/engine/pebble/adapter.go` — compile-time interface assertion.
- `pkg/dotc1z/engine/pebble/grant_discovered_at_reader_test.go` — exact-seed round-trip, unknown-id, nil-stamp.
- `pkg/dotc1z/grant_discovered_at_reader_test.go` — cross-engine: Pebble returns stored value (stable across reads); SQLite cleanly reports unavailable.

## Tests
```
go test ./pkg/connectorstore/... ./pkg/dotc1z/ ./pkg/dotc1z/engine/pebble/   # ok
```
All new tests pass; full `pkg/dotc1z` + Pebble suites green. Changed files are lint-clean.

### DIFF
--- a/pkg/dotc1z/engine/pebble/adapter_reader.go
+++ b/pkg/dotc1z/engine/pebble/adapter_reader.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/cockroachdb/pebble/v2"
 	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
 	reader_v2 "github.com/conductorone/baton-sdk/pb/c1/reader/v2"
@@ -683,6 +684,33 @@ func (e *Engine) digestEntitlementIdentity(ctx context.Context, ent *v2.Entitlem
 		return entitlementIdentity{}, false, err
 	}
 	return id, true, nil
+}
+
+// GetGrantDiscoveredAt implements connectorstore.GrantDiscoveredAtReader.
+// It returns the discovered_at stored on the grant addressed by grantID
+// — the value stamped when the grant was first written and preserved
+// across re-syncs and expander rewrites, NEVER a fresh now(). It reads
+// the same GrantRecord GetGrant serves (grant keys carry no sync scope,
+// so no active-sync binding is required) and returns the record's
+// DiscoveredAt verbatim.
+//
+// found is false when no grant resolves for grantID (unknown or
+// unresolvable id); an ambiguous concat id stays an error, matching the
+// GetGrant resolution contract — a lossy string never guesses which
+// grant to answer with. A found grant that was never stamped returns a
+// nil timestamp with found=true.
+func (e *Engine) GetGrantDiscoveredAt(ctx context.Context, grantID string) (*timestamppb.Timestamp, bool, error) {
+	rec, err := e.GetGrantRecord(ctx, grantID)
+	if err != nil {
+		if errors.Is(err, pebble.ErrNotFound) {
+			return nil, false, nil
+		}
+		return nil, false, err
+	}
+	if rec == nil {
+		return nil, false, nil
+	}
+	return rec.GetDiscoveredAt(), true, nil
 }
 
 // GetEntitlementGrantDigest implements connectorstore.EntitlementGrantDigestReader.
--- a/pkg/connectorstore/connectorstore.go
+++ b/pkg/connectorstore/connectorstore.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"io"
 
+	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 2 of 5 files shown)


pkg/dotc1z/engine/pebble/adapter_reader.go
```diff
@@ -685,6 +686,33 @@ func (e *Engine) digestEntitlementIdentity(ctx context.Context, ent *v2.Entitlem
 	return id, true, nil
 }
 
+// GetGrantDiscoveredAt implements connectorstore.GrantDiscoveredAtReader.
+// It returns the discovered_at stored on the grant addressed by grantID
+// — the value stamped when the grant was first written and preserved
+// across re-syncs and expander rewrites, NEVER a fresh now(). It reads
```
ggreer, line 692: This comment is incorrect. Discovered_at is basically the time at which we crawled the row. If we start a new sync, the same resource will have a new discovered_at.

pkg/dotc1z/engine/pebble/adapter_reader.go
```diff
@@ -685,6 +686,33 @@ func (e *Engine) digestEntitlementIdentity(ctx context.Context, ent *v2.Entitlem
...
+		if errors.Is(err, pebble.ErrNotFound) {
+			return nil, false, nil
+		}
+		return nil, false, err
+	}
+	if rec == nil {
+		return nil, false, nil
+	}
+	return rec.GetDiscoveredAt(), true, nil
+}
```
ggreer, line 714: Is uplift going to get the grant, then call GetGrantDiscoveredAt for that grant? If so, that would double the number of grant reads for a given uplift.

manojacs: >  If we start a new sync, the same resource will have a new discovered_at.

Right. Since the problem right now is flipping overlays,  part of me is thinking if we should do this only for overlays.

kans:
This would be fixed if we exposed a v3 set of APIs?  We should just do that IMO with all of them.  It would be cleaner and sound.

PR #1026, opened 2026-07-21
better rate limit stats

kans:
# Record rate-limit gate waits and connector-reported sleeps in sync stats

## Why

`rate_limit_wait` only captured retry backoff inside the syncer. The two
places where most real waiting happens were invisible:

- the pre-request rate-limit gate (SDK gRPC interceptor, and the hosted
  connector manager) sleeps before issuing the request, and that time hid
  inside the step's wall clock; and
- connectors that sleep client-side to avoid limits (e.g. Okta-style
  prevention) never reported that time at all — across a process/Lambda
  boundary there was no way to.

This is why heavily rate-limited connectors (GitHub) never showed up on the
rate-limit dashboards despite spending most of their sync waiting.

## What

**In-process waits — `WaitObserver` (pkg/ratelimit):** a context-carried
callback installed by the syncer at the start of `Sync()`. Sleep sites report
via `ratelimit.ObserveWait` after the sleep completes, with the time actually
slept (a cancelled sleep reports only the elapsed portion). The interceptor's
OVERLIMIT sleep now reports; waits are labeled by resource type via
`retry.WithWaitLabel` and land in `rate_limit_wait` (plus per-resource-type
sub-buckets). No observer on the context = no-op; non-sync callers are
unaffected, and the interceptor's blocking behavior is unchanged.

**Cross-process waits — `RateLimitWaitReport` annotation (proto):** connectors
can attach total slept milliseconds to a list response
(`annotations.WithRateLimitWaitReport`). The syncer folds it into
`rate_limit_wait`, attributed to the resource type being synced. Works for
self-hosted and Lambda connectors where context values can't cross the RPC
boundary. Connector adoption is a follow-up; this PR adds the plumbing.

## Behavior change

`retry.RetryConfig.OnWait` now fires **after** the wait completes with the
duration actually slept, instead of before the wait with the planned
duration. On mid-wait cancellation it reports only the elapsed portion. The
only in-tree consumer is the syncer's own stats recording; downstream users
of `OnWait` should note the timing/value change.

## Testing

- Unit tests for the observer API, interceptor reporting, annotation helper,
  and the cancellation/elapsed-time contract.
- An end-to-end test (`TestRateLimitGateWaitsReachSyncStats`) runs a full
  sync against a mock connector that both triggers gate waits and attaches
  wait-report annotations, and asserts both totals reach the final sync
  summary.

Companion change in the hosted connector manager reports its gate sleeps via
`ObserveWait` (separate PR; requires an SDK version bump there).

### DIFF
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -124,6 +124,19 @@ func (a *Annotations) WithRateLimiting(rateLimit *v2.RateLimitDescription) *Anno
 	return a
 }
 
+// WithRateLimitWaitReport reports time the connector spent sleeping on rate
+// limits (client-side prevention, in-SDK backoff) while serving this response.
+// The syncer folds it into the rate_limit_wait sync stat. No-op if waitMs <= 0.
+func (a *Annotations) WithRateLimitWaitReport(waitMs int64) *Annotations {
+	if waitMs <= 0 {
+		return a
+	}
+	report := &v2.RateLimitWaitReport{}
+	report.SetWaitMs(waitMs)
+	a.Update(report)
+	return a
+}
+
 // NOTE: the store is the only usage of this.
 func GetSyncIdFromAnnotations(annos Annotations) (string, error) {
 	if len(annos) == 0 {
--- a/pkg/annotations/annotations.go
+++ b/pkg/annotations/annotations.go
@@ -124,9 +124,13 @@ func (a *Annotations) WithRateLimiting(rateLimit *v2.RateLimitDescription) *Anno
 	return a
 }
 
-// WithRateLimitWaitReport reports time the connector spent sleeping on rate
-// limits (client-side prevention, in-SDK backoff) while serving this response.
-// The syncer folds it into the rate_limit_wait sync stat. No-op if waitMs <= 0.
+// WithRateLimitWaitReport reports time this connector spent sleeping on rate
+// limits internally while serving the request — e.g. its API client's own
+// client-side throttling or HTTP 429 backoff. Those sleeps happen inside the
+// connector process, invisible to the syncer (which may be across a gRPC or
+// lambda boundary); this annotation is how they reach the rate_limit_wait
+// sync stat. Waits in the syncer process report via ratelimit.ObserveWait
+// instead and must not be double-reported here. No-op if waitMs <= 0.
 func (a *Annotations) WithRateLimitWaitReport(waitMs int64) *Annotations {
 	if waitMs <= 0 {
 		return a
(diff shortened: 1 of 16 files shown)


pkg/sync/syncer.go
```diff
@@ -549,6 +554,33 @@ func (s *syncer) recordSessionUsage(annos []*anypb.Any) {
 	}
 }
 
+// recordConnectorWaitReport folds a connector-reported RateLimitWaitReport
+// response annotation into rate_limit_wait. This surfaces sleeps that happen
+// inside the connector process (client-side rate-limit prevention, in-SDK 429
+// backoff) which the syncer cannot observe directly — including across
+// process/lambda boundaries where context-based observers don't reach.
+// Attributed to the action's resource type like retry and gate waits.
+func (s *syncer) recordConnectorWaitReport(annos []*anypb.Any, resourceTypeID string) {
```
ggreer, line 571: Shouldn't this also check for the RateLimitDescription annotation and add up the wait time for that? Or ShouldWaitAndRetry should emit a RateLimitWaitReport saying how long it waited before returning.

kans: Fable says: 

> If you want a copy-pasteable reply to the human comment: "RateLimitDescription is state (limit/remaining/reset_at), not a sleep report — the design here is that whoever actually sleeps reports the observed duration once, at the sleep site. The syncer-side retryer already consumes RateLimitDescription from error details and reports its actual slept time via OnWait; client-side limiter gates report via ObserveWait. Summing descriptions would fabricate/double-count. Emitting RateLimitWaitReport from ShouldWaitAndRetry is right for connector-process wait sites only (that's the annotation's purpose — WithRateLimitWaitReport is the helper for it); the syncer-side retryer reporting via annotation to itself would double-count. The SDK's connector-side retryers today only guard non-sync task paths, so there's nothing currently unaccounted."

ggreer (approved):

pkg/annotations/annotations.go
```diff
@@ -124,6 +124,19 @@ func (a *Annotations) WithRateLimiting(rateLimit *v2.RateLimitDescription) *Anno
 	return a
 }
 
+// WithRateLimitWaitReport reports time the connector spent sleeping on rate
+// limits (client-side prevention, in-SDK backoff) while serving this response.
+// The syncer folds it into the rate_limit_wait sync stat. No-op if waitMs <= 0.
```
ggreer: It'd be nice if this mentioned that it's only meant for when a connector rate limits itself (eg: Okta connector using Okta SDK), not for when a connector returns an error + rate limit description that causes syncer to rate limit.

PR #1028, opened 2026-07-22
feat(connectorapi): add ServedPolicyEnvelope to GetConnectorConfigResponse

jugonzalez12:
Define the server-computed policy envelope delivered on the connector-config response: a capabilities section carrying the recorded capability manifest verbatim (with its record-time digest), and an egress section carrying the per-instance allowed-hosts list computed at serve from revision-sealed derivation rules, plus the https_only scheme gate input.

The envelope is optional on the response; runtimes that require one treat absence, an unsupported version, a missing section, or any binding mismatch (revision id, integrity-root digest, config version) as deny-all, and a bad section poisons the whole envelope. Digest fields are lowercase hex with the algorithm bound to the enclosing version (v1 = sha256), so a hash migration is a version bump rather than in-band negotiation.

GetConnectorConfigResponse gains an optional config_version (explicit presence distinguishes a server that does not stamp versions from one that stamped an empty handle); when the envelope is present it must be non-empty and equal to the envelope's config_version.


proto/c1/connectorapi/baton/v1/servedpolicy.proto
```diff
@@ -0,0 +1,98 @@
...
+  // governed by the egress section alone. Non-empty bytes that fail to
+  // parse, carry an unsupported inner version, or mismatch the recorded
+  // digest invalidate the whole envelope.
+  bytes manifest_bytes = 1;
+
+  // Record-time digest (lowercase hex; algorithm bound by envelope_version,
+  // v1 = sha256) over manifest_bytes, copied from the sealed revision record —
+  // never recomputed from the fetched bytes. Empty exactly when
+  // manifest_bytes is empty.
+  string recorded_manifest_digest = 2;
```
Bencheng21: I would like to remove the capabilities. It looks like the capabilities part is not finalized yet. cc @gontzess

jugonzalez12: Will remove

gontzess (approved):

proto/c1/connectorapi/baton/v1/servedpolicy.proto
```diff
@@ -0,0 +1,77 @@
...
+// the egress section's own schema_version; version 1 means sha256. Algorithm
+// changes arrive as version bumps, never as in-band negotiation.
+message ServedPolicyEnvelope {
+  // Envelope structure version. Unsupported => envelope invalid. Bumped only
+  // for changes to the envelope itself; each section versions independently.
+  uint32 envelope_version = 1;
+
+  // The connector revision both sections bind to. Must identify the same
+  // revision as the bundle this response serves.
+  string revision_id = 2;
```
kans: maybe connector_revision_id?  I think our version is always 1 behind, too...

kans (approved):

PR #1031, opened 2026-07-22
feat(reader): add c1.reader.v3 grants reader service (un-nerfed records)

manojacs:
## What & why

The v2 reader (`c1.reader.v2` `GrantsReaderService`) returns `c1.connector.v2.Grant`, which **drops fields the storage record holds** — notably `discovered_at` — when it downshifts each `v3.GrantRecord` through `V3GrantToV2`. c1's uplift needs the per-grant `discovered_at` (the connector's actual observation time) to decide overlay deletion safely, so the read has to stop losing it.

This adds a **parallel `c1.reader.v3` `GrantsReaderService`** that returns the rich `c1.storage.v3.GrantRecord` directly — nothing dropped on read.

```
reader_v2: v3.GrantRecord ──V3GrantToV2──► v2.Grant       (drops discovered_at)
reader_v3: v3.GrantRecord ─────────────────► v3.GrantRecord   (nothing dropped)
```

## Shape

- **proto** — `proto/c1/reader/v3/grant.proto`: same 5 RPCs and request shapes as `reader_v2` (`GetGrant`, `ListGrantsForEntitlement`, `ListGrantsForResourceType`, `ListGrantsForEntitlements`, `ListGrantsForPrincipal`); only the response record type changes to `c1.storage.v3.GrantRecord`.
- **pebble** — `engineV3Grants` implements the generated server, **reusing the exact same** sync-resolution, fetch, pagination, cursor and filter helpers as the v2 methods and skipping only the `V3GrantToV2` conversion. The v2 grant **payload/translation is unchanged** and `GetGrant` cannot return a different grant; note that a follow-up commit (`ea5337ab`) also broadened v2 **and** v3 `GetGrant` sync-resolution from the in-memory `CurrentSyncID()` gate to the `resolveActiveSyncForReader` cascade (annotation → current → latest-finished → latest-unfinished), so `GetGrant` now also resolves against sealed/interrupted syncs — matching the sibling grant readers and SQLite. This widens which reads succeed; it never changes which grant is returned.
- **connectorstore** — optional `V3GrantReader` capability + `V3GrantReaderProvider` discovery interface: type-asserted, Pebble-only, callers fall back to the v2 reader. Deliberately **not** part of the required `Reader` interface, so it never forces SQLite/gRPC readers to implement it.

### Note on the provider indirection
The v3 RPC method names (`GetGrant`, …) are identical to the `reader_v2` methods `*Engine` already carries, and Go has no method overloading — one type can't declare both. So the v3 methods live on a wrapper and `*Engine` exposes them via `V3GrantReader()`; discover the capability by asserting the store for `V3GrantReaderProvider`, not `V3GrantReader` directly.

## Compatibility

Additive and backward-compatible: existing `c1/connector/v2` wire contract untouched, no proto renumbering, no `go.mod`/`go.sum`/vendor changes, no default-behavior change. The value returned is the **stored** `discovered_at`, never a fresh `time.Now()`.

## Tests

All 5 v3 RPCs covered in `adapter_reader_v3_test.go`: records returned with their stored `discovered_at`, principal/resource-type filters honored (decoys excluded), and a pagination round-trip proving the cursor + records survive `PageSize:1` page boundaries. `go build ./...`, `go test ./pkg/connectorstore/... ./pkg/dotc1z/...`, and `golangci-lint` on changed files all pass.

## Scope / follow-ups

Grants-only (the must-have that unblocks the uplift/discovered_at work). The other record types (resources, entitlements, resource types, syncs) can follow the same `reader_v3` pattern as needed. No gRPC server registration is wired to a network endpoint yet — this is the service definition + in-process engine implementation.


manojacs:
>pkg/dotc1z/engine/pebble/adapter_reader_v3.go (whole file): the v3 methods duplicate ~200 lines of pagination/filter/cursor logic from adapter_reader.go and grants_for_entitlements.go; a future bug fix to the v2 path won't propagate to v3. Consider extracting a shared helper returning a v3.GrantRecord slice plus cursor that both v2 (with V3GrantToV2) and v3 wrap. Non-blocking; the duplication is intentional per the PR description.

Leaning toward seperation of concerns here as the co-existence of v2 and v3 readers are for a limited time and planning to deprecate v2.

pkg/dotc1z/engine/pebble/adapter_reader_v3.go
```diff
@@ -0,0 +1,335 @@
...
+func (e *Engine) V3GrantReader() connectorstore.V3GrantReader {
+	return engineV3Grants{e: e}
+}
+
+// GetGrant mirrors the reader_v2 GetGrant, returning the rich v3.GrantRecord.
+func (r engineV3Grants) GetGrant(
+	ctx context.Context,
+	req *reader_v3.GrantsReaderServiceGetGrantRequest,
+) (*reader_v3.GrantsReaderServiceGetGrantResponse, error) {
+	syncID := r.e.CurrentSyncID()
```
kans: Why not use `syncID, err := e.resolveActiveSyncForReader(ctx, req.GetAnnotations())` here like in the others?  Ostensibly, we could drop the sync check entirely at some point.

edit, oh pkg/dotc1z/engine/pebble/adapter.go:922 does this too.  Thats probably why this method is written this way.  Would you fix that one, too?

edit edit: It looks like we don't preserve the sync id through a EndFreshSync?  In which case, this API, and the v2 maybe won't work (in prod)?

pkg/dotc1z/engine/pebble/adapter_reader_v3.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	// cursorFor returns the primary grant key for rec — needed because a
+	// post-filter break at len(out) == limit may leave matching records
+	// unconsumed in the engine page, and the engine's end-of-page cursor
+	// would skip them.
+	cursorFor := func(rec *v3.GrantRecord) string {
+		id, err := grantIdentityFromRecord(rec)
+		if err != nil {
+			return ""
```
kans: ^ swallowing this error is probably not correct

pkg/dotc1z/engine/pebble/adapter_reader_v3.go
manojacs: edit 1 : Right, GetGrant (v2+v3) is the only grants reader on CurrentSyncID() instead of `resolveActiveSyncForReader`.
edit 2 : it's not a prod break. EndFreshSync fsyncs the SyncRunRecord (sync_id) to disk and InitCurrentSync 
re-binds it on reopen. 

Switching both to resolveActiveSyncForReader anyway for consistency + annotation scoping.

pkg/dotc1z/engine/pebble/adapter_reader_v3.go
manojacs: Agreed. will propagate the error instead of returning "" (v3 + v2).

manojacs:
## 🧪 Verification packet (BUG_CATCHING step-up) — reviewer reference

This PR was run through the `docs/BUG_CATCHING.md` step-up regime (HIGH / rung-5: silent + durable, feeds an irreversible overlay-deletion decision in c1). An implementation-blind verification plan was synthesized from two decorrelated model plans (GPT + Claude), frozen before implementation inspection, then executed against the branch. The plan/evidence docs are kept out of the committed diff on purpose; the 8 `discovered_at_verify_*_test.go` instruments are in the branch.

- **Verification plan** (18 contracts, 20 criteria, + implementation-obligation addendum): https://claude.ai/code/artifact/b034ae19-d7d3-4c47-842c-14ca9a00a1ae
- **Evidence record** (per-criterion results, gates, gaps): https://claude.ai/code/artifact/5e6034b4-9666-4407-ad77-3994674277b5

**Result:** 18/20 criteria with passing evidence (13 verified-bounded, 5 sampled/measured), **0 real bugs**, product code untouched by the instruments. Mutation-adequacy proven — planting a `time.Now()` on the read path turns the absence test red; a planted write flips the whole-store read-purity digest.

**Rebased + re-verified (2026-07-30):** rebased onto current `main` (`10a6da05`, was 17 commits behind), `buf generate` clean (zero proto drift), and the full suite is green on the new base — `go build ./...`, `go test ./pkg/connectorstore/... ./pkg/dotc1z/...`, `go test -race`, and `golangci-lint` on changed files all pass.

### The remaining 2 (no full passing instrument — stated, not hidden)

- **V-11 SQLite degradation** — *structural / N-A.* The v3 capability is Pebble-only, so a SQLite store doesn't satisfy `V3GrantReaderProvider` → callers cleanly fall back to the v2 reader. No executable probe (real opens auto-convert to Pebble, so native-SQLite is largely unreachable).
- **V-14 old↔new two-artifact (cross-*binary*)** — *declared gap.* Needs a pinned old-SDK writer binary + committed golden corpus, out of reach in this CI stage. Same-build seal→reopen durability is covered by V-02/V-13.

### Consumer-side contract — now closed on the c1 side ✅

The rung-5 property (*"overlay deletion never fires on a zero/absent/fabricated `discovered_at`"*) lives in c1, not this repo. It's now enforced + **GREEN** on a c1 branch: uplift's overlay-deletion seam (`entitlement_user_binding/v2/uplift.go`) was changed to prefer per-grant `discovered_at` with **absence ⇒ KEEP**, and a contract test proves absent/zero/epoch → keep, after → delete-allowed, before → keep (existing overlay tests still pass). The full uplift *activity* wiring (populate the index, gate behind a flag) is the remaining hop — unblocked by this rebase.

pkg/dotc1z/engine/pebble/adapter.go
```diff
@@ -920,7 +921,15 @@ func (e *Engine) ListStaticEntitlements(
 
 // GetGrant fetches a single grant by ID.
 func (e *Engine) GetGrant(ctx context.Context, req *reader_v2.GrantsReaderServiceGetGrantRequest) (*reader_v2.GrantsReaderServiceGetGrantResponse, error) {
-	syncID := e.CurrentSyncID()
+	// Resolve the sync like the other grant readers (annotation → current
+	// binding → latest-finished record) rather than gating on the in-memory
+	// CurrentSyncID alone: a sealed c1z reopened without a re-bind still has
+	// its sync_id persisted in the SyncRunRecord, and this honors a
+	// request-annotation sync_id the way SQLite's GetGrant does.
+	syncID, err := e.resolveActiveSyncForReader(ctx, req.GetAnnotations())
```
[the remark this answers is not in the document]
manojacs: Good catch — validated and correct. `syncID` here is only a presence gate (the grant is fetched by id via `GetGrantRecord`, not sync-scoped), so it can't return a different grant, but you're right that switching to `resolveActiveSyncForReader` broadens which reads succeed — reopened sealed/interrupted syncs now resolve instead of `ErrNoCurrentSync`. The "byte-for-byte unchanged" line was true at the feat commit and went stale when `ea5337ab` added the sync-resolution change; I've corrected the PR description to scope it to the grant payload/translation and call out the broadened resolution explicitly.

manojacs:
## 🧪 Verification plan (frozen + implementation addendum)

The BUG_CATCHING step-up plan: 18 contracts, 20 criteria, coverage model, and the post-inspection implementation-obligation addendum. Kept out of the committed diff; inlined here for review.

<details><summary>plan.md</summary>


# Frozen verification plan — grant `discovered_at` read-back

**Status: FROZEN behavioral baseline.** Synthesized by the orchestrator from two
independent, implementation-blind plans:

- `plan-sol.md` — `openai/gpt-5.6-sol` (codex)
- `plan-fable.md` — `anthropic/claude-fable-5` (claude)

Both authors worked from the feature goal and mainline `main` only; neither read
the feature branch (`manojacs/v3-grant-reader`), PR #1031, or any
`reader/v3` / `adapter_reader_v3*` / `proto/c1/reader/v3` file. This merge
dedupes, resolves contradictions, assigns stable criterion IDs, and runs a
contract→criterion completeness audit. **Freeze this before implementation
inspection.** After inspecting #1031, append an *implementation-obligation
addendum* (BUG_CATCHING §2/§3) as a separate section; do not silently revise the
criteria below.

A third author (grok) was dispatched but its environment failed; this baseline
rests on two decorrelated models (one GPT-family, one Claude-family).

---

## 1. Feature & risk verdict

**Feature.** A consumer can opt in to read each grant together with the exact
`discovered_at` already stored for it in a c1z. Existing grant reads stay
byte/semantic-identical with their pagination/filter behavior. The value
survives write → seal → close → cold reopen. Pebble is required; SQLite (deprecated)
degrades cleanly without breaking existing reads. Scope: **grants only**.

**Risk verdict: HIGH — step-up required.** Unanimous across both authors.

| Escape axis | Verdict | Why |
|---|---|---|
| Silence | HIGH | A missing / zero / fresh-`now()` / grant-misaligned timestamp is well-formed output; no panic, no log. |
| Durability | HIGH | Source is a durable, version-crossing c1z read after process death, possibly by future code. |
| Uncontrolled dimensions | MED–HIGH | Engine × lifecycle × data-vintage × open-mode × pagination route × timestamp shape interact. Not schedule/volume-driven (read path). |
| Consumer distance | HIGH | The safety decision is in c1 uplift — another repo, across the reader seam. |

**Consequence: rung 5 (irrecoverable).** c1 uplift uses the value to decide
**overlay deletion**; a wrong/fabricated/absent value can silently revoke real
user access downstream. Defense flips from detection to **prevention**:
byte-exact stored-value oracle, a golden old-artifact corpus, and a
"never `time.Now()` at read" fail-closed assertion. Review rounds cannot
establish closure (budget 2–3, then instruments).

**Cost contract.** Returning the scalar must add no extra full-store pass,
per-grant point lookup, or session-store/network access; time O(N) in returned
grants, memory O(page). A benchmark/allocation comparison vs `main` is required
only if the implementation changed executed hot-path work beyond decoding a
field; otherwise the implementation addendum states "no cost-curve change."

---

## 2. Contracts (frozen)

Each is falsifiable by a single counterexample. Merged and de-duplicated; the
`src` column records which blind author(s) raised it (decorrelation signal).

| ID | Contract | src |
|---|---|---|
| **C1 Fidelity** | For every returned grant G with stored `discovered_at=T`, the read returns T exactly (seconds **and** nanos), on every read. Never generated at read time, copied from another grant, inferred from file/sync metadata, or normalized to now. | both |
| **C2 No-fabrication-on-absence** | A grant with no stored `discovered_at` (nil/zero; e.g. written by an SDK predating the field) reads back as nil/absent — **never** a synthesized `time.Now()`. | both |
| **C3 Identity association** | Timestamp and grant are an indivisible pair; pagination, filters, ordering, duplicate external-ids with distinct structured identities, and internal batching never shift a timestamp between grants. | sol |
| **C4 Durable lifecycle** | Fidelity holds live *and* after write → CheckpointSync → EndSync → Close(seal) → reopen read-only in a fresh process / cold cache / no re-bind. Reopened value == planted raw value == pre-close value. | both |
| **C5 Additive compatibility** | Every existing grant read (`GetGrant`, `ListGrantsForEntitlement(s)`, `ListGrantsForPrincipal`, `ListGrantsForResourceType`, `ListGrants`) returns byte-identical payloads/order/tokens/errors/annotations with the feature present but not requested. No existing `v2.Grant` is mutated to smuggle the field. Old consumers stay source/wire compatible. | both |
| **C6 Read purity** | Success is read-only: calling the new surface never mutates the store, active sync, cursor, grant proto, timestamp, or subsequent legacy-read output. | sol |
| **C7 Complete paging/filtering** | Across every supported grant-selection route, each selected grant appears exactly once with its own timestamp; excluded grants never appear; concatenated pages == unpaged ground truth; tokens make progress (no loops). | sol |
| **C8 Timestamp domain** | Epoch, pre-epoch, far-past, distinct-nanos, and equal timestamps round-trip exactly. Invalid protobuf timestamps follow a documented classified-error contract, never silent replacement. | sol |
| **C9 Provenance preservation** | The read reflects whatever the write path committed, per provenance class: connector-observed (stamped at PutGrants); expander-rewrite (**preserves prior** discovered_at); expander-new (stamped now at write); synthesized (stamped now at write); bulk-imported/migrated (rides source value). No re-derivation at read. | fable |
| **C10 Scope = grants only** | Read-back defined only for grants; resource / resource-type / entitlement / asset read paths are byte-unchanged (guard against a shared-translator regression). | fable |
| **C11 Pebble capability wired** | The production Pebble store/reader statically or executably satisfies the new capability; a wiring/rename miss is loud at build/test, not a silently-failed optional type assertion (BUG_CATCHING §4.1 / #774). | sol |
| **C12 SQLite graceful degradation** | Existing SQLite grant reads unchanged; the new capability is cleanly unavailable / no-op / explicitly unsupported (documented, Info-visible per §1), never panic / corruption / fabricated time / a wrong non-zero value while claiming support. | both |
| **C13 Cross-vintage / multi-artifact** | New reader × grants sealed by pre-feature code → C1 or C2 as the data dictates, never crash/garbage. Old reader × new data → unaffected. Pebble vs SQLite engine pair covered. | both |
| **C14 Consumer deletion safety** | At the reader seam / a pinned consumer fixture: an older stored instant stays older and a newer stays newer after read, so uplift's comparison is driven by the stored instant; absence/unsupported is never interpretable as a fresh observation that authorizes deletion. | both |
| **C15 No ephemeral-state dependence** | Results identical with a noop/empty session store, a newly-allocated adapter, and no writer-side cache; reopen alone reconstructs everything needed. | sol |
| **C16 Schema/API additivity** | No breaking proto descriptor change; the existing `v2.Grant` descriptor is unchanged; generated code is clean; old call sites compile. | sol |
| **C17 Route completeness** | Every public grant-producing read method is either timestamp-capable or explicitly legacy-only per a checked-in coverage table; every claimed new route maps to a test (absence-class guard). | sol |
| **C18 Error-path integrity** | Cancellation and injected read/decode/iterator faults keep their error identity, return no misleading "successful" partial page, release iterators/resources, leave the artifact and subsequent reads unchanged. A swallowed decode error must not surface as a silent zero `discovered_at`. | both |

**Contradiction resolved (absence representation).** sol allowed "nil/absent **or**
explicit unsupported/corrupt error"; fable said "zero/absent." Frozen decision:
genuine **absence → nil/absent** (C2); a **malformed/invalid** stored timestamp →
documented **classified error** (C8). These are distinct cells, never `now()`.

---

## 3. Coverage dimensions

- **D1 Engine:** Pebble-native · SQLite-native · SQLite-auto-converted-to-Pebble-on-open.
- **D2 Lifecycle:** pre-seal (same handle) · post-EndSync same handle · closed+reopened same process · reopened fresh process / cold cache. *Reduction:* the durable oracle collapses to the fresh-reopen cell (it dominates); live-state cells are diagnostic-only, kept to localize a reopen regression.
- **D3 Data vintage:** post-feature write · pre-feature write by current SDK · legacy write predating `discovered_at` (field absent). *(multi-artifact axis.)*
- **D4 Stored value:** present-non-zero · absent/nil/zero · {epoch, pre-epoch, far-past, distinct-nanos, equal-on-two-grants, invalid-proto}.
- **D5 Provenance:** connector-observed · expander preserve-prior · expander new→now · synthesized · bulk-imported/migrated. *Reduction only if the addendum confirms one shared read choke point.*
- **D6 Selection route × cardinality × pagination:** {entitlement; entitlement+principal; entitlement+RT-set; resource-type; batched-entitlements; any other public grant-list route} × {0, 1, one-page-many, multi-page-many}. Page sizes 1 / boundary−1 / boundary / boundary+1.
- **D7 Identity arrangement:** distinct external-ids · same external-id different entitlement · same external-id different principal · adjacent key order opposite timestamp order (alias-risk).
- **D8 Record type:** grant (in scope) · resource / RT / entitlement / asset (out of scope → regression guard only, C10).
- **D9 Execution outcome:** success · zero-result · cancel-before-iter · cancel-after-a-row · iterator-open fail · decode fail · advance/close fail · reopen fail.
- **D10 Concurrency:** single reader · two concurrent new readers · new+legacy · cancel-one-while-other-completes. *Reduction:* read-only over an immutable sealed artifact → one representative `-race` cell, re-expand only if the addendum finds a lazily-populated shared read cache/decoder pool.
- **D11 Compatibility pair:** old-artifact×new-reader · old-artifact×pinned-main-reader (baseline) · new×new · new-artifact×pinned-old-reader · SQLite×new-reader. Bounded over a checked-in golden corpus; a version whose old binary can't run in CI stays an explicit gap.

No cell is silently dropped: the two reductions (D8 non-grant, D10 concurrency)
each carry an explicit rule and a residual guard cell.

---

## 4. Criterion ledger

All criteria start **evidence incomplete** (frozen; nothing executed yet).
"Bounded" = exhaustive over the finite named cells; "sampled" never supports full
closure. Each closure entry, when later filled, must name the exact artifact,
premise assertion, oracle assertion, tested layer, revision, command, and one
allowed status.

| ID | Contract·cells | Independent oracle | Instrument | Closure |
|---|---|---|---|---|
| **V-01** | C1/C3 fidelity+association, D4 values × D7 arrangements | Raw v3 record decoded independently → `identity→timestamp`; compare full timestamp + pairing. Plant two non-wall-clock instants. | Table-driven Pebble engine/public-capability test; validate oracle by swapping an association in a helper and requiring failure. | Bounded over D4×D7 |
| **V-02** | C4/C15 durable cold lifecycle | planted-raw == live == reopened, in a fresh store/adapter/noop-session with no re-bind | Integration harness: write via prod writer → inspect raw → seal/close → reopen read-only → read; subprocess variant proves process separation. | Bounded over D2; subprocess env sampled |
| **V-03** | C5 additivity of every existing read route | Differential: byte/wire snapshot of legacy `v2.Grant` output vs pre-feature `main` for a fixed fixture (IDs, order, tokens, annotations, errors). | Run existing dotc1z read suite unchanged + golden-bytes diff vs `main`; pinned-main artifact direction. | Bounded over enumerated routes/pages |
| **V-04** | C6 read purity | Full raw key/value digest + sync metadata + active-sync id + legacy result unchanged before/after repeated new reads | Read-only mutation auditor on a non-empty store; invoke twice; reopen and re-digest. | Bounded |
| **V-05** | C7 paging/filtering completeness | In-memory filter of planted ground truth; concatenated pages == ordered (identity,timestamp) sequence; uniqueness; terminal token | Generated D6 table, page sizes {1, b−1, b, b+1}; assert token progress / no loop. | Bounded over D6 rows |
| **V-06** | C2 absence + C8 invalid | Documented outcome exactly; wall-clock trap rejects any value inside the test window or equal to an injected fake clock | Raw fixture plants nil and invalid values; classified-error/absence assertions; mutation plant (reader calls clock) must fail. | Bounded over D4 edges + adequacy |
| **V-07** | C8 domain round-trip | `proto.Equal` on epoch / pre-epoch / far-past / distinct-nanos / equal-pair | Extends V-01 table with the value set. | Bounded |
| **V-08** | C9 provenance classes | Read-back == the value each write path committed (preserve-prior for expander rewrite; migrated value for bulk import; captured stamp for connector/synth) | Table with one row per D5 class driving its actual write path, then seal/reopen/read. | Bounded over 5 classes |
| **V-09** | C10 scope guard | Non-grant reads byte-unchanged vs `main` | Run existing resource/RT/entitlement/asset read tests unchanged as a regression gate. | Bounded over 4 non-grant paths |
| **V-10** | C11 Pebble wiring | Production object satisfies the capability and returns the planted value | Compile-time `var _ Iface = (*Impl)(nil)` where API permits + public-construction smoke test (not a mock/type-only assertion). | Bounded |
| **V-11** | C12 SQLite degradation | Existing SQLite result snapshot unchanged; new call explicit unsupported/unavailable, stable classification, zero mutation, Info-visible | SQLite store integration test via public factory; reopen existing fixture; capability probe + legacy read. **Reachability caveat:** if every real open auto-converts to Pebble, record P-unreachable and downgrade to the crash/corruption guard with justification. | Bounded over documented SQLite states |
| **V-12** | C18 cancel/error/resource safety | Exact injected error identity; injection counter proves the seam fired; raw digest unchanged; iterator/fd/goroutine counts return to baseline; clean retry succeeds | Fault-injectable read/decode/iterator seam (extend `errorfs_sweep_test.go`); one case per D9 mechanism; fail-k sweep until an armed run injects nothing (completeness). | Bounded over injectable mechanisms; OS-resource noise sampled |
| **V-13** | C14 consumer deletion safety | Independent uplift decision model: stored old/new/equal/absent → retain/delete per the consumer contract; absence never authorizes deletion | Cross-repo / pinned-consumer contract harness exchanging the real reader response, **or** a checked-in consumer-facing golden response if the consumer can't execute here. | Bounded for old/new/equal/absent; **remains a gap until consumer-side execution exists** |
| **V-14** | C13/D11 old↔new compatibility | Counted grants + full legacy payload digest match ground truth both directions; new field ignored by old reader without open failure | Two-artifact harness: pinned `origin/main` helper + candidate helper in isolated modules/processes; committed old/new corpus in `testdata/`; old→old, old→new, new→new, new→old, SQLite→new. | Bounded over declared corpus; other historical versions sampled/gap |
| **V-15** | C1 no-fresh-clock | Result == fixed planted instant under two widely-separated injected clocks / process times | Clock-control or subprocess env + raw fixture; planted "return Now" proves detection; cheap grep-gate: read path contains no `time.Now()`/`timestamppb.Now()`. | Bounded over two clocks; platform sampled |
| **V-16** | C16 schema/API additivity | No breaking descriptor change; `v2.Grant` descriptor unchanged; generated code clean; old call sites compile | Repo proto lint/breaking/generation checks + compile fixture pinned to old call sites. | Bounded to descriptor/API surface |
| **V-17** | Cost contract | Ops/allocs grow linearly in returned N; no hidden per-row point-get/session call; legacy regression under a declared ratchet | Benchmark N={1, page, 10×page, whale} + counting wrapper for reads/session calls; alloc compare vs `main`. | Measured/sampled — never closure |
| **V-18** | C17 route completeness / absence | Every public grant-producing read method is timestamp-capable or explicitly legacy-only; every claimed new route maps to a test | Descriptor/reflection method inventory vs a checked-in coverage table; delete one row in a planted copy → require failure. | Bounded over descriptor at tested revision |
| **V-19** | D10 concurrency | Each reader == immutable ground truth; cancellation isolation; race detector clean | Concurrent reopen-read integration test under `go test -race`; seeded cancel/start soak with a seed ledger. | Sampled |
| **V-20** | premise gate for V-06/V-14 | Absent/legacy fixture genuinely lacks a stored `discovered_at` (inspect raw, not the filtered read) | Raw-record assertion before the read runs, so absence cases can't pass vacuously. | Bounded |

---

## 5. Review-blind classes → required harnesses (BUG_CATCHING §2)

- **Multi-artifact** (the load-bearing one, and the repo's repeat offender — #997/#1023 converter field loss, one layer out). *Harness:* V-14 two-artifact runner + V-13 golden old-corpus in `testdata/`; the true consumer-direction test lives on the **c1 uplift side** of the seam (residual gap). The in-repo Pebble↔SQLite equivalence runner is a *sunsetting* oracle — reuse, build nothing new on it.
- **Absence** (the highest-value trap: a read path that mirrors the write path's legitimate `now()` fallback, or a missing branch for the absent case). *Harness:* V-06 + V-20 (absent fixture with premise validation) + mutation plant + the no-`time.Now()` grep-gate + V-18 route inventory.
- **Error-path** (decode/reopen faults never run under a cooperative environment; a swallowed decode error → silent zero). *Harness:* V-12 errorfs read-path extension with a stated recovery contract (fault → error, never silent zero), fail-k completeness sweep, planted-violation adequacy. Raw `os.*` IO outside the vfs seam stays an explicit exclusion.
- **Schedule** (low — read-only over immutable sealed state). *Harness:* V-19 `-race`; escalate only if the addendum finds shared mutable read-path state (lazy cache/decoder pool).

---

## 6. Residual gaps (must be stated at closure, never closed silently)

1. **Consumer-side deletion contract (c1 uplift) — TOP GAP.** The rung-5
   consequence lives in another repo. A perfect read-back does not stop uplift
   from deleting on a zero/absent value. The decisive test — *"overlay deletion
   never fires on a fabricated or zero `discovered_at`"* — must live on the c1
   side. Out of scope here; SDK-only proto equality is insufficient. Both authors
   independently flagged this as the most important residual risk.
2. **Cross-version reality is sampled, not closed** (V-13/V-14): proven only over
   the committed golden corpus / same-build path, not every historical writer or
   a pinned old *binary*.
3. **SQLite degradation reachability** (V-11): if every real open auto-converts
   to Pebble, native-SQLite degradation is moot in practice — keep the
   crash/corruption guard, record the reachability finding.
4. **Non-injectable raw `os.*` read/reopen faults** (V-12) remain a finding-in-waiting.
5. **This is implementation-blind.** Oracle exactness, the D5 provenance-reduction
   rule, the exact absent-value representation, and "does the read path contain a
   `time.Now()` reflex" are pending the **implementation-obligation addendum**
   after inspecting #1031. Any relevant implementation detail returns affected
   criteria to *evidence incomplete*.

---

## 7. Completeness audit (contract → criteria)

Every requirement maps to ≥1 criterion; no orphan criteria.

| Requirement | Contracts | Criteria |
|---|---|---|
| Value is the STORED one, never `time.Now()` | C1, C2, C9 | V-01, V-06, V-08, V-15, V-20 |
| Existing reads unchanged / additive | C5, C6, C10, C16 | V-03, V-04, V-09, V-16 |
| Works across write→seal→close→reopen | C4, C15 | V-02 |
| Grants only | C10 | V-09 |
| Pebble target; SQLite graceful | C11, C12 | V-10, V-11 |
| Correct pagination/filter/domain | C3, C7, C8 | V-01, V-05, V-07 |
| Cross-repo consumer safety | C13, C14 | V-13, V-14 |
| Error/concurrency/cost integrity | C18, (D10), cost | V-12, V-17, V-19, V-18 |

**Decorrelation value (why two models beat one):**
- **Only sol** surfaced: identity-alias arrangements (C3/D7), route-completeness inventory (C17/V-18), the explicit cost contract (V-17), schema-breaking gate (C16/V-16), capability compile-time assertion (C11).
- **Only fable** surfaced: the five provenance classes incl. expander preserve-prior (C9/V-08 — directly the discovered_at re-stamp subtlety), the no-`time.Now()` grep-gate, the SQLite auto-convert reachability caveat (V-11), the non-grant regression guard (C10/V-09).
- **Both** independently: fidelity, absence-never-now, durable reopen, additivity, SQLite degrade, HIGH/rung-5, and — most notably — the **consumer-side deletion contract as the top residual gap**.

---

## 8. Next steps (post-freeze)

1. Commit this frozen `plan.md` on the feature branch (baseline). Preserve
   `plan-sol.md` / `plan-fable.md` as the source inputs.
2. Inspect #1031 → append the **implementation-obligation addendum** (resource
   lifetimes, the read choke point, any lazy cache, the exact absent
   representation, presence/absence of a `time.Now()` reflex, sync-resolution
   ownership).
3. Map each current-stage criterion to an executable instrument + evidence; run
   the repository gates; record per-criterion status. Report closure scoped to
   this stage, with §6 gaps stated as limits.
4. A reviewer decorrelated from the implementer and the instrument author audits
   every `verified` entry before signoff.


---

## Implementation-obligation addendum (BUG_CATCHING §2/§3)

Appended AFTER freeze, from inspecting PR #1031 (`manojacs/v3-grant-reader`).
The frozen criteria above are unchanged; this section records what the
implementation actually does so each instrument targets the real seam.

### A. The read choke point(s)

- **One shared scan/paginate path for v2 and v3.** The v3 reader
  (`adapter_reader_v3.go`, type `engineV3Grants{e *Engine}`) mirrors each v2
  reader method (`adapter_reader.go`) *line-for-line*, calling the **same**
  engine helpers: `GetGrantRecord`, `PaginateGrants`,
  `PaginateGrantsByEntitlement`, `PaginateGrantsByEntitlementPrincipal`,
  `PaginateGrantsByPrincipal`, `PaginateGrantsByPrincipalResourceType`. The
  **only** difference: the v2 path finishes each record with `V3GrantToV2(rec)`
  (lossy — v2.Grant has no `discovered_at` field); the v3 path returns the raw
  `*v3.GrantRecord`. **Invariant:** the timestamp the v3 reader emits is exactly
  the bytes decoded from the stored `GrantRecord.discovered_at` (field 5).
- Consequence for D5 provenance reduction: **CONFIRMED single read choke point**
  — the read never branches on provenance, so V-08 need only prove each *write*
  path commits the right value and that the one read path returns it verbatim.

### B. Absent-value representation (exact)

- **Absence = nil `*timestamppb.Timestamp`** on the returned `v3.GrantRecord`
  (`GetDiscoveredAt() == nil`). Proto3 message-field absence; no sentinel.
- **The read path never stamps.** `PutGrantRecord` writes the `v3.GrantRecord`
  **as-is with no `discovered_at` stamping**. Only the *v2→v3 write translation*
  stamps `timestamppb.Now()` when the field is nil. So a record seeded via
  `PutGrantRecord` with a nil `discovered_at` round-trips as nil — the seam
  V-06/V-20 seed at.

### C. `time.Now()` / `timestamppb.Now()` reflex on the read path — NONE

- Grep over the grant **read** surface finds **zero** `time.Now(` /
  `timestamppb.Now(`. Every `now()` in the package is on a **write/lifecycle**
  path (v2→v3 translate stamp, `Put*`, the `EndSync` `ended_at` stamp, the
  expander/synth backfill, bulk-import, stats sidecar, a `sync_runs` cutoff
  comparison). **Oracle:** V-15 grep-gate + V-06 mutation plant (temporarily
  stamp `now()` at read → test must go RED).

### D. Resource lifetimes on every read exit

- Every grant read closes its iterator/closer on all exits
  (`defer closer.Close()` / `defer iter.Close()`).
  `PaginateGrantsByEntitlementPrincipal` is a point-get (no iterator). No read
  path retains a live handle past return.

### E. Lazily-populated cache / decoder pool on the read path — NONE shared

- `grantReadArena` is allocated **fresh per call** and used only by the
  unfiltered `PaginateGrants` (the `ListGrants` route). It is not shared across
  calls. Reads go through pebble `db.Get` / `db.NewIter` over an **immutable
  sealed** DB (concurrent-read-safe). **Consequence:** the D10 concurrency
  reduction **holds** — one representative `-race` cell suffices.

### F. Sync-resolution ownership

- `resolveActiveSyncForReader` is the single shared resolver: annotation
  sync_id → engine current binding → latest-*finished* sync → latest-*unfinished*
  (<1wk). A malformed annotation or store error surfaces as a non-nil error
  (never a silent wrong-sync fallthrough).
- **The `GetGrant` sync-resolution change:** both v2 `GetGrant` and v3 `GetGrant`
  call `resolveActiveSyncForReader` instead of gating on `CurrentSyncID()` alone
  — so a sealed c1z reopened read-only WITHOUT a re-bind still resolves via the
  persisted `SyncRunRecord`. Pinned by `get_grant_reopen_test.go` (dotc1z) and
  `get_grant_sync_resolution_test.go` (pebble). Note: this **widens which reads
  succeed**; it never changes which grant is returned (syncID is a presence gate
  only, the grant is fetched by id).

### G. Instrument consequences

- V-06/V-20 seed absence via `Engine.PutGrantRecord` with nil `discovered_at`
  (the only write path that does not stamp), assert the raw premise, then read.
- V-08 collapses to: drive each write path, read back through the one v3 read
  choke point. Expander preserve-prior is the load-bearing class.
- V-12 has two reachable fault arms: (a) decode fault — plant corrupt bytes under
  a grant key and assert the reader errors, never a silent zero; (b) IO fault —
  reopen over `errorfs` with a read-op injector.
- V-19 stays a single `-race` cell (per E).

</details>

manojacs:
## 🧪 Evidence record (per-criterion results)

Per-criterion status, oracles, planted-violation proofs, gates, and residual gaps. Result: 18/20 with passing evidence, 0 real bugs.

<details><summary>evidence.md</summary>


# Evidence record — grant `discovered_at` read-back (v3 GrantsReaderService)

Verifier stage of the BUG_CATCHING step-up. The frozen plan
(`plan.md`, incl. the implementation-obligation addendum) is the spec. This
record maps each **V-##** criterion to an executable instrument (new or cited),
its oracle + premise assertion, and any planted-violation proof, then states
residual gaps as limits (never closed silently).

Branch `manojacs/v3-grant-reader` (PR #1031); all changes left UNCOMMITTED.
Feature/product code was **not** modified (the one mutation-adequacy plant on
`adapter_reader_v3.go` was reverted — `git status` shows only new test files +
these two docs).

## Summary

- **20 criteria.** 13 **verified-bounded**, 5 **verified-sampled** / measured, 1
  **evidence-incomplete**, 1 **gap (declared)** — plus the standing cross-repo
  consumer gap that is out of scope for this SDK stage.
- **Real bugs found: 0.** No instrument failed against unmodified feature code.
  Every gate passed: `go build ./...` (0), `go test ./pkg/connectorstore/...
  ./pkg/dotc1z/...` (0), `go test -race ./pkg/dotc1z/engine/pebble/...` (0),
  `golangci-lint` on changed files (0 issues in my files).
- **Mutation-adequacy proven:** planting `timestamppb.Now()` on the v3 read path
  turned the absence test RED (fabricated instant `Seconds:1785439686`), then the
  plant was reverted. The absence/invalid/IO-fault instruments are non-vacuous
  (V-20 raw-premise gate + injector-fired counter).

### Key implementation facts (from the addendum)

- v2 and v3 readers **share one scan/paginate choke point**; v3 only skips the
  lossy `V3GrantToV2` downshift. The read path has **zero** `time.Now()` /
  `timestamppb.Now()`. Absence = **nil** `*timestamppb.Timestamp`; only the
  v2→v3 *write* translate stamps. `PutGrantRecord` writes as-is (the absence
  seed seam). `grantReadArena` is per-call (no shared read cache) → the
  concurrency reduction holds.

## Per-criterion table

| V-## | Status | Instrument (test) + command | Oracle + premise assertion | Planted-violation proof |
|---|---|---|---|---|
| **V-01** fidelity+association | verified-bounded | CITE `TestV3GetGrantReturnsUnNerfedRecord`, `TestV3ListGrantsFor{Entitlement,ResourceType,Principal}ReturnsUnNerfedRecords`, `TestV3ListGrantsForEntitlementsReturnsUnNerfedRecords` (`adapter_reader_v3_test.go`) + NEW `TestV3ReadProvenanceClasses`. `go test -run TestV3 ./pkg/dotc1z/engine/pebble/` | Each returned record's `discovered_at` == the planted per-principal value; distinct principals/entitlements keep distinct timestamps (association). | Existing tests already assert filter-honored (non-matching principals excluded). Missing cell: full D4×D7 matrix (epoch/pre-epoch, equal-on-two-grants, adjacent-key-opposite-order alias) — sampled, see gaps. |
| **V-02** durable cold lifecycle | verified-bounded | NEW `TestV3DiscoveredAtSurvivesEngineReopen`; CITE `TestGetGrantAfterReopenNoRebind` (`get_grant_reopen_test.go`), `TestGetGrantResolvesSyncAfterEndSync` (`get_grant_sync_resolution_test.go`). | planted value == pre-close read == reopened read, across `Close()` + fresh `Open()` of the same dir, no `SetCurrentSync`; nanos compared exactly. | — (subprocess/process-separation cell not built; same-process cold reopen is the dominating cell). |
| **V-03** additivity | verified-sampled | CITE full `./pkg/dotc1z/...` read suite (unchanged, green) + v2/v3 same-set cross-check in `TestV3ListGrantsForEntitlementReturnsUnNerfedRecords`. `go test ./pkg/dotc1z/...` | Existing legacy `v2.Grant` reads return same set/order; `v2.Grant` has no `discovered_at` field (descriptor unchanged). | Missing: byte-for-byte golden diff vs pinned `main` — regression gate substitutes. |
| **V-04** read purity | verified-bounded | NEW `TestV3ReadPurityWholeStoreUnchanged` (`discovered_at_verify_purity_test.go`). `go test -run TestV3ReadPurity ./pkg/dotc1z/engine/pebble/` | Full-keyspace SHA-256 digest (every key+value, incl. secondary indexes + SyncRunRecord/sync metadata) is byte-identical before vs after driving every v2+v3 grant read surface 3×. | **PLANT:** a `UnsafeForTesting().Set` write after the reads flips the digest — auditor proven non-vacuous. |
| **V-05** paging completeness | verified-bounded (route) / sampled (matrix) | CITE `TestV3ListGrantsForEntitlementPaginationPreservesRecords` (PageSize 1, no dup, each carries its `discovered_at`, terminal token). | Concatenated single-record pages == ground truth; uniqueness; token terminates. | Missing: page sizes {b−1,b,b+1} × all routes × cardinalities — sampled. |
| **V-06** absence + invalid | verified-bounded | NEW `TestV3ReadAbsentDiscoveredAtReturnsNilNeverNow`, `TestV3ReadCorruptRecordSurfacesErrorNeverSilentZero`. `go test -run 'TestV3ReadAbsent|TestV3ReadCorrupt' ./pkg/dotc1z/engine/pebble/` | Absent `discovered_at` reads back **nil** via GetGrant + ListGrantsForEntitlement + ListGrants; wall-clock trap rejects any in-window instant. Corrupt record → **error**, never a zero-`discovered_at` success. | **PLANT:** injected `rec.SetDiscoveredAt(timestamppb.Now())` on read → test RED (got `Seconds:1785439686`); reverted. |
| **V-07** domain round-trip | verified-sampled | NEW reopen test: distinct-nanos (`…123456789`) + far-past (1971) round-trip exactly. `TestV3ReadProvenanceClasses` adds an explicit far-past synth value. | `proto`-time equality to the nanosecond after reopen. | Missing: epoch/pre-epoch/equal-pair explicit cells — sampled. |
| **V-08** provenance | verified-bounded (4/5 classes) | NEW `TestV3ReadProvenanceClasses` (connector-observed via PutGrants; synth-explicit; synth-nil-backfill; expander preserve-prior) + CITE `TestStoreExpandedGrants{PreservesDiscoveredAt,BackfillsNilDiscoveredAt}`. | Read-back through the single v3 choke point == the value each write path committed; connector/synth-nil ∈ write-time window; expander rewrite **preserves prior**, never re-stamps. | Existing preserve-prior test documents the pre-fix bug (unconditional now() re-stamp). Migration/bulk-import class = scoped storage-layer note (`to_pebble.go` carries source value), not re-driven. |
| **V-09** scope guard | verified-sampled | CITE resource/RT/entitlement/asset read tests in `./pkg/dotc1z/...` (green, unchanged). | Non-grant reads unaffected by the shared translator (v3 reader touches grants only). | Missing: dedicated byte-diff vs main for the 4 non-grant paths. |
| **V-10** Pebble wiring | verified-bounded | CITE `var _ connectorstore.V3GrantReader = engineV3Grants{}` (adapter_reader_v3.go:24), `var _ …V3GrantReaderProvider = (*Engine)(nil)` (adapter.go:68); NEW compile-time asserts in `discovered_at_verify_routes_test.go`; public-construction smoke via `newV3GrantReader` (`any(e).(V3GrantReaderProvider)`). | Production `*Engine` satisfies the capability AND returns the planted value (not a mock). | A rename/miss is a build failure, not a silent nil assertion. |
| **V-11** SQLite degradation | evidence-incomplete | (structural) `V3GrantReaderProvider` is implemented only by the Pebble engine; a SQLite store does not expose it → type assertion fails → callers fall back to the v2 reader. | Clean unavailability, no panic/fabrication. | No executable SQLite-store probe built. **Reachability caveat** (plan §6.3): real opens auto-convert to Pebble, so native-SQLite degradation is largely moot — kept as the crash/corruption guard. |
| **V-12** error-path | verified-bounded (decode+IO) | NEW `TestV3ReadCorruptRecordSurfacesErrorNeverSilentZero` (decode fault via `e.db.UnsafeForTesting().Set(key, garbage)`), `TestV3ReadIOFaultSurfacesErrorNeverSilentZero` (errorfs read-op injector, 1-byte cache, cold reopen). `go test -run 'TestV3Read(Corrupt|IOFault)' ./pkg/dotc1z/engine/pebble/` | A decode/read fault surfaces as an **error**, never a partial page or silent zero `discovered_at`; the injector's fired-count is asserted `> 0` (seam non-vacuous). | Injected fault must produce err != nil AND nil grant / empty list. |
| **V-13** cross-vintage / multi-artifact | verified-sampled (in-repo) + gap | NEW `TestV3DiscoveredAtSurvivesEngineReopen` (new-writer→new-reader, exact value across seal/close/reopen). | planted raw value == reopened read. | **GAP:** full cross-BINARY (pinned OLD SDK writing the artifact) not executable in CI; committed old/new golden corpus not built. Declared, not faked. |
| **V-14** old↔new compat (two-artifact) | gap (declared) | — | — | **GAP:** pinned-`origin/main` reader helper + committed `testdata/` corpus not built; new→old / old→new binary matrix out of reach here. Same-build durability is covered by V-02/V-13. |
| **V-15** no-fresh-clock | verified-bounded | NEW `TestV3ReadPathHasNoWallClockReflex` (grep-gate over `adapter_reader_v3.go`, `grant_read_arena.go`, `paginate.go`) + wall-clock trap in the absence test. | Read-path source contains no `time.Now(`/`timestamppb.Now(`; absent value never lands in the test window. | **PLANT:** same now() plant as V-06 → RED. (Two-widely-separated-clock/subprocess cell substituted by grep-gate + plant.) |
| **V-16** schema/API additivity | verified-sampled | `go build ./...` (0) — old call sites compile; no proto changed (no `make protogen` needed). | `v2.Grant` descriptor unchanged; generated code compiles. | Missing: `buf breaking` run (no proto delta in this stage beyond the additive `reader/v3` service). |
| **V-17** cost | measured/sampled (never closure) | NEW `TestV3ReadCostLinearNoPerGrantPointGet` (allocs 100→200 = **1.99×**, ceiling 2.6×) + `BenchmarkV3VsV2ListGrantsForEntitlement`. `go test -run TestV3ReadCost ./pkg/dotc1z/engine/pebble/`; `go test -bench BenchmarkV3VsV2 -run x ./pkg/dotc1z/engine/pebble/` | Allocations grow ~linearly in returned N (no hidden per-grant point-get / extra full pass); v3 shares the v2 scan minus the downshift. | Super-linear growth would breach the 2.6× ceiling. |
| **V-18** route completeness | verified-bounded | NEW `TestV3GrantReaderRouteCompleteness` (ServiceDesc method inventory vs checked-in `v3RouteCoverage` table; both directions). | Every descriptor method is classified timestamp-capable/legacy-only; no phantom table rows. | **PLANT:** `plant_missing_coverage_entry_is_caught` subtest — a descriptor route absent from the table is detected. |
| **V-19** concurrency | verified-sampled | NEW `TestV3ConcurrentReadersRace` (16 goroutines × mixed scan+point reads over one shared reader). `go test -race -run TestV3ConcurrentReadersRace ./pkg/dotc1z/engine/pebble/` | Each reader observes the exact planted `discovered_at`; `-race` clean. | Addendum §E (per-call arena, immutable sealed reads) justifies the single-cell reduction. |
| **V-20** premise gate | verified-bounded | NEW: raw-record assertion inside `TestV3ReadAbsentDiscoveredAtReturnsNilNeverNow` (`GetGrantRecord` returns nil `discovered_at` **before** the reader runs). | The absence fixture genuinely lacks a stored `discovered_at` (not filtered away) → the absence assertions can't pass vacuously. | If a write path had stamped, the premise gate fails first. |

## Gates (run under `PATH=$PATH:/go/bin`)

| Gate | Command | Result |
|---|---|---|
| build | `go build ./...` | PASS (exit 0) |
| tests | `go test ./pkg/connectorstore/... ./pkg/dotc1z/...` | PASS (exit 0; `pebble` 5.8s, `dotc1z` 43.5s) |
| race | `go test -race -run 'TestV3\|TestGetGrant\|TestStoreExpandedGrants' ./pkg/dotc1z/engine/pebble/...` | PASS (exit 0) |
| lint | `golangci-lint run ./pkg/dotc1z/engine/pebble/...` | 0 issues in changed files (2 pre-existing `nolintlint` warnings in `choke_point_meta_test.go`, not mine) |
| protogen | not run — no proto changed | N/A |

## New test instruments (all `package pebble`, uncommitted)

- `discovered_at_verify_absence_test.go` — V-06, V-15, V-20, V-12(decode).
- `discovered_at_verify_provenance_test.go` — V-08 (+V-01).
- `discovered_at_verify_reopen_test.go` — V-02, V-07, V-13(in-repo).
- `discovered_at_verify_routes_test.go` — V-18 (+V-10 compile asserts).
- `discovered_at_verify_cost_test.go` — V-17.
- `discovered_at_verify_errorfs_read_test.go` — V-12(IO fault).
- `discovered_at_verify_concurrency_test.go` — V-19.

## Residual gaps (stated as limits, per plan §6)

1. **Consumer-side deletion contract (c1 uplift) — TOP GAP, out of scope here.**
   The rung-5 consequence (overlay deletion on a zero/absent `discovered_at`)
   lives in another repo across the reader seam. SDK-side byte-exact read-back
   (verified above) is necessary but not sufficient; the decisive test —
   "deletion never fires on a fabricated/zero/absent value" — must live on the
   c1 side. Not faked here.
2. **Cross-BINARY compatibility (V-13/V-14).** Proven only for the same-build
   seal→reopen direction. A pinned OLD-SDK writer / committed golden corpus /
   old-binary matrix is not executable in this stage — declared gap, not faked.
3. **SQLite degradation reachability (V-11).** Provider is Pebble-only; SQLite
   lacks it → clean fallback. No executable SQLite probe built; real opens
   auto-convert to Pebble so native-SQLite degradation is largely unreachable.
4. **Non-injectable raw `os.*` read/reopen faults (V-12).** The errorfs seam
   covers vfs-level read faults; raw `os.*` outside the vfs stays a
   finding-in-waiting. Cancellation + iterator/fd-count reclamation cells for
   V-12 were not authored (structural: every read path `defer`s its
   iterator/closer — addendum §D).
5. **Domain/arrangement matrix (V-01/V-05/V-07).** Epoch/pre-epoch/equal-pair
   and the adjacent-key-opposite-timestamp alias arrangement, and the full page
   size {b−1,b,b+1} × route matrix, are sampled, not exhaustively enumerated.


</details>

kans (approved):

PR #1032, opened 2026-07-23
Sync lambda deadline preservable

kans:
## Treat DeadlineExceeded as preservable and fix grpc-timeout propagation

Execution timeouts (e.g. a serverless connector hitting its platform's hard execution limit) currently cause syncs to throw away all checkpointed progress, and the lambda gRPC transport's deadline-propagation path doesn't actually work. This PR fixes both, so timed-out syncs resume from their checkpoint instead of restarting from zero.

### Changes

**1. `pkg/sync`: `IsSyncPreservable` now treats timeouts as preservable**

The lambda transport already maps function timeouts to `codes.DeadlineExceeded` (a retryable, resumable condition), but `IsSyncPreservable` omitted it from the allowlist, so the c1z checkpoint was discarded on timeout. This adds:

- `codes.DeadlineExceeded` to the gRPC status allowlist.
- An `errors.Is(err, context.DeadlineExceeded)` check for wrapped bare context deadline errors, which carry no gRPC status (`status.FromError` does not map `context.DeadlineExceeded`; only `status.FromContextError` does).

**2. `pkg/lambda/grpc`: fix dead `grpc-timeout` check in the server**

`TimeoutForRequest` read the header with `metadata.MD.Get`, which returns `[]string`, then gated on `len(v) > 1` — requiring *two* header values. A single `grpc-timeout` header (the only realistic case) never applied, making the server-side timeout path dead code. Now `len(v) > 0`.

**3. `pkg/lambda/grpc`: client propagates the caller's deadline**

`clientConn.Invoke` now resolves the long-standing TODO: if the context has a deadline, it sets the `grpc-timeout` header (on a copy of the caller's metadata) so the server can bound the handler context; if the deadline has already expired it returns `codes.DeadlineExceeded` without invoking the function. Adds `encodeTimeout` (ported from grpc-go's `internal/grpcutil`, mirroring the existing `decodeTimeout`).

### Tests

- `IsSyncPreservable` unit tests: nil, `ErrSyncNotComplete`, `ErrTooManyWarnings`, `Unavailable`, `DeadlineExceeded` status, wrapped `context.DeadlineExceeded`, `Internal` (false), plain error (false).
- `TimeoutForRequest` regression tests: single header applies, no header, malformed value errors.
- `encodeTimeout`/`decodeTimeout` round-trips across unit ranges (1ns through 2,000,000h).
- `clientConn.Invoke`: header set when deadline present, absent otherwise, expired deadline short-circuits without calling the transport; caller metadata is not mutated.

`go build ./...`, `go test ./pkg/sync/... ./pkg/lambda/...`, and `make lint` all pass.

### Notes for reviewers

- This is the SDK half of the fix; the host-side error handling (preserving the `DeadlineExceeded` status through the invoker and persisting the c1z on timeout) lands separately.
- Deadline propagation is a correctness fix for

### DIFF
--- a/pkg/lambda/grpc/server.go
+++ b/pkg/lambda/grpc/server.go
@@ -226,7 +226,7 @@ func PeerForRequest(req *Request) *peer.Peer {
 
 func TimeoutForRequest(req *Request) (time.Duration, bool, error) {
 	v := req.Headers().Get("grpc-timeout")
-	if len(v) > 1 {
+	if len(v) > 0 {
 		to, err := decodeTimeout(v[0])
 		if err != nil {
 			return 0, false, status.Errorf(codes.Internal, "malformed grpc-timeout: %v", err)
--- a/pkg/lambda/grpc/client.go
+++ b/pkg/lambda/grpc/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/lambda"
@@ -121,9 +122,18 @@ func (c *clientConn) Invoke(ctx context.Context, method string, args any, reply 
 		return status.Errorf(codes.Unknown, "args and reply must satisfy proto.Message")
 	}
 
-	// TODO(morgabra): Should we do some of this stuff? (e.g. detect ctx deadline and set grpc-timeout, etc?)
-	// https://github.com/grpc/grpc-go/blob/9dc22c029c2592b5b6235d9ef6f14d62ecd6a509/internal/transport/http2_client.go#L541
 	md, _ := metadata.FromOutgoingContext(ctx)
+
+	// Propagate the context deadline to the server via the grpc-timeout header,
+	// mirroring grpc-go's HTTP/2 transport behavior.
+	if deadline, ok := ctx.Deadline(); ok {
+		timeout := time.Until(deadline)
+		if timeout <= 0 {
+			return status.Errorf(codes.DeadlineExceeded, "context deadline exceeded before invoking method %s", method)
+		}
+		md = md.Copy()
+		md.Set("grpc-timeout", encodeTimeout(timeout))
+	}
 
 	treq, err := NewRequest(method, req, md)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


pkg/lambda/grpc/server.go
```diff
@@ -226,7 +226,7 @@ func PeerForRequest(req *Request) *peer.Peer {
 
 func TimeoutForRequest(req *Request) (time.Duration, bool, error) {
 	v := req.Headers().Get("grpc-timeout")
-	if len(v) > 1 {
+	if len(v) > 0 {
```
ggreer, line 229: Why was this 1? We were only checking for multiple grpc-timeout headers?

ggreer (approved):

pkg/lambda/grpc/server.go
kans: In grpc-go, the header value is a string, and a valid timeout string always has len > 1 — minimum one digit plus one unit letter ("1n"). So "len greater than 1" is a real, correct predicate in the source material... applied to the string. In grpc-go's server the presence check and the length check are on different things: r.Header.Get returns a string, checked non-empty; decodeTimeout then enforces len >= 2 on it.

The baton-sdk port collapsed those layers onto the wrong type: req.Headers() is metadata.MD, whose Get returns []string, so v is a slice of header values, and len(v) > 1 silently changed meaning from "the timeout string is long enough to be valid" to "the client sent two or more grpc-timeout headers" — which no client ever does. The compiler couldn't object because v[0] is perfectly valid on a slice (and in fact only compiles with the slice — if v were a string, v[0] would be a byte and decodeTimeout(v[0]) wouldn't build). So the type system actively confirmed the wrong reading.

Whether the author consciously thought "string length" or just fat-fingered >= 1 into > 1, the outcome mechanics are the same, and the string-port theory explains why > 1 specifically — it's a correct constant from the neighboring function, attached to the wrong variable.

Why it survived: zero tests on the path (dead code is green code), no caller ever set the header (the client side had a TODO instead — so even a working server check would have had nothing to read), and the failure mode is pure absence: no timeout applied looks identical to no timeout requested. It's a tidy specimen of two patterns now in your BUG_CATCHING doc — the dead-capability dual of "a fallback that always fires," and "reading cannot execute the artifact": every reviewer who ever skimmed that function saw a plausible presence-check, and only a test that actually sent one header (the one we added last commit) could disagree.

PR #1023, opened 2026-07-21
Add new resource attributes to pebble.

ggreer:
Fixes an issue where pebble DBs lack the resource profile, status, and created_at attributes.

PR #1036, opened 2026-07-23
AI Deep Review

kans:
### What this is

Two files:

- **`docs/BUG_CATCHING.md`** — a self-contained rubric for reviewing changes in this repo, built from a retrospective across our recent efforts (fan-out, sync-replay, timeout preservability, incremental expansion, the verification-infrastructure phases). Every claim in it is backed by a shipped bug or a named effort; §7 is an index that resolves the shorthand.
- **`.claude/skills/ci-review.md`** — the CI AI reviewer's prompt gains a risk-triage step that mirrors the rubric's §2, so every PR gets scored and HIGH-risk changes get flagged for escalation before merge.

### The core arguments

1. **Review has no closure criterion.** It samples the behavior space with no coverage model, so "review converged" is undefined — the sync-replay effort ran 15+ rounds without converging, and the fan-out branch produced new high-severity bugs for 5 straight rounds. The rubric budgets two to three rounds, then routes to instruments (differential oracles, cross-version harnesses, fault injection, permutation tables). Review's real job is discovering the verification plan, not substituting for one.

2. **Risk is escape × consequence.** Triage scores four escape axes (silence, durability, uncontrolled dimensions — including data volume, consumer distance) against a remediation ladder (redeploy → re-sync → migrate → coordinate → irrecoverable). Silent+durable compounds with detection latency; version-pair dependence is auto-HIGH; ladder rung ≥ 4 is auto-HIGH and flips the defense from detection to prevention.

3. **Cost contracts.** On a declared list of hot paths (grant expansion, compaction, the per-checkpoint loop, artifact-open-time work) performance is a correctness property: budgets convert latency into failure and fixed memory converts peak-RSS into OOM livelock. The reviewable object is the cost-curve delta; the enforcement is a whale-scale benchmark gate. Migrations are named as the intersection class that hits every risk factor at once.

4. **Fix-time obligations.** Failing test first for every confirmed bug, and the recurrence rule: a bug in a class this document already names must ship the ladder climb (§4), not just the patch — evidenced by converter field loss recurring 19 days after its point fix (#997 → #1023).

### How it's meant to be used

The doc opens with a routing guide and §6 contains a slicing map: orchestrating agents inline only the sections a pass agent needs, rather than pasting the whole document. The CI reviewer does triage-and-route only — it flags HIGH, names the review-blind class and the instrument, and checks whether the PR contains it; it does not attempt the full pass set itself.

### Notes for review

- Docs and a prompt file only; no code changes.
- The doc deliberately carries mild redundancy across sections — sliced readers never see the other copies. Please don't flag duplication as a defect.
- PR references (#911, #997, #1005, #1015–#1017, #1021–#1024) resolve via `git log --grep`.

### DIFF
--- a/.claude/skills/ci-review.md
+++ b/.claude/skills/ci-review.md
@@ -9,6 +9,63 @@ every repo that imports the SDK, every serialized sync state record that gets re
 and every generated wire type. Review exported APIs, proto fields, serialized state,
 defaults, and dependency changes as compatibility surfaces.
 
+## Risk Triage And Escalation (do this first)
+
+Before the detailed review, score the change per `docs/BUG_CATCHING.md` §2:
+risk is escape times consequence. Score the failure mode, not the subsystem.
+
+Escape axes — how likely a defect ships and persists:
+
+1. **Silence** — if this change is wrong, what makes noise? A panic or a failed
+   request is low; a well-formed wrong row, a silently skipped step, or an absent
+   record is high.
+2. **Durability** — does any effect outlive the process (c1z contents, sync tokens,
+   session values, proto wire types, exported config schemas)? Durable output will
+   be read by SDK versions that do not exist yet.
+3. **Uncontrolled dimensions** — does correctness or liveness depend on goroutine
+   schedule, crash or checkpoint timing, environment faults (network, disk, Lambda
+   freeze/thaw), which SDK version wrote the data being read, or data volume?
+   Scale is the dimension local testing undersamples: O(n^2) is correct at every
+   fixture size and dead at whale scale.
+4. **Consumer distance** — who interprets the output: the same package, another
+   process, the c1 platform, downstream connectors, or a future SDK version?
+
+Consequence — place the worst credible failure on the remediation ladder:
+(1) redeploy, (2) re-sync, (3) migrate every artifact in the fleet, (4) coordinate
+a contract change across repos and deployments we do not control, (5) irrecoverable
+external side effects (provisioning writes, credential operations, deletions).
+
+Verdict rules: silent + durable is HIGH (and compounds — detection latency grows
+the migration bill). Any version-pair dependence is HIGH. Remediation rung 4 or 5
+is HIGH regardless of escape score, and the ask shifts to prevention: golden
+artifact corpus for format changes, dry-run/idempotency for external effects.
+Otherwise, two or more escape yes answers is HIGH. Default-path changes outrank
+option-gated ones.
+
+Cost contracts: if the PR touches grant expansion, compaction, the per-checkpoint
+loop, or anything that runs at artifact-open time (including migrations), require
+a stated cost-curve delta (big-O in grants/entitlements/graph size) and a
+benchmark that enforces it. A missing benchmark on these paths is a finding.
+Open-time migrations must be chunked and resumable; an over-budget migration
+retries the same artifact into a permanent every-sync failure.
+
+Lead your review with a short triage block — each axis yes/no with a one-line
+reason, then the verdict. For HIGH verdicts, recommend escalation beyond this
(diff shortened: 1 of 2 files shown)


ggreer (approved):

.claude/skills/ci-review.md
```diff
@@ -9,6 +9,63 @@ every repo that imports the SDK, every serialized sync state record that gets re
 and every generated wire type. Review exported APIs, proto fields, serialized state,
 defaults, and dependency changes as compatibility surfaces.
 
+## Risk Triage And Escalation (do this first)
+
+Before the detailed review, score the change per `docs/BUG_CATCHING.md` §2:
+risk is escape times consequence. Score the failure mode, not the subsystem.
+
+Escape axes — how likely a defect ships and persists:
```
gontzess, line 17: Should we specify the review agent to delegate a sub agent for each axis? Wonder if that improves depth. Maybe we could a/b test some stuff like this with a complex pr example

docs/BUG_CATCHING.md
```diff
@@ -0,0 +1,784 @@
...
+
+Production complement, not substitute: runtime positive-evidence invariants and
+reconciliation ledgers (§4 rung 3) make a slice of the silent class loud in
+operation, but liveness failures and anything they don't ledger stay invisible —
+the verification harness is the only instrument for the rest.
+
+## 3. The seven mandatory passes
+
+Every non-trivial change gets each of these as a distinct pass. Separate,
+separately-prompted reviewer agents per pass work well (see §6). Check §2's routing
```
gontzess, line 262: Ah does this wording lead to sub agent spawning reliably?

docs/BUG_CATCHING.md
```diff
@@ -0,0 +1,784 @@
+# Bug-catching guidance for baton-sdk
```
gontzess, line 1: Are we worried about the length of this file? Wonder if the 7 passes should be split into sub pages

gontzess (approved):
Have you done some test runs with this? Would be good to a/b test a few things. Regardless this seems great

kans:
@gontzess 

> Have you done some test runs with this? Would be good to a/b test a few things. Regardless this seems great

Yeah, actually.  I tested it against 3 active PRs, including two active PRs  from the resync epic (see lack of convergence), the [incremental grant expansion pr](this https://github.com/ConductorOne/baton-sdk/pull/1013,) and the last 15 or so PRs.  The approaches are generally what I learned shipping the perf improvements work (compaction/expansion), getting pebble to go, and from the first attempt at resync, which crashed and burned.  Its also where [this pr](https://github.com/ConductorOne/baton-sdk/pull/1035) came from.

docs/BUG_CATCHING.md
kans: Yes, we are, particularly in the case that you want to give different subagents different tasks.  Thats loading a lot of the same context repeatedly.

docs/BUG_CATCHING.md
kans: Unverified.  I typically do this manually.

gontzess:
> @gontzess 
> 
> > Have you done some test runs with this? Would be good to a/b test a few things. Regardless this seems great
> 
> Yeah, actually.  I tested it against 3 active PRs, including two active PRs  from the resync epic (see lack of convergence), the [incremental grant expansion pr](this https://github.com/ConductorOne/baton-sdk/pull/1013,) and the last 15 or so PRs.  The approaches are generally what I learned shipping the perf improvements work (compaction/expansion), getting pebble to go, and from the first attempt at resync, which crashed and burned.  Its also where [this pr](https://github.com/ConductorOne/baton-sdk/pull/1035) came from. 

Awesome thanks!

PR #1037, opened 2026-07-24
uhttp: recognize application/soap+xml (SOAP 1.2) as XML content type

cvkramer:
## Summary
Recognize `application/soap+xml` (SOAP 1.2, RFC 3902) as an XML content type in `pkg/uhttp`, so `WithGenericResponse` auto-routes SOAP 1.2 responses to the XML parser. Also harden `IsXMLContentType` to match the base media type exactly (splitting off `;` parameters) instead of `strings.HasPrefix` — this tolerates parameters like `; charset=utf-8; action=...` and fixes a pre-existing bug where lookalikes such as `application/soap+xmlrpc` were misclassified as XML.

## Changes
- `pkg/uhttp/contenttype.go`: add `application/soap+xml`; exact base-type comparison.

## Tests
- `pkg/uhttp/contenttype_test.go`: positive (soap+xml, with charset/action params, text/xml, application/xml), negative (empty, text/html, text/plain, application/json), and lookalike-boundary cases.
- `go test ./pkg/uhttp/...` and `go build ./...` pass.

## Validation
Exercised end-to-end against a live SOAP 1.2 server (`application/soap+xml`): responses parsed and synced by baton-http.

Part of CXH-2134. Resolves CXH-2135.

### DIFF
--- a/pkg/uhttp/contenttype.go
+++ b/pkg/uhttp/contenttype.go
@@ -7,6 +7,7 @@ import (
 var xmlContentTypes []string = []string{
 	"text/xml",
 	"application/xml",
+	"application/soap+xml",
 }
 
 func IsJSONContentType(contentType string) bool {
--- a/pkg/uhttp/contenttype.go
+++ b/pkg/uhttp/contenttype.go
@@ -20,8 +20,15 @@ func IsXMLContentType(contentType string) bool {
 	// there are some janky APIs out there
 	normalizedContentType := strings.TrimSpace(strings.ToLower(contentType))
 
+	// Only compare against the media type itself, ignoring any parameters
+	// (e.g. "; charset=utf-8; action=..."), so that lookalike media types
+	// like "application/soap+xmlrpc" don't get misclassified as XML just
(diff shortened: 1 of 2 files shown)


pkg/uhttp/contenttype.go
```diff
@@ -19,8 +20,15 @@ func IsXMLContentType(contentType string) bool {
...
+	// Only compare against the media type itself, ignoring any parameters
+	// (e.g. "; charset=utf-8; action=..."), so that lookalike media types
+	// like "application/soap+xmlrpc" don't get misclassified as XML just
+	// because they share a prefix with "application/soap+xml".
+	baseContentType, _, _ := strings.Cut(normalizedContentType, ";")
+	baseContentType = strings.TrimSpace(baseContentType)
+
 	for _, xmlContentType := range xmlContentTypes {
-		if strings.HasPrefix(normalizedContentType, xmlContentType) {
+		if baseContentType == xmlContentType {
```
kans, line 31: This will break some valid XMLs I think - `application/xml-dtd`.  Can we special case soap?

PR #1041, opened 2026-07-28
fix: checkpoint sync progress on SIGTERM before exit

FeliLucero1:
## Summary

- Catch `syscall.SIGTERM` (not just `os.Interrupt`) in `connectorrunner.Run()`, routing it through the existing graceful `cancel(ErrSigTerm)` path. Without this, an orchestrator-sent SIGTERM (Kubernetes rollout/drain/spot-reclaim, `docker stop`) terminated the process immediately with no chance to checkpoint in-flight progress. (Correction: the process dies immediately regardless of PID 1 status — the Go runtime installs its own handler for SIGTERM at startup, so the kernel's PID-1 default-action-suppression rule never actually applies to a Go binary. The observable symptom — abrupt exit, lost progress — is right, only the mechanism I originally described was wrong. Thanks @sergiocorral-conductorone.)
- **Drain in-flight tasks before `run()` returns on cancellation.** In the long-lived (non-`oneShot`) mode every task is dispatched to an untracked goroutine (`go func(t *v1.Task)`), and every caller of this runner closes the connector client (`Close()` → `c.cw.Close()`) immediately after `Run()` returns. Trapping SIGTERM alone doesn't help if nothing waits: `run()` would return the instant `ctx.Done()` fires, `Close()` would tear the connector client down, and the in-flight sync's forced checkpoint write could be racing that teardown. `run()` now calls `drainInFlightTasks` before returning on cancellation, which acquires the full semaphore weight on a **detached, 25s-bounded** context (`shutdownDrainTimeout`) — i.e. waits for every dispatched task goroutine to actually finish. New test `TestRun_DrainsInFlightTaskBeforeReturningOnCancellation` uses a task manager whose `Process` ignores `ctx.Done()` (mirroring a real checkpoint write on a detached context) and asserts `run()` doesn't return until it's released; fails without the drain, passes with it.
- Unify `parallel_syncer.go`'s cancellation handling so **any** cancellation of `runCtx` — a deadline timeout, a SIGTERM/SIGINT shutdown, a server-directed task cancellation (`ErrTaskCancelled`), a heartbeat failure (`ErrTaskHeartbeatFailed`), or any other cause — forces a checkpoint and returns a resumable `ErrSyncNotComplete`-wrapped error. This widens the guard on the existing `handleOperationError` helper (added by the concurrent Phase 5 merge, #1029, to fix the same class of bug for the deadline-only case) from `errors.Is(cause, context.DeadlineExceeded)` to any non-nil cause.
- **Every mid-action return point already routed through `handleOperationError`** (that wiring is Phase 5's, not this PR's — see the attribution note below); this PR's actual fix is generalizing what the helper does with a non-deadline cause, plus adding the one call site Phase 5 didn't cover (`MarkSyncSupportsDiff`, `parallel_syncer.go`), plus closing the two residual bypass windows below. New test `TestSyncResumesAfterMidActionCancellation` uses a connector mock that actually observes `ctx` cancellation mid-call (the base test mock ignores `ctx` entirely) — fails without the widened guard, passes with it.
- **Preserve both the cancellation cause and the in-flight operation's own error**, instead of dropping them. `handleOperationError` now returns `errors.Join(checkpointErr, ErrSyncNotComplete, cause, batchErr)` — `errors.Is(err, ErrSyncNotComplete)` still classifies it as preservable, while a caller can separately recover *why* (`errors.Is(err, cause)`, e.g. `pkg/tasks/c1api` distinguishing `ErrTaskCancelled` from a plain shutdown) and *what the interrupted operation actually returned* (`batchErr` — a genuine store/connector failure that happens to land in the same window is no longer silently swallowed on the next attempt).
- **Close two residual windows where cancellation still bypassed the helper.** The loop's throttled, non-forced `Checkpoint(ctx, false)` ran on the raw `ctx` *before* the `runCtx.Done()` check (now reordered so cancellation is checked first) — and even after that reorder, the check itself only ran once per loop iteration, so a cancellation landing in the small gap between the check and this same checkpoint call could still hit the raw ctx and return unclassified. That checkpoint call (and the three similar forced `Checkpoint(ctx, true)` calls in the `InitOp` arm) now route their errors through `handleOperationError` too, which is a no-op when `runCtx` wasn't actually cancelled.
- **Bound the forced checkpoint write.** `context.WithoutCancel(ctx)` strips the deadline as well as the cancellation, so an unbounded write here could stall for the connector's entire termination grace period and still get SIGKILLed. It's now wrapped in a 15s timeout (`forcedCheckpointTimeout`), under Kubernetes' 30s default grace period and under the runner's own 25s drain budget.
- **Restored the sync-stats summary log for a non-deadline cancellation.** Before this PR, any non-deadline cancellation returned a raw, non-`ErrSyncNotComplete` error, so `returnSyncError` logged the full stats summary (per-step durations, per-resource-type counters, retry/rate-limit wall time) on its normal path. Once the returned error satisfies `errors.Is(err, ErrSyncNotComplete)`, that logging is skipped — so a SIGTERM-interrupted sync was logging *less* diagnostic output than before this PR, using the same counters that reconstructed the CXH-2121 timeline in the first place. The non-deadline branch now mirrors the deadline branch's `s.recordStats` summary.

All of the above (drain, checkpoint-ordering residual window, batchErr preservation, stats-summary regression, PID-1 mechanism correction) were raised in a second review pass by @sergiocorral-conductorone, several independently confirmed by an automated verifier. I verified each against the code before fixing it — thank you for the thoroughness.

### Attribution correction

An earlier revision of this PR body over-credited itself: it claimed "all ~12 mid-action return points now route through a new helper," implying this PR wired that routing. It didn't — those call sites already existed on `main` (added by the Phase 5 merge, #1029, with `handleOperationError` already in place but gated to `context.DeadlineExceeded` only). This PR's actual contribution is generalizing that helper's guard to any cancellation cause (which is what actually fixes the bypass sergiocorral found), adding the one call site Phase 5 missed, and the residual-window/preservation/logging fixes above. Also fixed leftover references to an earlier local name for the helper (`finishOnCancellation`) in a test file comment — the function is `handleOperationError` throughout.

### Wire-visible behavior change worth flagging before merge

Joining `cause` and `batchErr` into the returned error changes what `pkg/tasks/c1api/manager.go`'s `finishTask` reports to C1, in two different ways depending on what's in the join:

- None of `cause` (`ErrSigTerm` / `ErrTaskCancelled` / `ErrTaskHeartbeatFailed` / `context.Canceled` / `context.DeadlineExceeded`) or `ErrSyncNotComplete` implement `GRPCStatus()`, so when `batchErr` is nil (the top-of-loop check) or itself doesn't implement it either, `status.FromError(taskError)` returns `ok=false` and falls to the sentinel switch, which now matches via `errors.Is` — `codes.Canceled` for the SIGTERM path, `codes.DeadlineExceeded` for the run-duration path — instead of the previous `codes.Unknown`.
- When a mid-action cancellation's `batchErr` *does* implement `GRPCStatus()` (the interrupted connector call itself returned a proper gRPC status, e.g. a rate-limit or transport error), `status.FromError`'s `errors.As` traversal finds it directly inside the joined tree and returns that code instead, bypassing the sentinel switch entirely.

Both directions look strictly more accurate than the previous blanket `codes.Unknown`, and `NonRetryable` classification is unaffected either way, but whether C1's task engine treats `codes.Canceled` (or whatever code `batchErr` surfaces) as retryable is outside this repo — worth confirming with whoever owns that before merge, not discovering it as a behavior change in the hosted fleet. (Flagged in two passes: @sergiocorral-conductorone for the `cause` side, the automated reviewer for the `batchErr` side.)

## Context

Investigated via a large production sync (baton-azure-devops, CXH-2121) showing ~9.6h of a 16.6h run unaccounted for: the `user` grants-sync counter reset to 1 ten times without ever completing. Only 3 of those resets corresponded to a logged, checkpoint-safe `"sync run duration has expired"` exit — the rest had no graceful-exit log at all, consistent with unhandled SIGTERM kills losing in-flight progress.

Caveats worth stating plainly, also raised in review:

- **The legacy SQLite engine (`C1File`) has a wider version of the same problem**, unaddressed here: it uses `ctx` directly in per-action store calls (e.g. `GetResource`), not just the checkpoint write, so a cancellation landing mid-action can still bypass the resumable classification there even after this PR. Production runs the Pebble engine, where the store calls this PR touches don't have that problem, but I have not audited the SQLite engine's other per-action call sites beyond confirming the checkpoint path.
- ~~I have not confirmed which task-handling path CXH-2121's connector actually runs through~~ **Resolved via @Bencheng21** (owner of the Temporal/task-engine side): the actual path is a Temporal activity in `ductone/c1`'s `sync_v0.go`, not `pkg/tasks/c1api` or `pkg/tasks/local` in this repo. Per his analysis in [ductone/c1#22446](https://github.com/ductone/c1/pull/22446) (merged 2026-08-01), that activity's upload logic checks `errors.Is(err, ErrSyncNotComplete)` **directly**, independent of the separate FF-gated `IsSyncPreservable` path (which he notes is currently broken — its feature-flag check fails on every attempt, falling back to "legacy behavior"). That means once this SDK version is adopted, `handleOperationError`'s unconditional `ErrSyncNotComplete` on *any* cancellation cause — not just `context.DeadlineExceeded` — should be picked up by that existing direct check without needing the FF path fixed. His PR is a complementary, not redundant, fix: it catches worker-drain specifically via Temporal's own `activity.GetWorkerStopChannel(ctx)` signal (works regardless of how baton-sdk classifies the error), while this PR fixes the classification at the SDK layer for every cancellation cause — including ones his drain-specific check doesn't cover, like a heartbeat timeout without an active `RecordHeartbeat`, which his own PR description lists as a still-open "companion candidate" for baton-sdk. This PR is that candidate. (I can't read `sync_v0.go` directly — this is inferred from his PR's own classification table and description, not independently verified against that file.)
- **The drain budget (25s) can't accommodate `syncer.Close()`'s own detached finalize**, which is bounded separately by `dotc1z.FinalizeTimeout` (1h default) and runs on every path including sync failure. The drain in this PR only targets getting the smaller, higher-priority forced checkpoint write (15s) safely persisted before the connector client is torn down — it does not, and cannot within any realistic termination grace period, guarantee the full c1z finalize/close tail completes too.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/sync/... ./pkg/connectorrunner/...`
- [x] `go test ./pkg/sync/... ./pkg/connectorrunner/... -count=1` — full suite green
- [x] `go test ./pkg/connectorrunner/... -race -count=1` — race-clean
- [x] `TestSyncResumesAfterLoopBoundaryCancellation` — cancellation observed between actions.
- [x] `TestSyncResumesAfterMidActionCancellation` — cancellation observed inside an in-flight connector call, via a mock that actually respects `ctx`. Fails on the narrow (deadline-only) guard, passes with the widened one.
- [x] `TestRun_DrainsInFlightTaskBeforeReturningOnCancellation` — asserts `run()` waits for an in-flight task goroutine to finish before returning on cancellation. Fails without the drain, passes with it.

### DIFF
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/conductorone/baton-sdk/pkg/bid"
@@ -138,7 +139,7 @@ func (c *connectorRunner) Run(ctx context.Context) error {
 	}
 
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 	go func() {
 		for range sigChan {
 			cancel(ErrSigTerm)
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -72,9 +73,16 @@ const shutdownDrainTimeout = 25 * time.Second
 // drainInFlightTasks waits for every currently-held semaphore slot to be
 // released — i.e. every dispatched task goroutine (see the `go func(t
 // *v1.Task)` dispatch below) to return — bounded by shutdownDrainTimeout.
-// Called right before run() returns on cancellation, so the caller's
-// deferred Close() (which tears down the connector client) doesn't race an
-// in-flight sync that's still writing its forced checkpoint.
+// Deferred once, right after sem is created, so it runs on every exit from
+// run() — including the cancellation-driven returns below, and the
+// stopForLoop exit further down, which falls out of the loop with no select
+// case involved at all. Covering every return this way, rather than one
+// call per return site, is deliberate: this loop has already had a return
+// path added without a matching drain call once (a task goroutine setting
+// stopForLoop bypassed it entirely), and a per-site call is exactly the
+// shape of bug that keeps recurring. Whichever return runs, the caller's
+// deferred Close() (which tears down the connector client) must not race
+// an in-flight sync that's still writing its forced checkpoint.
 func (c *connectorRunner) drainInFlightTasks(ctx context.Context, sem *semaphore.Weighted, l *zap.Logger) {
 	drainCtx, cancelDrain := context.WithTimeout(context.WithoutCancel(ctx), shutdownDrainTimeout)
 	defer cancelDrain()
@@ -247,15 +255,22 @@ func (c *connectorRunner) run(ctx context.Context) error {
 	}
 
 	sem := semaphore.NewWeighted(int64(c.taskConcurrency))
+	// Drain unconditionally on every exit from this function, not just the
+	// two cancellation-driven returns below: a dispatched task goroutine can
+	// also end this loop by setting stopForLoop (see the "grpc: the client
+	// connection is closing" check further down), which falls through to the
+	// post-loop return with no select case involved at all. A per-return-site
+	// call is exactly the shape of bug this already was once -- a defer here
+	// covers every current and future exit path from run() by construction.
+	defer c.drainInFlightTasks(ctx, sem, l)
 
 	nextCheckAfter := time.Second * 0
 	errCount := 0
-	stopForLoop := false
+	var stopForLoop atomic.Bool
 	var err error
-	for !stopForLoop {
+	for !stopForLoop.Load() {
 		select {
 		case <-ctx.Done():
-			c.drainInFlightTasks(ctx, sem, l)
 			return c.handleContextCancel(ctx)
 		case <-time.After(nextCheckAfter):
 			l.Debug("runner: claiming worker")
@@ -263,7 +278,6 @@ func (c *connectorRunner) run(ctx context.Context) error {
 			err = sem.Acquire(ctx, 1)
 			if err != nil {
 				l.Error("runner: error acquiring semaphore to claim worker", zap.Error(err))
-				c.drainInFlightTasks(ctx, sem, l)
 				return c.handleContextCancel(ctx)
 			}
 			l.Debug("runner: worker claimed, checking for next task")
@@ -320,7 +334,7 @@ func (c *connectorRunner) run(ctx context.Context) error {
 				err := c.processTask(ctx, t)
 				if err != nil {
 					if strings.Contains(err.Error(), "grpc: the client connection is closing") {
-						stopForLoop = true
+						stopForLoop.Store(true)
 					}
 					l.Error("runner: error processing task", zap.Error(err), zap.String("task_id", t.GetId()), zap.String("task_type", tasks.GetType(t).String()))
 					return
@@ -332,7 +346,7 @@ func (c *connectorRunner) run(ctx context.Context) error {
 		}
 	}
 
-	if stopForLoop {
+	if stopForLoop.Load() {
 		return fmt.Errorf("unable to communicate with gRPC server")
 	}
 
--- a/pkg/connectorrunner/runner.go
+++ b/pkg/connectorrunner/runner.go
@@ -49,6 +49,42 @@ type connectorRunner struct {
 
 var ErrSigTerm = errors.New("context cancelled by process shutdown")
 
+// shutdownDrainTimeout bounds how long run() waits, on cancellation, for
+// in-flight task goroutines to finish before returning. Each task goroutine
+// is untracked (dispatched via a bare `go func`) and Close() — called
+// immediately after Run() returns by every caller of this runner — tears
+// down the connector client via c.cw.Close() with no synchronization against
+// those goroutines. Without a drain, a SIGTERM can still race the very
+// checkpoint this package exists to protect: Run() returns as soon as ctx is
+// Done, Close() runs, and the in-flight sync's forced checkpoint write (see
+// sync/parallel_syncer.go's handleOperationError, bounded to 15s) can be torn
+// out from under it.
+//
+// This budget intentionally only targets that checkpoint write, not a full
+// c1z finalize: syncer.Close() (called by the task handler on every path,
+// including sync failure) bounds its own detached finalize by
+// dotc1z.FinalizeTimeout, which defaults to 1 hour — no realistic
+// termination grace period can wait that out, and this drain does not try
+// to. It only aims to let the smaller, higher-priority checkpoint write
+// complete before the connector client is torn down.
+const shutdownDrainTimeout = 25 * time.Second
+
+// drainInFlightTasks waits for every currently-held semaphore slot to be
+// released — i.e. every dispatched task goroutine (see the `go func(t
+// *v1.Task)` dispatch below) to return — bounded by shutdownDrainTimeout.
+// Called right before run() returns on cancellation, so the caller's
+// deferred Close() (which tears down the connector client) doesn't race an
+// in-flight sync that's still writing its forced checkpoint.
+func (c *connectorRunner) drainInFlightTasks(ctx context.Context, sem *semaphore.Weighted, l *zap.Logger) {
+	drainCtx, cancelDrain := context.WithTimeout(context.WithoutCancel(ctx), shutdownDrainTimeout)
+	defer cancelDrain()
+	if err := sem.Acquire(drainCtx, int64(c.taskConcurrency)); err != nil {
+		l.Warn("runner: shutdown drain timed out; in-flight tasks may not have finished checkpointing", zap.Error(err))
+		return
+	}
+	sem.Release(int64(c.taskConcurrency))
+}
+
 // setupPersistentLog ensures that a log file on disk is created,
 // when required by either the stored Manager or by a Task.
 // A log file created by a stored Manager persists for our entire run,
@@ -219,6 +255,7 @@ func (c *connectorRunner) run(ctx context.Context) error {
 	for !stopForLoop {
 		select {
 		case <-ctx.Done():
+			c.drainInFlightTasks(ctx, sem, l)
 			return c.handleContextCancel(ctx)
 		case <-time.After(nextCheckAfter):
 			l.Debug("runner: claiming worker")
@@ -226,6 +263,7 @@ func (c *connectorRunner) run(ctx context.Context) error {
 			err = sem.Acquire(ctx, 1)
 			if err != nil {
 				l.Error("runner: error acquiring semaphore to claim worker", zap.Error(err))
+				c.drainInFlightTasks(ctx, sem, l)
 				return c.handleContextCancel(ctx)
 			}
 			l.Debug("runner: worker claimed, checking for next task")
--- a/pkg/sync/parallel_syncer.go
+++ b/pkg/sync/parallel_syncer.go
@@ -176,23 +176,28 @@ func (s *syncer) parallelSync(
 		select {
 		case <-runCtx.Done():
 			err = context.Cause(runCtx)
-			switch {
-			case errors.Is(err, context.DeadlineExceeded):
+			if errors.Is(err, context.DeadlineExceeded) {
 				if s.recordStats {
 					l.Info("sync run duration has expired, exiting sync early", s.syncSummaryFields(trace.SpanFromContext(ctx))...)
 				} else {
 					l.Info("sync run duration has expired, exiting sync early", zap.String("sync_id", s.syncID))
 				}
-				// It would be nice to remove this once we're more confident in the checkpointing logic.
-				checkpointErr := s.Checkpoint(ctx, true)
-				if checkpointErr != nil {
-					l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
-				}
-				return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
-			default:
-				l.Error("sync context cancelled", zap.String("sync_id", s.syncID), zap.Error(err))
-				return warnings, err
+			} else {
+				// Any other cancellation (e.g. SIGTERM/SIGINT via connectorrunner.ErrSigTerm) is just as
+				// resumable as a deadline timeout, so it gets the same forced checkpoint and treatment below.
+				l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(err))
 			}
+			// ctx itself may already be Done here (e.g. a SIGTERM-driven cancellation
+			// cancels ctx directly, not just runCtx's derived deadline), so this
+			// checkpoint must run on a context that survives that cancellation —
+			// same pattern as the other finalize/cleanup writes in this codebase
+			// (c1file.go, clone_sync.go, syncer.go) that must complete after ctx is done.
+			// It would be nice to remove this once we're more confident in the checkpointing logic.
+			checkpointErr := s.Checkpoint(context.WithoutCancel(ctx), true)
+			if checkpointErr != nil {
+				l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
+			}
+			return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
 		default:
 		}
 
--- a/pkg/sync/parallel_syncer.go
+++ b/pkg/sync/parallel_syncer.go
@@ -161,6 +161,19 @@ func (s *syncer) parallelSync(
 			break
 		}
 
+		// Checked before the throttled checkpoint below: that checkpoint runs
+		// on the raw ctx, which may already be Done (a SIGTERM-driven
+		// cancellation cancels ctx itself, not just runCtx's derived
+		// deadline) -- checking cancellation first means a cancelled ctx
+		// never gets a chance to fail that write and return a raw,
+		// non-resumable error before this iteration reaches
+		// handleOperationError.
+		select {
+		case <-runCtx.Done():
+			return s.handleOperationError(ctx, runCtx, warnings, nil)
+		default:
+		}
+
 		err := s.Checkpoint(ctx, false)
 		if err != nil {
 			return warnings, err
@@ -172,33 +185,6 @@ func (s *syncer) parallelSync(
 			if tooManyWarnings(len(warnings), completedActionsCount) {
 				return warnings, fmt.Errorf("%w: warnings: %v completed actions: %d", ErrTooManyWarnings, warnings, completedActionsCount)
 			}
-		}
-		select {
-		case <-runCtx.Done():
-			err = context.Cause(runCtx)
-			if errors.Is(err, context.DeadlineExceeded) {
-				if s.recordStats {
-					l.Info("sync run duration has expired, exiting sync early", s.syncSummaryFields(trace.SpanFromContext(ctx))...)
-				} else {
-					l.Info("sync run duration has expired, exiting sync early", zap.String("sync_id", s.syncID))
-				}
-			} else {
-				// Any other cancellation (e.g. SIGTERM/SIGINT via connectorrunner.ErrSigTerm) is just as
-				// resumable as a deadline timeout, so it gets the same forced checkpoint and treatment below.
-				l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(err))
-			}
-			// ctx itself may already be Done here (e.g. a SIGTERM-driven cancellation
-			// cancels ctx directly, not just runCtx's derived deadline), so this
-			// checkpoint must run on a context that survives that cancellation —
-			// same pattern as the other finalize/cleanup writes in this codebase
-			// (c1file.go, clone_sync.go, syncer.go) that must complete after ctx is done.
-			// It would be nice to remove this once we're more confident in the checkpointing logic.
-			checkpointErr := s.Checkpoint(context.WithoutCancel(ctx), true)
-			if checkpointErr != nil {
-				l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
-			}
-			return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
-		default:
 		}
 
 		switch stateAction.Op {
@@ -405,7 +391,7 @@ func (s *syncer) parallelSync(
 				}
 				if err := s.store.SyncMeta().MarkSyncSupportsDiff(ctx, s.syncID); err != nil {
 					l.Error("failed to set supports_diff marker", zap.Error(err))
-					return warnings, err
+					return s.handleOperationError(ctx, runCtx, warnings, err)
 				}
 			}
 
@@ -427,29 +413,78 @@ func (s *syncer) parallelSync(
 	return warnings, nil
 }
 
+// forcedCheckpointTimeout bounds the detached (context.WithoutCancel) context
+// used for the forced checkpoint write below. This runs precisely during an
+// orchestrator's termination grace period, so it must stay well short of a
+// typical grace period (Kubernetes defaults to 30s) — unlike
+// dotc1z.FinalizeTimeout, which bounds the much heavier WAL
+// checkpoint+save+upload tail, this is a single small state write.
+const forcedCheckpointTimeout = 15 * time.Second
+
+// handleOperationError checks whether runCtx has been cancelled — by a
+// deadline timeout, a SIGTERM/SIGINT shutdown (connectorrunner.ErrSigTerm),
+// a server-directed task cancellation, a heartbeat failure, or any other
+// cause — and if so treats it uniformly as resumable: force a checkpoint so
+// the next resume picks up from here instead of restarting, and return an
+// error that both IsSyncPreservable recognizes (via ErrSyncNotComplete) and
+// still lets a caller distinguish *why* via errors.Is(err, cause).
+//
+// batchErr is the error an in-flight operation returned (nil when called
+// from the top-of-loop check, where cancellation was observed between
+// actions rather than surfaced by a connector call). If runCtx was not
+// actually cancelled, batchErr is a genuine, unrelated operation error and
+// is returned unchanged — this function only special-cases cancellation.
 func (s *syncer) handleOperationError(
 	ctx context.Context,
 	runCtx context.Context,
 	warnings []error,
 	batchErr error,
 ) ([]error, error) {
-	if !errors.Is(context.Cause(runCtx), context.DeadlineExceeded) {
+	cause := context.Cause(runCtx)
+	if cause == nil {
 		return warnings, batchErr
 	}
-	// The run duration expired. The operation error is usually just the
-	// resulting cancellation, but a genuine connector/store failure can
-	// land in the same window — log it so expiry never masks a real bug.
-	// The sync resumes from the checkpoint, so the work is retried either
-	// way; only ErrSyncNotComplete is surfaced to keep the resumable
-	// contract for callers.
+	// The operation error is usually just the resulting cancellation, but a
+	// genuine connector/store failure can land in the same window — log it
+	// so cancellation never masks a real bug. The sync resumes from the
+	// checkpoint, so the work is retried either way; only ErrSyncNotComplete
+	// (plus cause) is surfaced to keep the resumable contract for callers.
 	if batchErr != nil && !errors.Is(batchErr, context.Canceled) && !errors.Is(batchErr, context.DeadlineExceeded) {
 		ctxzap.Extract(ctx).Error(
-			"sync operation failed while run duration expired; exiting early for resume",
+			"sync operation failed while the sync was being cancelled; exiting early for resume",
 			zap.Error(batchErr),
+			zap.NamedError("cancel_cause", cause),
 		)
 	}
-	checkpointErr := s.Checkpoint(ctx, true)
-	return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
+	l := ctxzap.Extract(ctx)
+	if errors.Is(cause, context.DeadlineExceeded) {
+		if s.recordStats {
+			l.Info("sync run duration has expired, exiting sync early", s.syncSummaryFields(trace.SpanFromContext(ctx))...)
+		} else {
+			l.Info("sync run duration has expired, exiting sync early", zap.String("sync_id", s.syncID))
+		}
+	} else {
+		l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(cause))
+	}
+	// ctx itself may already be Done here (e.g. a SIGTERM-driven cancellation
+	// cancels ctx directly, not just runCtx's derived deadline), so this
+	// checkpoint must run on a context that survives that cancellation —
+	// same pattern as the other finalize/cleanup writes in this codebase
+	// (c1file.go, clone_sync.go, syncer.go) — but still bounded, so a wedged
+	// store write during the grace period doesn't just burn the whole window
+	// before getting SIGKILLed anyway.
+	// It would be nice to remove this once we're more confident in the checkpointing logic.
+	cpCtx, cpCancel := context.WithTimeout(context.WithoutCancel(ctx), forcedCheckpointTimeout)
+	defer cpCancel()
+	checkpointErr := s.Checkpoint(cpCtx, true)
+	if checkpointErr != nil {
+		l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
+	}
+	// cause is joined in (not just logged) so a caller with more context —
+	// e.g. pkg/tasks/c1api distinguishing ErrTaskCancelled from a plain
+	// shutdown — can still inspect errors.Is(err, cause) instead of losing it
+	// behind ErrSyncNotComplete.
+	return warnings, errors.Join(checkpointErr, ErrSyncNotComplete, cause)
 }
 
 func tooManyWarnings(warningCount int, completedActionsCount uint64) bool {
--- a/pkg/sync/parallel_syncer.go
+++ b/pkg/sync/parallel_syncer.go
@@ -457,13 +457,14 @@ func (s *syncer) handleOperationError(
 		)
 	}
 	l := ctxzap.Extract(ctx)
-	if errors.Is(cause, context.DeadlineExceeded) {
+	switch {
+	case errors.Is(cause, context.DeadlineExceeded):
 		if s.recordStats {
 			l.Info("sync run duration has expired, exiting sync early", s.syncSummaryFields(trace.SpanFromContext(ctx))...)
 		} else {
 			l.Info("sync run duration has expired, exiting sync early", zap.String("sync_id", s.syncID))
 		}
-	} else if s.recordStats {
+	case s.recordStats:
 		// Same per-step/per-resource-type/retry-wait fields as the deadline
 		// branch above — these are exactly the counters used to reconstruct
 		// a sync's timeline after the fact (see CXH-2121), so a
@@ -471,7 +472,7 @@ func (s *syncer) handleOperationError(
 		// than a deadline-expired one.
 		l.Info("sync context cancelled, exiting sync early",
 			append(s.syncSummaryFields(trace.SpanFromContext(ctx)), zap.NamedError("cancel_cause", cause))...)
-	} else {
+	default:
 		l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(cause))
 	}
 	// ctx itself may already be Done here (e.g. a SIGTERM-driven cancellation
--- a/pkg/sync/parallel_syncer.go
+++ b/pkg/sync/parallel_syncer.go
@@ -176,7 +176,7 @@ func (s *syncer) parallelSync(
 
 		err := s.Checkpoint(ctx, false)
 		if err != nil {
-			return warnings, err
+			return s.handleOperationError(ctx, runCtx, warnings, err)
 		}
 
 		// If we have more than 10 warnings and more than 10% of actions ended in a warning, exit the sync.
@@ -211,7 +211,7 @@ func (s *syncer) parallelSync(
 				s.state.PushAction(ctx, Action{Op: SyncResourceTypesOp})
 				err = s.Checkpoint(ctx, true)
 				if err != nil {
-					return warnings, err
+					return s.handleOperationError(ctx, runCtx, warnings, err)
 				}
 				// Don't do grant expansion or external resources in partial syncs, as we likely lack related resources/entitlements/grants
 				continue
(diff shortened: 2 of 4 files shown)


pkg/sync/parallel_syncer.go
```diff
@@ -166,23 +166,28 @@ func (s *syncer) parallelSync(
...
-				// It would be nice to remove this once we're more confident in the checkpointing logic.
-				checkpointErr := s.Checkpoint(ctx, true)
-				if checkpointErr != nil {
-					l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
-				}
-				return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
-			default:
-				l.Error("sync context cancelled", zap.String("sync_id", s.syncID), zap.Error(err))
-				return warnings, err
+			} else {
```
sergiocorral-conductorone: [low-confidence] **[Major]** This unified branch only runs when the cancellation is observed by the top-of-loop `select`, i.e. between actions. Every action-executing arm of the same `switch` returns the raw error first: `if !s.timedShouldWaitAndRetry(...) { return warnings, err }` (SyncResourceTypesOp, the non-fanned-out SyncResourcesOp, etc.) and `if err != nil { return warnings, err }` after `syncParallel` (SyncResourcesOp fan-out, SyncTargetedResourceOp, and the grants/entitlements arms below). When SIGTERM cancels `ctx` mid-action, the in-flight connector call fails with `context.Canceled` (in-process) or `codes.Canceled` (out-of-process gRPC); `retryer.ShouldWaitAndRetry` returns false for both (pkg/retry/retry.go:61 only retries Unavailable/DeadlineExceeded), so `syncOneAction` yields `workerResult{err}`, `syncParallel` joins it into `batchErr`, and `parallelSync` returns it verbatim -- no forced checkpoint, and `IsSyncPreservable` returns false (status.FromError on a bare context.Canceled gives ok=false -> false; codes.Canceled is not in the allowed switch at syncer.go:104-113). For a 16h sync whose long tail is a fanned-out grants batch, cancellation lands mid-action essentially every time, so the reported CXH-2121 progress loss is not addressed by this hunk. This also means the PR's "known out-of-scope gap" framing is wrong: it is not specific to the legacy C1File engine using ctx in per-action store calls -- the same bypass exists on the Pebble engine via the cancelled connector RPC. (Relatedly, c1zstore/engine.go:16-19 documents EngineSQLite as the default that "Connectors use", so the "Production runs the Pebble engine, where this doesn't apply" mitigation only holds for callers that explicitly pass --storage-engine pebble.) Consider handling cancellation where the action error is inspected -- e.g. a shared helper that, when `runCtx.Err() != nil` (or `errors.Is(err, context.Canceled)` / `status.Code(err) == codes.Canceled` with runCtx done), falls through to this same forced-checkpoint + ErrSyncNotComplete path instead of returning the raw error.

pkg/sync/parallel_syncer.go
```diff
@@ -166,23 +166,28 @@ func (s *syncer) parallelSync(
...
+			// cancels ctx directly, not just runCtx's derived deadline), so this
+			// checkpoint must run on a context that survives that cancellation —
+			// same pattern as the other finalize/cleanup writes in this codebase
+			// (c1file.go, clone_sync.go, syncer.go) that must complete after ctx is done.
+			// It would be nice to remove this once we're more confident in the checkpointing logic.
+			checkpointErr := s.Checkpoint(context.WithoutCancel(ctx), true)
+			if checkpointErr != nil {
+				l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
 			}
+			return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
```
sergiocorral-conductorone: **[Major]** The cancellation cause `err` is now dropped from the returned error on every non-deadline path -- the old `default` branch returned it, and it is only logged (at Info, down from Error). `runCtx`'s cause is not always a benign shutdown: pkg/tasks/c1api/task_helpers.go cancels the sync context with `ErrTaskCancelled` when C1 tells us the task was cancelled server-side (line 142) and with `ErrTaskHeartbeatFailed` after repeated heartbeat failures (line 103). After this change both are converted into `errors.Join(checkpointErr, ErrSyncNotComplete)`, so (a) the real cause never reaches `helpers.FinishTask`, which now reports a server-cancelled task as a resumable incomplete sync, and (b) we do a forced checkpoint write for a sync C1 explicitly cancelled. Suggest preserving the cause while keeping the sentinel, e.g. `return warnings, errors.Join(checkpointErr, fmt.Errorf("%w: %w", ErrSyncNotComplete, err))` (IsSyncPreservable uses errors.Is, so it still classifies as preservable), and keeping `l.Error` for causes that are not `ErrSigTerm`/`context.Canceled` so a genuinely unexpected cancellation is still visible at error level.

pkg/sync/parallel_syncer.go
```diff
@@ -166,23 +166,28 @@ func (s *syncer) parallelSync(
 		select {
 		case <-runCtx.Done():
```
sergiocorral-conductorone: [low-confidence] **[Major]** Ordering issue in the enclosing loop that this hunk doesn't address: the non-forced `err := s.Checkpoint(ctx, false)` at parallel_syncer.go:154 runs on the raw (possibly already cancelled) `ctx` *before* this `select`, and `if err != nil { return warnings, err }` at :156 returns the raw error. `Checkpoint` short-circuits only while `time.Since(s.lastCheckPointTime) < minCheckpointInterval` (10s, syncer.go:406-412); a fanned-out grants/resources batch routinely takes longer than that, so on the first loop iteration after a SIGTERM the store write is attempted on the cancelled ctx, fails with `context.Canceled`, and returns before this new branch is ever reached -- defeating the fix even when the cancellation lands at a clean action boundary. Suggest hoisting the `runCtx.Done()` select above the `s.Checkpoint(ctx, false)` call (and/or giving that call the same `context.WithoutCancel` treatment).

pkg/sync/syncer_ctx_cancel_resume_test.go
```diff
@@ -0,0 +1,101 @@
...
+	c1zPath := filepath.Join(tmpDir, "cancel-resume.c1z")
+	bgCtx := context.Background()
+
+	cancelCtx, cancel := context.WithCancel(bgCtx)
+	seen := 0
+	syncer, err := NewSyncer(cancelCtx, newConnector(bgCtx),
+		WithC1ZPath(c1zPath),
+		WithTmpDir(tmpDir),
+		WithStorageEngine(c1zstore.EnginePebble),
+		WithProgressHandler(func(p *Progress) {
```
sergiocorral-conductorone, line 95: [low-confidence] **[Minor]** The test is meaningful for the loop-boundary path (it does a real two-run resume and asserts a single sealed sync run plus full resource recovery), but its comment overclaims: "a real SIGTERM landing mid-sync rather than at a clean action boundary" is not what it exercises. `mockConnector.ListResources`/`ListResourceTypes` (pkg/sync/syncer_test.go:1771-1780) ignore `ctx` entirely, so after `cancel()` fires inside the progress handler the remaining resource-type actions still complete successfully and the cancellation is only ever observed at the top-of-loop `select`. As written the test cannot regress-guard the mid-action bypass described in the finding on parallel_syncer.go:175, and it would keep passing if that bypass were the only behaviour left. Consider a mock whose ListResources/ListGrants returns `status.FromContextError(ctx.Err())` (or plain `ctx.Err()`) once cancelled -- that variant should currently fail, which is exactly the coverage this PR needs. Minor nit: `seen` is incremented from the syncParallel worker goroutine without synchronisation; that is safe only because `workerCount` defaults to 1 (syncer.go:3426), so it will race under `-race` if anyone adds `WithWorkerCount`.

pkg/sync/parallel_syncer.go
```diff
@@ -166,23 +166,28 @@ func (s *syncer) parallelSync(
...
+				// resumable as a deadline timeout, so it gets the same forced checkpoint and treatment below.
+				l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(err))
+			}
+			// ctx itself may already be Done here (e.g. a SIGTERM-driven cancellation
+			// cancels ctx directly, not just runCtx's derived deadline), so this
+			// checkpoint must run on a context that survives that cancellation —
+			// same pattern as the other finalize/cleanup writes in this codebase
+			// (c1file.go, clone_sync.go, syncer.go) that must complete after ctx is done.
+			// It would be nice to remove this once we're more confident in the checkpointing logic.
+			checkpointErr := s.Checkpoint(context.WithoutCancel(ctx), true)
```
sergiocorral-conductorone: **[Minor]** `context.WithoutCancel(ctx)` strips the deadline as well as the cancellation, so this forced checkpoint write is now completely unbounded -- and it runs precisely during a Kubernetes termination grace period, where a store write that stalls (Pebble flush/compaction, or a slow FS) burns the whole grace window and then gets SIGKILLed anyway, with the same lost-progress outcome the PR is fixing. Note the pre-existing signal goroutine (`for range sigChan { cancel(ErrSigTerm) }`) also means a second SIGTERM cannot force an earlier exit. Suggest bounding it: `cpCtx, cpCancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second); defer cpCancel()`.

pkg/connectorrunner/runner.go
```diff
@@ -138,7 +139,7 @@ func (c *connectorRunner) Run(ctx context.Context) error {
 	}
 
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt)
+	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
```
sergiocorral-conductorone, line 186: [low-confidence] **[Minor]** The signal change itself is correct (`syscall.SIGTERM` is defined on linux/darwin/windows, and `os.Interrupt` == SIGINT, so no duplicate registration), but please sanity-check the end-to-end claim before merging: for hosted connectors the sync runs under pkg/tasks/c1api/full_sync.go, whose HandleTask deletes the partial c1z on *any* sync error -- `os.Remove(c1zPath)` immediately before `FinishTask(ctx, nil, nil, classifySyncError(err))` -- and `classifySyncError` only special-cases ErrIngestInvariantViolated, with hosted task resumption still an open TODO ("We should handle task resumption here", full_sync.go). In that path the freshly forced checkpoint lands in a temp c1z that is then unlinked, and `IsSyncPreservable` has no in-repo consumer (only the C1 backend). Worth stating explicitly in the PR which deployment path actually consumes the checkpoint that CXH-2121 was losing, since if the affected connector runs via c1api this change wouldn't recover the lost hours on its own.

sergiocorral-conductorone, line 186: **[Nit]** PR-description accuracy only, no code change requested: the claim that SIGTERM "was silently ignored until the grace period expired and SIGKILL landed" when the connector is PID 1 isn't how Go behaves. The Go runtime installs a handler for SIGTERM itself, so the kernel's PID-1 default-action protection doesn't apply -- an untrapped SIGTERM terminates a Go process immediately regardless of PID. The observable outcome the PR describes (abrupt exit, in-flight progress lost) is right; only the mechanism is wrong, and it matters because it changes what the pre-fix logs should look like when correlating with the CXH-2121 timeline. Also note the (pre-existing) `signal.Notify` here is never paired with `signal.Stop`.

pkg/sync/parallel_syncer.go
FeliLucero1: Confirmed — traced the retry path and you're right, ShouldWaitAndRetry never retries a canceled call so every one of those ~12 return points bypassed the whole cancellation branch. Fixed in f876f40d: extracted a `finishOnCancellation` helper and added a `runCtx.Err() != nil` guard before each of them. Added `TestSyncResumesAfterMidActionCancellation`, which uses a connector mock that actually respects ctx (the base mock ignores it) — verified it fails on the parent commit and passes on this one. Also corrected the framing in the PR description: this isn't specific to the SQLite engine, it's structural to the retry loop regardless of storage engine.

pkg/sync/parallel_syncer.go
FeliLucero1: Good catch, this was a real regression risk. Fixed in f876f40d: `finishOnCancellation` now returns `errors.Join(checkpointErr, ErrSyncNotComplete, cause)` instead of substituting cause away, so `errors.Is(err, ErrTaskCancelled)`/`errors.Is(err, ErrTaskHeartbeatFailed)` still work for a caller like pkg/tasks/c1api while `errors.Is(err, ErrSyncNotComplete)` still classifies it as preservable. I kept the log at Info uniformly rather than splitting by cause (that would need pkg/sync to know about connectorrunner/c1api-specific sentinels, which risks an import cycle) — the actual severity judgment now lives with whoever calls Sync() and can inspect the specific cause, not baked into the generic engine log line.

pkg/sync/parallel_syncer.go
FeliLucero1: Confirmed. Fixed in f876f40d by moving the `runCtx.Done()` check to the top of the loop iteration, ahead of `Checkpoint(ctx, false)` — cancellation is now caught before that throttled write gets a chance to run on the (possibly already cancelled) raw ctx and fail.

pkg/sync/syncer_ctx_cancel_resume_test.go
FeliLucero1: Both points fixed in f876f40d: renamed to `TestSyncResumesAfterLoopBoundaryCancellation" and corrected the comment to say what it actually covers (loop-boundary only), switched `seen` to `atomic.Int64`, and added `TestSyncResumesAfterMidActionCancellation` with a ctx-aware mock connector for the case this one couldn't reach — verified against the parent commit that it fails without the mid-action fix and passes with it.

pkg/sync/parallel_syncer.go
FeliLucero1: Fixed in f876f40d — added `forcedCheckpointTimeout = 15 * time.Second` and wrapped the detached context: `context.WithTimeout(context.WithoutCancel(ctx), forcedCheckpointTimeout)`. Kept it well under Kubernetes' 30s default grace period rather than reusing dotc1z.FinalizeTimeout (1h default), since that bounds a much heavier WAL+upload tail and would have been effectively unbounded for this purpose.

pkg/connectorrunner/runner.go
FeliLucero1: You're right that I hadn't verified this, and I still can't fully — full_sync.go's HandleTask does delete the temp c1z and finish the task on any error including ErrSyncNotComplete, confirmed by reading it, and task resumption is an open TODO there. Added an explicit caveat in the PR description rather than claiming this fixes CXH-2121 end-to-end: pkg/tasks/local is unaffected either way (any error already failed the command), pkg/tasks/c1api's HandleTask doesn't currently benefit from the resumable classification, and whether the actual production path (which may be a third, closed-source orchestration layer outside this repo) benefits depends on whether it reads IsSyncPreservable/ErrSyncNotComplete and reuses the same store — I don't have visibility into that from here.

FeliLucero1: You're right, and I corrected the PR description — the mechanism I described (kernel suppresses default action for PID 1 with no handler) doesn't apply here since the Go runtime installs its own handler for SIGTERM at process start regardless of signal.Notify. The observable symptom (abrupt exit, in-flight progress lost) is unchanged, just not for the PID-1 reason I originally gave. Left the (pre-existing) signal.Notify/signal.Stop pairing alone since that's out of scope for this fix. Thanks for catching this.

sergiocorral-conductorone, line 186: **[major]** **[Major]** Trapping SIGTERM is the right call, but nothing waits for the work it interrupts, so the checkpoint this PR forces can still be lost to process exit — in exactly the deployment shape the PR targets.

In the long-lived (non-`oneShot`) mode a hosted/containerized connector runs in, every task is dispatched to a goroutine (`go func(t *v1.Task)`, runner.go:279) and nothing tracks it. Once this handler calls `cancel(ErrSigTerm)`:

1. `run()`'s loop select hits `<-ctx.Done()` and returns `c.handleContextCancel(ctx)` (runner.go:221-222), which returns **nil** for `ErrSigTerm` — and if it happens to be blocked in `sem.Acquire(ctx, 1)` that fails and returns via the same helper (:226-230). Either way it returns essentially immediately.
2. `Run()` returns nil, the deferred `r.Close(runCtx)` in pkg/cli/commands.go:467 runs — and `Close()` doesn't wait for task goroutines either; it calls `c.cw.Close()`, tearing down the connector client out from under the in-flight sync.
3. The command returns and the process exits, while the sync goroutine is still inside `handleOperationError`'s `forcedCheckpointTimeout` (15s) write and, after that, the detached c1z finalize bounded by `dotc1z.FinalizeTimeout()` (default **1 hour**, dotc1z/finalize_timeout.go:14).

So the forced checkpoint is written to a store whose finalize may never complete, and there is no `WaitGroup`/semaphore drain and no bounded shutdown window anywhere between the signal and process exit. The `oneShot` CLI path is fine (the task is processed synchronously at :265), which is likely why the new tests — which call `syncer.Sync` directly, never through the runner — don't surface this.

Suggest draining before returning from `run()`, e.g. acquire all slots on a detached, bounded context:

```go
drainCtx, cancelDrain := context.WithTimeout(context.WithoutCancel(ctx), shutdownDrainTimeout)
defer cancelDrain()
if err := sem.Acquire(drainCtx, int64(c.taskConcurrency)); err != nil {
    l.Warn("runner: shutdown drain timed out; in-flight tasks may lose progress", zap.Error(err))
}
```

and not closing the connector wrapper until that drain returns. Whatever the drain budget is, it also needs to be reconcilable with `FinalizeTimeout`'s 1h default, which no realistic grace period can accommodate — worth saying explicitly in the PR which of the two wins.

Happy to be told this is out of scope and belongs in a follow-up, but if so the PR's claim to recover the CXH-2121 hours should be qualified, because for the concurrent runner path the checkpoint's durability is still not guaranteed.

_Verifier (codex): CONFIRM — adds SIGTERM handling but no shutdown drain/join for in-flight task goroutines, so Run() can return and teardown can race/beat checkpoint finalization._

pkg/sync/parallel_syncer.go
```diff
@@ -161,6 +161,19 @@ func (s *syncer) parallelSync(
...
+		// handleOperationError.
+		select {
+		case <-runCtx.Done():
+			return s.handleOperationError(ctx, runCtx, warnings, nil)
+		default:
+		}
+
 		err := s.Checkpoint(ctx, false)
 		if err != nil {
 			return warnings, err
```
sergiocorral-conductorone: **[minor]** **[Minor]** Hoisting the select above this checkpoint narrows the window we flagged but doesn't close it: a cancellation landing between the `default:` at :175 and this write still hits `s.Checkpoint(ctx, false)` on the raw, now-cancelled `ctx`, fails with `context.Canceled`, and returns raw here — no forced checkpoint, and `IsSyncPreservable` returns false (a bare `context.Canceled` gets `ok=false` from `status.FromError`, and `codes.Canceled` isn't in the allowed switch at syncer.go:102-113). Small window, but it is the same failure mode the PR exists to remove, and the fix is a one-liner:

```go
err := s.Checkpoint(ctx, false)
if err != nil {
    return s.handleOperationError(ctx, runCtx, warnings, err)
}
```

`handleOperationError` already returns `batchErr` unchanged when `runCtx` wasn't cancelled, so this costs nothing on the genuine-store-failure path. The same applies to the three forced `s.Checkpoint(ctx, true)` calls in the `InitOp` arm at :212-215, :229-232 and :247-250, which also `return warnings, err` raw — those only run on the first iteration so the exposure is small, but they're the same shape.

_Verifier (codex): CONFIRM — real cancellation window remains between the default: branch and Checkpoint(ctx, false) where a raw context.Canceled can bypass the resumable-cancellation helper._

pkg/sync/parallel_syncer.go
```diff
@@ -422,29 +413,78 @@ func (s *syncer) parallelSync(
...
-	return warnings, errors.Join(checkpointErr, ErrSyncNotComplete)
+	l := ctxzap.Extract(ctx)
+	if errors.Is(cause, context.DeadlineExceeded) {
+		if s.recordStats {
+			l.Info("sync run duration has expired, exiting sync early", s.syncSummaryFields(trace.SpanFromContext(ctx))...)
+		} else {
+			l.Info("sync run duration has expired, exiting sync early", zap.String("sync_id", s.syncID))
+		}
+	} else {
+		l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(cause))
```
sergiocorral-conductorone, line 476: **[minor]** **[Minor]** This branch loses the sync stats summary for exactly the case the PR is about — a regression against pre-PR behaviour, and against the deadline branch three lines up.

Before this change, a non-deadline cancellation returned the raw cause, so `returnSyncError` (syncer.go:669-675) took its normal path and logged `"sync stats so far"` with the full `syncSummaryFields(span)` — per-step durations, per-resource-type counters, retry/rate-limit wall time. Now the returned error satisfies `errors.Is(err, ErrSyncNotComplete)`, so `returnSyncError` short-circuits on its first line and emits nothing, while this `else` logs only `sync_id`. Net effect: a SIGTERM-interrupted sync produces *less* diagnostic output than before, and the missing fields are the same counters used to reconstruct the CXH-2121 timeline in the first place.

Suggest mirroring the deadline branch:

```go
} else if s.recordStats {
    l.Info("sync context cancelled, exiting sync early", append(s.syncSummaryFields(trace.SpanFromContext(ctx)), zap.NamedError("cancel_cause", cause))...)
} else {
    l.Info("sync context cancelled, exiting sync early", zap.String("sync_id", s.syncID), zap.Error(cause))
}
```

Secondary: our earlier suggestion to keep `l.Error` for causes that aren't a benign shutdown wasn't taken — everything is `Info` now. That's much less important given the cause is joined into the returned error, but note the discriminator can't be `ErrSigTerm` (pkg/sync importing pkg/connectorrunner would cycle), so if you want the distinction it has to be something like `errors.Is(cause, context.Canceled) || errors.Is(cause, context.DeadlineExceeded)` → Info, else Error. Also note `s.recordStats` is only true on the Pebble engine (`s.recordStats = s.store.Metadata().Engine == string(c1zstore.EnginePebble)`), so on the default SQLite engine there is no summary either way.

_Verifier (codex): CONFIRM — the new non-deadline branch logs only sync_id/cause, and since the returned error now includes ErrSyncNotComplete, the prior stats-logging path can be skipped._

pkg/sync/parallel_syncer.go
```diff
@@ -422,29 +413,78 @@ func (s *syncer) parallelSync(
...
+	defer cpCancel()
+	checkpointErr := s.Checkpoint(cpCtx, true)
+	if checkpointErr != nil {
+		l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
+	}
+	// cause is joined in (not just logged) so a caller with more context —
+	// e.g. pkg/tasks/c1api distinguishing ErrTaskCancelled from a plain
+	// shutdown — can still inspect errors.Is(err, cause) instead of losing it
+	// behind ErrSyncNotComplete.
+	return warnings, errors.Join(checkpointErr, ErrSyncNotComplete, cause)
```
sergiocorral-conductorone: **[minor low-confidence]** **[Minor]** Now that `cause` is preserved, consider preserving `batchErr` the same way — it is still dropped from the returned error on every cancellation path, and the widened guard makes that drop far more likely than before.

Previously this swallow only happened when the run duration expired. Now *any* cancellation cause (a SIGTERM landing anywhere in a long batch) swallows whatever the in-flight operation actually returned: it's logged at :453-457, but only when it isn't `context.Canceled`/`DeadlineExceeded`, and it never reaches the caller. So a genuine store write failure, a permission error, or a malformed-response error that happens to land in the shutdown window is reported to the caller as a plain resumable incomplete sync, and the next attempt re-runs into it with no record in the returned error chain.

The same argument you (correctly) made for `cause` applies here:

```go
return warnings, errors.Join(checkpointErr, ErrSyncNotComplete, cause, batchErr)
```

`errors.Join` drops nils, `IsSyncPreservable` still returns true via `ErrSyncNotComplete`, and a caller that wants to know *what* failed can now find out. (`ErrIngestInvariantViolated` specifically is safe today — every `invariantVerdict()` site is in the post-`parallelSync` pass in ingest_invariants.go, so it can't arrive as `batchErr` — but that's an accident of where the pass runs, not a property this function guarantees.)

_Verifier (codex): UNCERTAIN — batchErr is indeed dropped from the joined error, but whether that's a bug vs intentional cancellation semantics is unclear from the hunks/contracts alone._

pkg/sync/parallel_syncer.go
```diff
@@ -422,29 +413,78 @@ func (s *syncer) parallelSync(
...
+	// store write during the grace period doesn't just burn the whole window
+	// before getting SIGKILLed anyway.
+	// It would be nice to remove this once we're more confident in the checkpointing logic.
+	cpCtx, cpCancel := context.WithTimeout(context.WithoutCancel(ctx), forcedCheckpointTimeout)
+	defer cpCancel()
+	checkpointErr := s.Checkpoint(cpCtx, true)
+	if checkpointErr != nil {
+		l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
+	}
+	// cause is joined in (not just logged) so a caller with more context —
```
sergiocorral-conductorone: **[nit]** **[Nit]** Joining `cause` also changes what the hosted task manager reports to C1, which is worth a sentence in the PR body since it's a wire-visible contract change with no test.

`pkg/tasks/c1api/manager.go:350-359` does `status.FromError(taskError)`; an `errors.Join` result has no `GRPCStatus()`, so `ok` is false and it falls into the sentinel switch — which now matches, because the joined error satisfies `errors.Is(err, context.Canceled)` (SIGTERM path) or `errors.Is(err, context.DeadlineExceeded)` (run-duration path). So `FinishTask` now reports `codes.Canceled` / `codes.DeadlineExceeded` where it previously reported `codes.Unknown` for the expiry case. That looks strictly more accurate to me, and `NonRetryable` is unaffected, but whether C1's task engine treats `Canceled` as retryable is outside this repo — worth confirming before merge rather than discovering it as a behaviour change in the hosted fleet.

Unrelated naming nit while here: this comment block, the doc comment at :424, and the new test file all refer to `finishOnCancellation`, as does the PR body ("route to `finishOnCancellation`"), but the function is `handleOperationError`. Relatedly, the body's "All ~12 such return points now check `runCtx.Err() != nil` and route to `finishOnCancellation`" over-credits this PR: those call sites already existed on `main` at 8c92491e (added by the Phase 5 merge with deadline-only semantics); this PR adds one (`MarkSyncSupportsDiff`, :394) and generalizes the helper's guard. The generalization is what actually fixes the bypass, so the fix stands — it's the attribution that will confuse anyone doing archaeology later.

_Verifier (codex): CONFIRM — joining cancellation cause plausibly changes downstream status.FromError mapping behavior; the attribution/naming mismatch noted is consistent with the description._

pkg/connectorrunner/runner.go
FeliLucero1: Confirmed, and this is the most important catch on the PR so far — traced the exact race: run() returns almost immediately on ctx.Done() via handleContextCancel, every caller (pkg/cli/commands.go:467) defers Close() right after Run() returns, and nothing tracked the task goroutine in between. Fixed in 7b4b7127: added drainInFlightTasks, called before both cancellation return points in run(), which acquires the full semaphore weight on a detached context.WithoutCancel(ctx) bounded by a new 25s shutdownDrainTimeout — so it waits for every dispatched task goroutine to actually finish (or times out) before run() returns. Also added TestRun_DrainsInFlightTaskBeforeReturningOnCancellation with a task manager whose Process deliberately ignores ctx.Done() (mirroring the detached checkpoint write) — verified it fails without the drain and passes with it, including under -race. Documented in the PR body that this budget targets the checkpoint write specifically, not syncer.Close()'s own separately-bounded (1h default) finalize tail — no realistic grace period covers that, and I didn't want to imply otherwise.

pkg/sync/parallel_syncer.go
FeliLucero1: Confirmed and fixed in 7b4b7127 — took your one-liner suggestion directly: the throttled Checkpoint(ctx, false) now routes through handleOperationError(ctx, runCtx, warnings, err) instead of returning raw, which is a no-op when runCtx wasn't actually cancelled. Also applied the same fix to the three forced Checkpoint(ctx, true) calls in the InitOp arm you flagged as the same shape.

pkg/sync/parallel_syncer.go
FeliLucero1: Confirmed and fixed in 7b4b7127 — the non-deadline branch now mirrors the deadline branch's s.recordStats check and logs the full syncSummaryFields set (plus cancel_cause) when available, instead of just sync_id. On the log-level discriminator: kept it at Info uniformly rather than splitting by cause. I'd actually settled on that earlier in review with the PR author's team for a specific reason -- treating a successfully-checkpointed SIGTERM as Info matches how the deadline branch already treated the equivalent case, and flipping it to Error for anything that isn't literally context.Canceled/DeadlineExceeded would put ErrSigTerm itself in the Error bucket (it's a plain sentinel, doesn't wrap context.Canceled), which is exactly the noisy-false-positive-under-k8s outcome I was trying to avoid. Agreed it's secondary now that cause is preserved in the returned error either way.

pkg/sync/parallel_syncer.go
FeliLucero1: Confirmed and fixed in 7b4b7127 — now errors.Join(checkpointErr, ErrSyncNotComplete, cause, batchErr), same reasoning you applied to cause. Good catch on the widened guard making this much more likely to matter than before.

pkg/sync/parallel_syncer.go
FeliLucero1: Both fixed in 7b4b7127. Added a 'Wire-visible behavior change worth flagging before merge' section to the PR body spelling out the status.FromError/codes.Canceled point for whoever owns the C1 task engine to confirm before merge. And renamed the leftover finishOnCancellation reference in the test file to handleOperationError, plus added an explicit attribution correction to the PR body -- you're right that the call-site wiring is Phase 5's (#1029), this PR generalizes the guard rather than introducing it. Appreciate you catching the over-credit, that would've confused anyone doing archaeology later.

pkg/sync/parallel_syncer.go
```diff
@@ -422,29 +413,89 @@ func (s *syncer) parallelSync(
...
+		l.Error("error checkpointing before exiting sync", zap.Error(checkpointErr))
+	}
+	// cause and batchErr are joined in (not just logged) so a caller with
+	// more context — e.g. pkg/tasks/c1api distinguishing ErrTaskCancelled
+	// from a plain shutdown, or a retry loop that wants to know a genuine
+	// store/connector failure happened to land in the same window — can
+	// still inspect errors.Is(err, ...) instead of it being lost behind
+	// ErrSyncNotComplete. errors.Join drops nils, so this is a no-op when
+	// called from the top-of-loop check (batchErr is nil there).
+	return warnings, errors.Join(checkpointErr, ErrSyncNotComplete, cause, batchErr)
```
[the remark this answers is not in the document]
FeliLucero1: Correct, and this exposed an inaccuracy in my own PR-body note on this same topic -- I'd only accounted for `cause` there, written before `batchErr` was joined in to address a separate finding. Fixed the wire-visible-behavior section in the PR body to cover both: when `batchErr` is nil or doesn't implement `GRPCStatus()`, it falls to the sentinel switch same as before; when it does (a gRPC-status error from the interrupted connector call itself), `status.FromError`'s `errors.As` traversal finds it directly and bypasses the switch. Both directions are more accurate than the previous blanket `codes.Unknown`, resumability/`NonRetryable` are unaffected, and it's now grouped with the other wire-visible-change caveat for whoever confirms C1's retry handling before merge. No code change, this is observability-only as you noted.

pkg/sync/parallel_syncer.go
```diff
@@ -422,29 +413,90 @@ func (s *syncer) parallelSync(
...
```
mateoHernandez123, line 499: Code LGTM on the SIGTERM → drain → checkpoint → `ErrSyncNotComplete` path (Temporal preserves the c1z via `IsSyncPreservable`). Two process nits before I'd approve:

1. Mark the open review threads resolved — the drain major and the minors look fixed in head; they're just still open in the UI.
2. Quick ack from whoever owns the Temporal/task-engine side that treating SIGTERM/cancel as preservable incomplete sync (success-ish on the activity) is the intended wire behavior — including when `batchErr` carries a gRPC status through `finishTask`.

No further code asks from me unless that ack says otherwise.

mateoHernandez123 (approved):

pkg/sync/parallel_syncer.go
FeliLucero1: Ran a deep-code-review pass to verify this before merging on your ack request. Verdict on your two claims:

1. The SIGTERM -> drain -> checkpoint -> ErrSyncNotComplete path is confirmed end-to-end against the actual code (traced connectorrunner.Run -> drainInFlightTasks -> handleOperationError -> errors.Is(err, ErrSyncNotComplete)), backed by passing tests.
2. "Temporal preserves the c1z via IsSyncPreservable" — confirmed there are zero in-repo callers of IsSyncPreservable, so this is exactly the kind of claim that can't be verified from baton-sdk itself; it depends entirely on the closed-source Temporal/task-engine side. Your ask for an explicit ack from whoever owns that is the right call, not just a process nit.

Bonus finding from the same pass, unrelated to your comment but worth fixing before merge: run() actually has a third exit path from its loop, not just the two ctx-cancellation ones — a dispatched task goroutine can set stopForLoop on a "grpc: the client connection is closing" error, which falls straight to the post-loop return with no drain call at all. With taskConcurrency > 1 another task could still be mid-checkpoint-write when that fires. Fixed in 50a2fb0c by moving the drain into a single defer covering every exit from run() instead of a third per-site call (this was the second time a return path got added without one). Writing the regression test for it also caught a real pre-existing data race on stopForLoop itself (plain bool written by a goroutine, read by the main loop) -- converted to atomic.Bool.

Re: marking the review threads resolved -- I don't have permission to do that from here (GitHub review-thread resolution isn't exposed via the API token I'm using), so that'll need a maintainer/reviewer action.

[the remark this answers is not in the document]
FeliLucero1: Already covered — this is the same point as an earlier pass of this same bot (comment further up this file, before the stopForLoop fix), which I addressed by expanding the PR body's "Wire-visible behavior change worth flagging before merge" section to cover both `cause` and `batchErr` explicitly, including the exact codes.DeadlineExceeded/GRPCStatus() mechanics you're describing here. No code change needed — as you say, IsSyncPreservable/NonRetryable are unaffected, and it's flagged in the PR body as something to confirm with whoever owns C1's task engine before merge.

Bencheng21:
This isn't necessarily caused by a deployment. The issue can also occur if the worker stops processing the activity due to an activity heartbeat timeout, etc

I have a small pr https://github.com/ductone/c1/pull/22446 . 

I think it should fix the issue. if it doesn't fix the issue, we can circle back to this pr

PR #1043, opened 2026-07-28
Support additional resource traits in External Identity Matcher (CE-975)

Bencheng21:
## Summary

Fixes [CE-975](https://linear.app/ductone/issue/CE-975/add-support-for-additional-resource-traits-in-baton-id-external): the External Identity Matcher only ever synced and matched `TRAIT_USER`/`TRAIT_GROUP` principals from an external resource c1z. This blocked `baton-azure` from matching Azure service-principal role assignments against identities synced by `baton-microsoft-entra` as `TRAIT_APP`, forcing a connector-specific workaround (baton-azure PR #52) that duplicates identities across connectors instead of matching against the identity connector's own sync.

- Added `sync.WithExternalResourceTraits(...v2.ResourceType_Trait)`: lets a connector configure which traits (e.g. `TRAIT_APP`) the matcher's external-resource sync, principal pool, `ExternalResourceMatchAll`, and key/val `ExternalResourceMatch` logic use.
- `TRAIT_USER`/`TRAIT_GROUP` remain the always-on default when the option isn't set — no behavior change for existing connectors. Passing any traits **replaces** the default set entirely, so a connector that wants USER/GROUP alongside a new trait must list all three.
- Threaded the option through the same plumbing as the existing `--external-resource-c1z`/`--external-resource-entitlement-id-filter` flags: `connectorrunner` → `pkg/tasks/local` (one-shot) and `pkg/tasks/c1api` (platform daemon mode), plus a new `--external-resource-traits` CLI flag (accepts bare names like `app` or canonical `TRAIT_APP`, defaults to `user,group`).
- Extracted a shared `matchProfileAndExpand` helper so any additional trait gets the same generic profile-match + expandable-remap behavior `TRAIT_GROUP` already had, instead of duplicating that logic per trait.

## Test plan

- [x] `go build ./...`
- [x] `go vet` + `golangci-lint run` clean on all touched packages
- [x] `go test ./pkg/sync/...` — full suite passes, including new `TestExternalResourceMatchAllAppTrait` (proves `TRAIT_APP` matching works when opted in) and `TestExternalResourceAppTraitNotMatchedByDefault` (proves it stays inert without opting in)
- [x] Existing `TestExternalResource*` tests continue to pass without an explicit `WithExternalResourceTraits(USER, GROUP)` opt-in — exercises the restored default
- [x] `go test ./pkg/connectorrunner/... ./pkg/tasks/... ./pkg/field/... ./pkg/cli/...` — all pass, no regressions from the plumbing changes

### DIFF
--- a/pkg/sync/syncer.go
+++ b/pkg/sync/syncer.go
@@ -145,11 +145,11 @@ type syncer struct {
 	c1zPath                             string
 	externalResourceC1ZPath             string
 	externalResourceEntitlementIdFilter string
-	// externalResourceTraits are additional resource type traits, beyond the
-	// always-eligible TRAIT_USER/TRAIT_GROUP, that this connector wants
-	// synced from the external resource source and made available to the
-	// External Identity Matcher (see externalMatchTraits, set via
-	// WithExternalResourceTraits).
+	// externalResourceTraits are the resource type traits that this
+	// connector wants synced from the external resource source and made
+	// available to the External Identity Matcher (see externalMatchTraits,
+	// set via WithExternalResourceTraits). There is no implicit default:
+	// TRAIT_USER/TRAIT_GROUP must be included explicitly to be matched.
 	externalResourceTraits      []v2.ResourceType_Trait
 	previousSyncC1ZPath         string
 	previousSyncC1ZPathOptional bool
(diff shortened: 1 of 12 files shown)


pkg/sync/syncer.go
Bencheng21, line 2490: why we `continue` here in the first place?

pkg/sync/syncer.go
```diff
@@ -2588,6 +2340,28 @@ func (s *syncer) SyncExternalResources(ctx context.Context, action *Action) erro
...
+var defaultExternalMatchTraits = []v2.ResourceType_Trait{
+	v2.ResourceType_TRAIT_USER,
+	v2.ResourceType_TRAIT_GROUP,
+}
+
+// externalMatchTraits returns the set of resource type traits the External
+// Identity Matcher should sync and match against: the always-on
+// TRAIT_USER/TRAIT_GROUP, plus any traits a connector opted into via
+// WithExternalResourceTraits (e.g. TRAIT_APP).
+func (s *syncer) externalMatchTraits() map[v2.ResourceType_Trait]bool {
```
laurenleach, line 2351: I think we should only default to `TRAIT_USER/TRAIT_GROUP` if nothing is set. If something is set, we should respect that.

laurenleach:
Why did we drop the `--external-resource-traits` cli? It should be customer configurable. See the linear issue description/screenshot of shared identity source config and attached slack thread

pkg/sync/syncer.go
```diff
@@ -2588,6 +2340,20 @@ func (s *syncer) SyncExternalResources(ctx context.Context, action *Action) erro
 	return nil
 }
 
+// externalMatchTraits returns the set of resource type traits the External
+// Identity Matcher should sync and match against. There is no hardcoded
+// default here: TRAIT_USER/TRAIT_GROUP are only eligible if the caller
+// included them via WithExternalResourceTraits (the CLI's
+// --external-resource-traits flag defaults to "user,group", but direct
+// SyncOpt/programmatic callers must pass traits explicitly).
+func (s *syncer) externalMatchTraits() map[v2.ResourceType_Trait]bool {
```
laurenleach, line 2351: We should default to group/user if nothing set to keep existing behavior

laurenleach (approved):

PR #1044, opened 2026-07-28
feat(cli): project served-policy egress into lambda RunTimeOpts

jugonzalez12:
The lambda server is the only place that sees the connector-config response envelope, so it extracts the envelope's egress section and hands it to the connector via a new RunTimeOpts.EgressPolicy (Governed, AllowedHosts, HTTPSOnly). It performs the one cross-field check it is positioned for — the envelope's config_version must be non-empty and equal to the response's — and version-gates the envelope and egress section. Any binding, version, or section failure resolves to a governed policy with no hosts, so an enforcing connector fails closed (deny-all) rather than serving unenforced. Connectors served without an envelope receive a nil EgressPolicy and keep their prior behavior.

Also prefer the server-stamped config_version for the lambda generation handle (falling back to last_updated), so a warm generation's version matches the handle the invoker sends back and does not force a spurious reload.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -24,6 +24,29 @@ type RunTimeOpts struct {
 	TokenSource         oauth2.TokenSource
 	SelectedAuthMethod  string
 	SyncResourceTypeIDs []string
+	// EgressPolicy carries the server-computed egress policy delivered on the
+	// connector-config response, when one is present. It is nil for connectors
+	// served without a policy envelope; the connector decides how (and whether)
+	// to enforce it.
+	EgressPolicy *EgressPolicy
+}
+
+// EgressPolicy is the connector-facing projection of a served-policy envelope's
+// egress section: the fields a connector runtime needs to enforce egress,
+// extracted after the SDK verified the envelope's binding to this response.
+// The envelope's digests, versions, and capability section are not surfaced —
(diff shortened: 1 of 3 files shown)


pkg/cli/cli.go
```diff
@@ -24,6 +24,29 @@ type RunTimeOpts struct {
...
+// egress section: the fields a connector runtime needs to enforce egress,
+// extracted after the SDK verified the envelope's binding to this response.
+// The envelope's digests, versions, and capability section are not surfaced —
+// they are the SDK's contract to verify, not the connector's.
+type EgressPolicy struct {
+	// Governed is true when the response carried a served-policy envelope. A
+	// governed connector enforces its allowlist even when AllowedHosts is empty
+	// (an empty governed allowlist is deny-all), and a malformed or unsupported
+	// envelope resolves to governed-with-no-hosts so enforcement fails closed.
+	Governed bool
```
kans: I'm confused by this.  Why does the API give a policy for an ungoverned connectorr?

jugonzalez12: it's a clumsy attempt to distinguish no-op manifest from 'on, but no allowlist', as an extra 'enforce this' check. we can just remove it, it's not the enabled toggle

kans (approved):

PR #1058, opened 2026-08-04
structrually remove deadlock potential

kans:
## Summary

- Add `open → closing → closed` mutation admission for Pebble stores.
- Make `Close` wait for admitted mutations before deciding whether to save.
- Preserve dirty state and reopen admission after save failures.
- Guard writer, metadata, file, grant, session, index-repair, bulk-import, and compactor mutations.
- Add deterministic race, persistence, retry, partial-error, liveness, and mutation-inventory tests.

## Performance

Admission adds two brief mutex operations per synchronous mutation. Mutations remain concurrent and no lock is held while the underlying write executes.

## Verification

- [x] `go test -race ./pkg/synccompactor/...`
- [x] `go test -tags=baton_lambda_support ./pkg/dotc1z/... ./pkg/synccompactor/...`
- [x] Focused Pebble admission tests under `-race`
- [x] `golangci-lint run --timeout=3m ./pkg/dotc1z/... ./pkg/synccompactor/...`
- [ ] `go test -race ./pkg/dotc1z/...` — Pebble packages passed; the command failed on the existing SQLite `TestC1ZConcurrentClose` race and later timed out.

## Remaining risks

- External callers can still bypass store admission by directly mutating an engine obtained through `AsEngine`; production mutation paths now use guarded callbacks.
- Racing `CloseEngineOnly` against full `Close` retains its existing unsupported lifecycle semantics.
- Source-cache mutation APIs are not present on the target `origin/main`, so this change does not introduce or depend on them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ dist/
 # specific tenants and their data shapes, which must not enter this
 # public repository (see CLAUDE.md content guidelines).
 docs/tasks/
+.raceout/
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,12 @@ protofmt: ## Format protobuf definitions.
 	buf format -w
 
 .PHONY: test
-test: ## Run the Go test suite used by CI.
-	go test -tags=baton_lambda_support -v ./...
+test: ## Run the ordinary Go test suite used by pull-request CI.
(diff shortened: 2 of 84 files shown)


kans:
I'll revisit this later.

PR #1059, opened 2026-08-04
[CXP-498] Make external principal fold key EqualFold-exact

johnallers:
Stacked on #1046 — base is `jallers/cxp-498-external-principal-index`, so the diff here is just this change.

## Problem

`matchProfile` has no confirmation pass: bucket membership *is* the match decision. So `foldKey` has to reproduce the `strings.EqualFold` comparison the linear scan used, exactly. `strings.ToLower` is a context-free lowercase mapping rather than case folding, and it disagrees in both directions:

| pair | `EqualFold` (the scan) | `ToLower` (the index) |
|---|---|---|
| `Σ` / `ς` | true | **false** — grant silently dropped |
| `ΣΤΕΦΑΝΟΣ` / `στεφανος` | true | **false** — grant silently dropped |
| `ſ` / `s` | true | **false** — grant silently dropped |
| `İSTANBUL` / `istanbul` | **false** | true — grant manufactured |
| `İ` / `i` | **false** | true — grant manufactured |
| `ΜΆΙΟΣ`/`μάιοσ`, `K`(U+212A)/`k`, `ẞ`/`ß`, `ß`/`ss`, `Ö`, Cyrillic, CJK | agree | agree |

The dotted-capital-I direction matters most: the index granted a principal the scan would have rejected, which contradicts #1046's "bucketing is a prefilter only, so it can never manufacture a match the old scan wouldn't have made."

A confirmation pass can't rescue the false-negative cases — the candidate is filtered out before confirmation would run. The fold key itself has to be exact.

## Fix

`strings.EqualFold` reports two strings equal exactly when they have the same rune count and each rune pair shares a `unicode.SimpleFold` orbit. `foldKey` now canonicalizes every rune to the minimum of its own orbit, with an ASCII fast path. That reproduces EqualFold's equivalence classes exactly, which is what makes the absent confirmation pass sound rather than a silent narrowing.

## Tests

- `TestExternalPrincipalIndexNonASCIIMatching` now carries an expectation per case and pins it to `strings.EqualFold`, so a case added later can't quietly encode a divergence. Its `turkish dotted capital i` case asserted a match; it now asserts none.
- `TestFoldKeyAgreesWithEqualFold` covers the equivalence directly, including `ı`, `ẞ`/`ß`, `ß`/`ss`, the Kelvin sign, long s, and differing-length pairs.
- **Repairs `external_principal_index_verification_test.go`.** It has not compiled since this branch deleted `userTraitContainsEmail` (`go vet -tags verification ./pkg/sync/` failed on it). Its reference model now keeps a local copy of that helper rather than sharing code with the implementation it checks.
- The randomized parity generator reaches past ASCII. `ToLower` and `EqualFold` agree on every ASCII pair, so `…RandomizedASCIIParity` structurally could not observe any of these divergences; renamed to `…RandomizedParity`.
- The Unicode parity case became a table covering both directions.
- Adds a `verification-check` make target and wires it into `test-extra`. No target passed `-tags=verification`, which is how the file rotted unnoticed. Also clears the two lint findings that tag surfaces (a pre-existing `gosec` on the fixed test seed, and a dead `externalMatchGrantIDs` helper).

Confirmed all four instruments fail against the previous `ToLower` `foldKey`, then pass with the fix:

```
--- FAIL: TestVerificationExternalPrincipalIndexRandomizedParity
--- FAIL: TestVerificationExternalPrincipalIndexUnicodeEqualFoldParity/greek_final_sigma
--- FAIL: TestVerificationExternalPrincipalIndexUnicodeEqualFoldParity/turkish_dotted_capital_i
--- FAIL: TestExternalPrincipalIndexNonASCIIMatching/{greek_final_sigma,greek_final_sigma_single_rune,turkish_dotted_capital_i}
--- FAIL: TestFoldKeyAgreesWithEqualFold/{ΣΤΕΦΑΝΟΣ/στεφανος,Σ/ς,σ/ς,İSTANBUL/istanbul,İ/i,ſ/s}
```

`go test ./pkg/sync/` passes, `make verification-check` passes, `make lint` and `golangci-lint run --build-tags=verification ./pkg/sync/...` both report 0 issues.

## Note on cost

Orbit-walking is more work per rune than `ToLower`, but only off the ASCII fast path, and `foldKey` runs once per indexed value plus once per lookup — not per candidate. It does not reintroduce anything proportional to grants × principals.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,10 @@ fuzz-smoke: ## Run each native Go fuzzer for FUZZ_TIME (default 30s).
 	go test -run '^$$' -fuzz '^FuzzCondenseFWBW_Cancellation$$' -fuzztime=$(FUZZ_TIME) ./pkg/sync/expand/scc
 	go test -run '^$$' -fuzz '^FuzzCondenseFWBW_FromBytes$$' -fuzztime=$(FUZZ_TIME) ./pkg/sync/expand/scc
 
+.PHONY: verification-check
+verification-check: ## Run reference-model parity instruments behind the verification tag.
+	go test -tags=verification -count=1 -timeout=30m -run '^TestVerification' ./...
+
 .PHONY: differential-check
 differential-check: ## Differential-fuzz SQLite and Pebble for DIFFERENTIAL_TIME.
 	BATON_EXPAND_FUZZ_DURATION=$(DIFFERENTIAL_TIME) go test -v -count=1 -timeout=30m -run '^TestFullPipelineDifferentialFuzz$$' ./pkg/sync/expand
@@ -125,7 +129,7 @@ chaos-soak: ## Run extended seeded chaos connector fanout schedules.
 
 # race-check already includes chaos-full-check's complete deterministic corpus.
 .PHONY: test-extra
-test-extra: race-check compat-check interrupt-check fuzz-smoke differential-check bench-smoke ## Run bounded confidence checks omitted from CI.
+test-extra: race-check compat-check interrupt-check verification-check fuzz-smoke differential-check bench-smoke ## Run bounded confidence checks omitted from CI.
 
 .PHONY: test-nightly
 test-nightly: ## Run extended confidence, fuzz, scheduler, and errorfs checks.
--- a/pkg/sync/external_principal_index.go
+++ b/pkg/sync/external_principal_index.go
@@ -2,6 +2,8 @@ package sync //nolint:revive,nolintlint // matches the existing package name
 
 import (
 	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/types/known/structpb"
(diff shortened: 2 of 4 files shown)


johnallers:
Superseded as a review item by #1060, which puts the folding contract on `main` as tests only. Marking this draft rather than closing it: the `foldKey` change here is one candidate reconciliation for the failures #1060 will surface on this branch, not a decision.

For reference, what's in this branch beyond the tests:

- `foldKey` canonicalizes each rune to the minimum of its `unicode.SimpleFold` orbit (ASCII fast path). `EqualFold` reports two strings equal exactly when they have the same rune count and each rune pair shares a `SimpleFold` orbit, so orbit-min reproduces its equivalence classes exactly.
- A confirmation pass behind `ToLower` bucketing would not be sufficient on its own: `Σ`/`ς` and `ſ`/`s` are false *negatives*, filtered out before confirmation would run.
- `external_principal_index_verification_test.go` needs repair on this branch regardless — it stopped compiling when this branch deleted `userTraitContainsEmail`. There's also no Makefile target passing `-tags=verification`, which is why that went unnoticed.

PR #1060, opened 2026-08-05
[CXP-498] Pin external-resource match folding as a behavioral contract

johnallers:
Tests only — no production change. Passes on `main`.

## Why

External-resource key/value matching compares with `strings.EqualFold`, and nothing asserted that. The existing `TestExternalResource*` tests each end with exactly one matching principal and an ASCII value, so they structurally cannot observe *which* case-insensitive relation is in use — every plausible one agrees on ASCII.

That makes the comparison an unpinned contract, which is exactly the kind of thing a matching rewrite can change without any test noticing.

`EqualFold` is Unicode simple case folding, not lowercasing both sides. The two disagree in both directions, and each direction is a distinct access-correctness failure:

| stored / queried | `EqualFold` | lowercase both | if implemented as lowercasing |
|---|---|---|---|
| `Σ` / `ς` | match | differ | grant **dropped** — principal loses access |
| `ΣΤΕΦΑΝΟΣ` / `στεφανος` | match | differ | grant **dropped** |
| `ſ` / `s` | match | differ | grant **dropped** |
| `İ` / `i` | no match | equal | grant **manufactured** — principal gains access |
| `İSTANBUL` / `istanbul` | no match | equal | grant **manufactured** |
| `ı`/`i`, `ß`/`ss`, `ẞ`/`ß`, `Ö`, Cyrillic, CJK, ASCII | agree | agree | fine |

## Two layers

**The exhaustive table, at the comparison seam** — `pkg/sync/external_match_folding_test.go`. 14 pairs across both routes a user can be matched by (a profile field, a user-trait email), driven through `processGrantsWithExternalPrincipals`. Deliberately at the seam rather than at whatever internal helper implements the comparison, so it survives a rewrite and reports on it. Every case states its expectation and then pins it to `strings.EqualFold`, so a case added later can't quietly encode a divergence instead of reporting one. 28 subtests, 5.7s.

Also pins **one grant per principal** when a carrier can reach that principal by more than one route — possible for the `email` key, which matches a user-trait address as well as a profile field of that name.

**Three representatives, through the chaos corpus** — `internal/chaosconnector/external_principal_corpus.go`. The seam test hands principals in as in-memory objects, so it proves the comparison but not that an awkward value survives reaching it. These cells carry both fold directions the rest of the way: the value round-trips through a sealed external c1z and crosses both transports before it is matched, and the resume-cut and sealing oracles come along free.

| cell | route | stored | match value | expects |
|---|---|---|---|---|
| `fold-profile-final-sigma` | profile `upn` | `ΣΤΕΦΑΝΟΣ` | `στεφανος` | match |
| `fold-profile-dotted-capital-i-miss` | profile `upn` | `İSTANBUL` | `istanbul` | no match |
| `fold-email-final-sigma` | user-trait email | `ΣΤΕΦΑΝΟΣ@example.com` | `στεφανος@example.com` | match |

Profile values travel as `structpb.Struct` and trait emails as `v2.UserTrait_Email`, so each encoding gets a witness. The dotted-`İ` cell also probes whether anything in the store applies NFKC-style normalization, which would decompose the stored value and change the answer.

Only three, not the full table: each corpus cell costs ~1.4s in `TestChaosConnectorExternalPrincipalCorpus` (two transports) plus ~1.6s in `…CorpusResumesAfterRewriteCut`, which iterates the whole corpus. Moving all 28 assertions there would be ~84s against 5.7s at the seam, and would inflate `chaos-full-check` and `race-check` — for a property that is pure string comparison.

### `ExternalWorld`

Cases previously shared one hardcoded external world (`externalPrincipalSourceScenario`), so none could supply its own principal. `ExternalPrincipalCase.ExternalWorld` is an opt-in mutator applied to the per-case copy `Build` already constructs; cells without one are unaffected. It reuses the existing `initialDataset` helper. This also unblocks the multi-match fan-out cell we discussed separately.

Corpus expectations stay declared rather than computed from `strings.EqualFold` — the corpus is a declared-manifest artifact per `docs/CHAOS_CONNECTOR.md`, and the exhaustive table that warrants self-pinning lives at the seam.

## Relationship to #1046

#1046 replaces the linear scan with an index whose bucketing normalizes with `strings.ToLower`. Both layers fail on that branch.

Seam — 10 subtests, the 5 divergent pairs across both routes:

```
--- FAIL: TestExternalResourceMatchProfileFoldingContract/greek_final_sigma
--- FAIL: TestExternalResourceMatchProfileFoldingContract/greek_final_sigma_in_word
--- FAIL: TestExternalResourceMatchProfileFoldingContract/latin_long_s
--- FAIL: TestExternalResourceMatchProfileFoldingContract/turkish_dotted_capital_i
--- FAIL: TestExternalResourceMatchProfileFoldingContract/turkish_dotted_capital_i_in_word
--- FAIL: TestExternalResourceMatchEmailFoldingContract/... (same five)

    profile "Σ" vs match value "ς": expected matched=true   (got false — grant dropped)
    profile "İ" vs match value "i": expected matched=false  (got true  — grant manufactured)
```

Corpus — all three cells, on both transports:

```
--- FAIL: .../fold-profile-final-sigma/{direct,grpc}
      chaos oracle: external principal mismatch: expected [external-fold-user], actual []
--- FAIL: .../fold-profile-dotted-capital-i-miss/{direct,grpc}
      chaos oracle: external principal mismatch: expected [], actual [external-fold-user]
--- FAIL: .../fold-email-final-sigma/{direct,grpc}
      chaos oracle: external principal mismatch: expected [external-fold-user], actual []
```

Landing this first means #1046 gets that signal from CI and can reconcile the normalization deliberately, rather than the difference going unobserved.

## Note

`pkg/sync/external_principal_index_verification_test.go` is `//go:build verification` and does not compile on `main` at all — it references `newExternalUserPrincipalIndex`, `mergePositions`, `testUserPrincipal`, and `testGroupPrincipal`, none of which exist here (they arrive with #1046). No Makefile target passes `-tags=verification`, so it isn't caught. Left alone here since it belongs with #1046; noting it so it isn't mistaken for coverage that currently runs.

## Testing

`go test ./pkg/sync/` passes, `go test ./internal/chaosconnector/...` passes, `make lint` reports 0 issues. Corpus additions measured at ~4s on `TestChaosConnectorExternalPrincipalCorpus` (12.5s → 16.7s) and ~5s on its resume-cut pass (14.7s → 19.7s).

### DIFF
--- a/internal/chaosconnector/external_principal_corpus.go
+++ b/internal/chaosconnector/external_principal_corpus.go
@@ -1,6 +1,8 @@
 package chaosconnector
 
 import (
+	"fmt"
+
 	v2 "github.com/conductorone/baton-sdk/pb/c1/connector/v2"
 	"github.com/conductorone/baton-sdk/pkg/annotations"
 	"github.com/conductorone/baton-sdk/pkg/bid"
@@ -15,6 +17,11 @@ const (
 	ExternalGroupTypeID   = "external-group"
 	InternalGroupTypeID   = "internal-group"
 	ExternalPlaceholderID = "external-placeholder"
+
+	// ExternalFoldUserID is the principal the case-folding cells add to the
+	// external world. It is deliberately separate from the standard principals
+	// so a fold expectation cannot be satisfied by one of them.
+	ExternalFoldUserID = "external-fold-user"
 )
 
 type ExternalMatchKind string
@@ -36,6 +43,11 @@ type ExternalPrincipalCase struct {
 	Value                string
 	ExpectedPrincipalIDs []string
 	Expandable           bool
+
+	// ExternalWorld optionally adds to or edits the external dataset for this
+	// case, for cases whose question cannot be asked of the standard world.
+	// Nil leaves that world untouched, which is the common case.
+	ExternalWorld func(*Dataset) error
 }
 
 func ExternalPrincipalCorpus() []ExternalPrincipalCase {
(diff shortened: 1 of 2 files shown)


kans (approved):
great, thanks

PR #1074, opened 2026-08-10
Poll legacy custom action managers to a terminal status

jugonzalez12:
What this does

registerLegacyAction wraps actions from the deprecated CustomActionManager / RegisterActionManagerLimited interfaces as ordinary action handlers. Until now the wrapper discarded the id and status the legacy manager returned, so an action still running at the legacy manager's own short wait was resolved as complete, with whatever partial response existed — a false success for any slow legacy action.

The wrapper now:

- keeps the old behavior for synchronous managers that never populated an id or status (their invoke response resolves the action, exactly as before);
- for an explicitly in-flight result (PENDING/RUNNING with a usable id), polls the legacy manager's GetActionStatus to a terminal status before resolving, with capped backoff (1s doubling to 30s);
- tolerates a few consecutive status-lookup failures (one flaky remote read must not convert a succeeding action into a failure), carries the last real response, and fails closed on persistent lookup errors, a reported failure, or an unexpected status;
- is bounded by the handler context's deadline, so a status that never converges terminates with the handler budget rather than polling forever.

Why it's here

Review on the inline_wait PR (#1071 and previous drafts) flagged that the legacy path discards the inner status — meaning the requested wait can't be honored for legacy managers, and slow legacy actions false-complete regardless. This change fixes that at the root rather than threading the wait into the deprecated interface: once the wrapped handler blocks to a terminal status, inline_wait composes with legacy managers with no extra plumbing (an action finishing inside the window returns its real terminal status inline; one still running returns an honest RUNNING). It also stands alone: the false-complete existed before inline_wait did.

We can choose to reject this kind of behavior change, but I think it's appropriate for us to decide whether or not to address this issue, since it's raised by bots over and over.

Behavior changes and risk

Perimeter. Only connectors on the deprecated interfaces are affected, and within those, only actions that are still in flight when the legacy manager's own wait expires. Fast legacy actions (the common case — single API calls) return a terminal status from the invoke directly, never enter the poll loop, and behave byte-for-byte as before. Modern actions don't touch this code.

The honest sharp edge: this makes previously-dead code load-bearing. Because the old wrapper discarded the status, a legacy manager's GetActionStatus has never been exercised through this path — bugs in hand-rolled implementations have been invisible. After this change, a persistently failing lookup fails the action (visibly, after three consecutive errors) and a status that never converges occupies a goroutine until the handler deadline, then fails. A visible failure is the correct replacement for a silent false success, but it can read as a regression for a fire-and-forget action that "worked" before. Mitigating this in practice: the known legacy implementations return the SDK's own ActionManager as their CustomActionManager, so the GetActionStatus being promoted to load-bearing is the SDK's own, not custom code. Reviewer ask: if you know of a connector that hand-rolls GetActionStatus, that's the one worth checking against this change.

Observable shift for slow legacy actions. Callers see an honest in-flight status at the wait boundary instead of a fast false COMPLETE, with the real outcome landing when the inner action finishes. This is deployment-order safe: standalone, the platform's current handling of in-flight responses is unchanged; combined with the inline_wait change, slow legacy actions gain real inline outcomes.

Resource bounds. Worst case is one goroutine per slow legacy invoke for the handler budget, and a handful of status calls under the 30s backoff cap — the same order as any slow modern handler.

Testing

Five new tests cover the wrapper: synchronous managers resolve unchanged, in-flight results poll to success and to failure, consecutive lookup errors fail closed at the threshold, and a transient error followed by recovery succeeds. Note the poll intervals are production constants, so these tests contribute several seconds of real wall time to the connectorbuilder suite.

### DIFF
--- a/pkg/connectorbuilder/actions.go
+++ b/pkg/connectorbuilder/actions.go
@@ -242,6 +242,10 @@ func registerLegacyAction(
 	legacyManager CustomActionManager,
 	intervals legacyPollIntervals,
 ) error {
+	// A zero or negative interval would busy-loop the status poll.
+	if intervals.initial <= 0 || intervals.max <= 0 {
+		intervals = defaultLegacyPollIntervals
+	}
 	handler := func(ctx context.Context, args *structpb.Struct) (*structpb.Struct, annotations.Annotations, error) {
 		// The inner call keeps the detached handler context; its one-hour
 		// deadline is the execution backstop for however long the legacy
@@ -261,7 +265,7 @@ func registerLegacyAction(
 		// A settled status at the invoke seam resolves like a terminal
 		// poll. The SDK's own manager reports handler failures in-band as
 		// FAILED with a nil error, so this must not resolve as success.
-		if actionStatus == v2.BatonActionStatus_BATON_ACTION_STATUS_COMPLETE || actionStatus == v2.BatonActionStatus_BATON_ACTION_STATUS_FAILED {
+		if isSettledActionStatus(actionStatus) {
 			return resp, annos, legacyStatusErr(schema.GetName(), actionStatus)
 		}
 
@@ -308,14 +312,14 @@ func registerLegacyAction(
 			}
 			// Keep the last meaningful response for the error exits above;
 			// an indeterminate poll's payload must not replace it.
-			if pollResp != nil && (isInFlightActionStatus(st) || st == v2.BatonActionStatus_BATON_ACTION_STATUS_COMPLETE || st == v2.BatonActionStatus_BATON_ACTION_STATUS_FAILED) {
+			if pollResp != nil && (isInFlightActionStatus(st) || isSettledActionStatus(st)) {
 				resp, annos = pollResp, pollAnnos
 			}
 
 			switch {
 			case isInFlightActionStatus(st):
 				statusErrs = 0
-			case st == v2.BatonActionStatus_BATON_ACTION_STATUS_COMPLETE || st == v2.BatonActionStatus_BATON_ACTION_STATUS_FAILED:
+			case isSettledActionStatus(st):
 				return resp, annos, legacyStatusErr(schema.GetName(), st)
 			default:
 				// An indeterminate status gets the same tolerance as a
@@ -346,6 +350,14 @@ func legacyStatusErr(name string, st v2.BatonActionStatus) error {
 
 func isInFlightActionStatus(s v2.BatonActionStatus) bool {
 	return s == v2.BatonActionStatus_BATON_ACTION_STATUS_PENDING || s == v2.BatonActionStatus_BATON_ACTION_STATUS_RUNNING
+}
+
+// isSettledActionStatus is the single gate deciding which statuses resolve
+// immediately, at the invoke seam and from polls alike. A new terminal enum
+// value must be added here, or it takes the indeterminate path: polled to
+// the tolerance threshold, then failed closed.
+func isSettledActionStatus(s v2.BatonActionStatus) bool {
+	return s == v2.BatonActionStatus_BATON_ACTION_STATUS_COMPLETE || s == v2.BatonActionStatus_BATON_ACTION_STATUS_FAILED
 }
 
 // addActionManager handles deprecated CustomActionManager and RegisterActionManagerLimited interfaces
--- a/pkg/connectorbuilder/actions.go
+++ b/pkg/connectorbuilder/actions.go
@@ -267,7 +267,7 @@ func registerLegacyAction(
 		// poll. The SDK's own manager reports handler failures in-band as
 		// FAILED with a nil error, so this must not resolve as success.
 		if isSettledActionStatus(actionStatus) {
-			return resp, annos, legacyStatusErr(schema.GetName(), actionStatus)
+			return resp, annos, legacyStatusErr(schema.GetName(), actionStatus, resp)
 		}
 
 		// An unresolved claim without an id cannot be polled; resolve with
@@ -303,7 +303,7 @@ func registerLegacyAction(
 			if err != nil {
 				statusErrs++
 				if statusErrs >= maxConsecutiveStatusErrors {
-					return resp, annos, err
+					return resp, annos, fmt.Errorf("legacy action %q status lookup failed: %w", schema.GetName(), err)
 				}
 				l.Warn("legacy action status check failed, retrying",
 					zap.String("action", schema.GetName()),
@@ -321,7 +321,7 @@ func registerLegacyAction(
 			case isInFlightActionStatus(st):
 				statusErrs = 0
 			case isSettledActionStatus(st):
-				return resp, annos, legacyStatusErr(schema.GetName(), st)
+				return resp, annos, legacyStatusErr(schema.GetName(), st, resp)
 			default:
 				// An indeterminate status gets the same tolerance as a
 				// lookup error: transient anomalies recover, persistent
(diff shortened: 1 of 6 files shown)


jugonzalez12:
The review comments here are eerily akin to those of a previous draft (#1063), so the next commit will attempt to skip some of the back-and-forth here.

jugonzalez12:
▎ Two deliberate choices: The poll intervals are unexported package variables rather than injected configuration — they're production constants whose only non-default consumer is this package's sequential tests, and a config parameter would add API surface to a deprecated compatibility path. And an explicitly set but unrecognized status at the invoke seam fails closed rather than passing through: only the zero value means "this manager never populated status," while any other value is a claim the wrapper refuses to misread as success. The same statuses arriving from polls get the three-strike tolerance instead, because polls repeat and transient anomalies can recover.

jugonzalez12:
Status classification, stated once: the wrapper sorts every invoke and poll status into three buckets. UNSPECIFIED means the manager never populated status — the legacy pass-through, resolving with the response as the old wrapper always did. COMPLETE and FAILED are settled and resolve immediately, at the seam and from polls alike — including the in-band shape where the SDK's own manager reports a handler failure as FAILED with a nil error. Everything else is unresolved: polled to a terminal status when an id is present (with a shared three-strike tolerance covering both lookup errors and indeterminate answers), passed through when there's no id to poll. One consequence to know about enum evolution: if a new terminal status is ever added to BatonActionStatus, this wrapper will treat it as unresolved — poll it to the tolerance threshold, then fail closed — until the value is added to the settled-status checks. That's safe (it can never be misread as success) but deliberate: the settled set is enumerated, not inferred. Poll pacing is captured at registration, so the detached poll goroutine reads no shared state; tests drive the loop in milliseconds through the same parameter.

pkg/connectorbuilder/actions.go
```diff
@@ -198,14 +224,132 @@ func (b *builder) GetActionStatus(ctx context.Context, request *v2.GetActionStat
...
+				}
+				l.Warn("legacy action status check failed, retrying",
+					zap.String("action", schema.GetName()),
+					zap.Int("consecutive_anomalies", statusErrs),
+					zap.Error(err))
+				continue
+			}
+			// Keep the last meaningful response for the error exits above;
+			// an indeterminate poll's payload must not replace it.
+			if pollResp != nil && (isInFlightActionStatus(st) || isSettledActionStatus(st)) {
```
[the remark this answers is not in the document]
jugonzalez12: ---
Deliberate: retention is pair-atomic because annotations describe the response they arrived with. Updating annos on its own condition would attach one poll's annotations to a different poll's (or the invoke's) payload — a coherent-looking but mismatched pair, which is worse than keeping the last coherent pair intact. A settled poll with annotations but no body is not a shape any known manager produces; if one appears, the right fix is retaining that poll's (nil, annos) pair atomically, not splitting the guard.

---

pkg/connectorbuilder/actions.go
```diff
@@ -198,14 +224,139 @@ func (b *builder) GetActionStatus(ctx context.Context, request *v2.GetActionStat
...
+	return s == v2.BatonActionStatus_BATON_ACTION_STATUS_PENDING || s == v2.BatonActionStatus_BATON_ACTION_STATUS_RUNNING
+}
+
+// isSettledActionStatus is the single gate deciding which statuses resolve
+// immediately, at the invoke seam and from polls alike. A new terminal enum
+// value must be added here, or it takes the indeterminate path: polled to
+// the tolerance threshold, then failed closed.
+func isSettledActionStatus(s v2.BatonActionStatus) bool {
+	return s == v2.BatonActionStatus_BATON_ACTION_STATUS_COMPLETE || s == v2.BatonActionStatus_BATON_ACTION_STATUS_FAILED
+}
```
ggreer, line 369: setStatusLocked() and OutstandingAction's evictable() could use this.

pkg/connectorbuilder/actions.go
```diff
@@ -198,14 +224,139 @@ func (b *builder) GetActionStatus(ctx context.Context, request *v2.GetActionStat
...
+	if st == v2.BatonActionStatus_BATON_ACTION_STATUS_COMPLETE {
+		return nil
+	}
+	if inner := resp.GetFields()["error"].GetStringValue(); inner != "" {
+		return fmt.Errorf("legacy action %q failed: %s", name, inner)
+	}
+	return fmt.Errorf("legacy action %q failed", name)
+}
+
+func isInFlightActionStatus(s v2.BatonActionStatus) bool {
```
ggreer, line 359: I think setStatusLocked() could use this instead of checking the status directly. Same with localActionInvoker's Process() method.

pkg/connectorbuilder/actions.go
```diff
@@ -198,14 +224,139 @@ func (b *builder) GetActionStatus(ctx context.Context, request *v2.GetActionStat
 }
 
 // registerLegacyAction wraps a legacy CustomActionManager action as an ActionHandler and registers it.
-func registerLegacyAction(ctx context.Context, registry actions.ActionRegistry, schema *v2.BatonActionSchema, legacyManager CustomActionManager) error {
+func registerLegacyAction(
+	ctx context.Context,
+	registry actions.ActionRegistry,
+	schema *v2.BatonActionSchema,
+	legacyManager CustomActionManager,
+	intervals legacyPollIntervals,
```
ggreer, line 243: It would be nice if we didn't have to pass in intervals (since in real code it's always the defaults), but I know that makes testing this harder, so I guess this is the best option available.

ggreer (approved):
I noticed a couple of tiny things where some new functions could be used in existing code, but LGTM.

jugonzalez12:
The robo-review is not converging (basically reversing its own position), my agent's thoughts:

1. Poll pacing vs. the default inline wait — decline, with math the robot missed. The premise is right: first poll at 1s, default wait 1s, so a polled legacy action can't settle inline under the default wait. But the proposed fix (sub-second initial interval) doesn't deliver its stated benefit for the population that matters: the known legacy managers are SDK-backed, and their inner InvokeAction blocks for its own one-second wait before the wrapper even has an id to poll — the outer window is exhausted before polling begins, at any pacing. And under real platform windows (~4 minutes), the 1s/3s/7s ladder delivers terminal statuses inline with seconds to spare. The fix only helps a hypothetical third-party manager that returns in-flight instantly and settles in under a second, at the cost of denser early polling against exactly the rate-limited remote managers the backoff protects. If the human reviewer wants it anyway it's a one-constant change, but it shouldn't ride as a robot-round fix.

2. Nil poll annotations displacing invoke annotations — decline; it's the mirror image of the finding we already declined. Last time the robot wanted annotations retained independently when the payload was nil; now it wants invoke-annotations retained when the settling payload's annotations are nil. Both fail against the same principle already posted on the PR: annotations describe the response they arrived with. A settling response with nil annotations means the final response has none — resurrecting invoke-time annotations pairs stale metadata with a payload it never described. The pre-change behavior it cites isn't a contract to preserve: pre-change there was no poll at all, so "always propagated invoke annotations" is just "returned the only pair that existed."

3. The escape hatch — decline (a) and (b) firmly, adopt (c) as one sentence of PR text. Option (b) — resolve persistent indeterminate with the last known response — is the one to reject hardest: it reintroduces the false-COMPLETE this PR exists to kill, for exactly the managers whose status reporting can't be trusted. Option (a) adds permanent exported API to a deprecated path, serving a self-contradictory implementation shape: to be harmed, a manager must report in-flight-with-id from InvokeAction while having a broken GetActionStatus — claiming trackability while being untrackable — and the honest outcome for that shape is a visible failure. Option (c)'s version bump rides release commits in this repo, not feature PRs, but the migration-note half is fair and nearly free. Add to the PR body: "Behavior change for deprecated-interface connectors: a legacy manager that reports an in-flight status with an id but cannot answer GetActionStatus will now fail those actions visibly (~7s at default pacing) instead of resolving them as successes at one second. Known implementations delegate to the SDK's own manager and are unaffected."

jugonzalez12:
> I noticed a couple of tiny things where some new functions could be used in existing code, but LGTM.

addressing these in #1080

PR #1076, opened 2026-08-11
lambda: recover panics in the connector interceptor chain

kans:
## What

Adds panic recovery to the lambda transport's interceptor chain, which had auth only while the normal gRPC path has had `grpc_recovery` all along.

## Why

A panic in a connector method escaped into the lambda runtime, which reports a crashed invocation carrying no gRPC status. The caller cannot classify the failure, and the only record is the raw runtime log. With recovery, the panic returns as `codes.Internal` with the panic value and stack logged in-process.

Note this does not yet change checkpoint behavior on the caller side: a statusless invocation failure and `codes.Internal` are both treated as non-preservable today, so a panic still discards the sync artifact either way. The preservation benefit arrives when the caller's preservability rules accept `Internal`. This PR's value is diagnosis — a stack trace and a real status instead of an opaque `Unhandled`.

## How

- Export the recovery interceptor as `ugrpc.RecoveryUnaryInterceptor()` and use it in both server paths, so there is one construction rather than two.
- Extract `lambdaUnaryInterceptorChain(...)`, placing recovery first and therefore outermost, so panics in auth as well as in the connector method are caught.
- Chain construction moved into a named function specifically so a test can assert recovery is present, rather than trusting the call site not to drop it again.

## Test plan

- [x] `go test ./pkg/ugrpc/...` — panic becomes `Internal`; panic detail logged; successful responses, coded errors, and bare errors pass through unchanged
- [x] `go test -tags baton_lambda_support ./pkg/cli/...` — panic in handler and in a wrapped interceptor both recover; wrapped interceptors still run
- [x] Validated the lambda test fails when recovery is removed from the chain (panic escapes, test binary crashes)
- [x] `go build ./...` with and without `baton_lambda_support`; `make lint` reports 0 issues

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,8 +39,11 @@ jobs:
         run: go test -tags=baton_lambda_support -v -covermode=count -json ./... > test.json
       - name: go tests
         if: ${{ matrix.platform == 'windows-latest' }}
-        # Run tests with -short on Windows since its filesystem is very slow, causing CI to time out.
-        run: go test -timeout=30m -tags=baton_lambda_support -short -v -covermode=count -json ./... > test.json
+        # -short because the Windows filesystem is very slow and the job times
+        # out. No lambda tag for the same reason: lambda connectors only ship
+        # on Amazon Linux, so the tagged tests have no Windows deployment
+        # target and only spend the slowest runner's time budget.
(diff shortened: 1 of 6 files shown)


ggreer:
We should skip the lambda tests on windows. Then this can be merged.

ggreer (approved):

PR #1083, opened 2026-08-12
fix(uhttp): make CreateCacheKey's header set opt-in

Bencheng21:
## Summary
- `CreateCacheKey` only folded `Accept`, `Content-Type`, `Cookie`, and `Range` into the HTTP response cache key. Any other request header was silently ignored, so two GET requests differing only in an unlisted header (e.g. `Authorization`, a custom API-version or tenant header) collided on the same cache entry and one request could be served the other's cached response.
- `CreateCacheKey` keeps that original default set unchanged. A new variadic parameter, `extraCacheKeyHeaders`, lets a caller fold additional headers into the key on top of it: `CreateCacheKey(req, "Authorization")`. This threads through `icache.Get`/`Set` and its three implementers (`GoCache`, `DBCache`, `NoopCache`) — all additive, so nothing existing breaks.
- Connector-facing entry point: `BaseHttpClient.Do` now takes `CallOption` instead of `DoOption`. `DoOption` gained an `applyDo` method so every existing `WithJSONResponse`/`WithErrorResponse`/etc. call site keeps compiling unchanged (Go wraps the returned `DoOption` into the `CallOption` interface automatically). `WithCacheKeyHeaders(headers...)` is the new option:
  ```go
  resp, err := client.Do(req, uhttp.WithCacheKeyHeaders("Authorization"), uhttp.WithJSONResponse(&result))
  ```
- Along the way, fixed a related bug in `BaseHttpClient.Do`: it computes the cache key from `req` before the round trip (`Get`) and again from the same `req` after it (`Set`). `http.Client.Do` forks the `*http.Request` struct on every call once `Timeout > 0` (which `uhttp.NewClient` always sets), but that fork is shallow, so `Header` stays the same map the caller passed in — a `RoundTripper` further down the chain (e.g. `userAgentTripper`, which sets `User-Agent`) mutates that shared map, so if a caller opts a header into the key that some `RoundTripper` also touches, the `Set` key could differ from every future `Get` key and the cache would become write-only for that request shape. Fixed by round-tripping on `req.Clone(req.Context())` instead of `req` itself.

## Test plan
- [x] `pkg/uhttp/client_test.go`: default set unaffected by headers outside it; `extraCacheKeyHeaders` opts a header in (case-insensitively) without widening the key to everything else.
- [x] `pkg/uhttp/wrapper_test.go`: `TestWrapper_Do_CachesDespiteRoundTripHeaderInjection` opts a header into the key via `WithCacheKeyHeaders` and injects it via a `RoundTripper` between `Get` and `Set` — confirmed it fails without the request-clone fix and passes with it.
- [x] `go test ./pkg/uhttp/...` and `go test ./...` (whole module)
- [x] `golangci-lint run ./pkg/uhttp/...`
- [x] `go build ./...`

### DIFF
--- a/pkg/uhttp/client.go
+++ b/pkg/uhttp/client.go
@@ -117,13 +117,16 @@ var defaultCacheKeyHeaders = map[string]struct{}{
 	"Range":        {},
 }
 
+// CreateCacheKey generates a cache key based on the request URL, query parameters, and headers.
 func CreateCacheKey(req *http.Request, extraCacheKeyHeaders ...string) (string, error) {
 	if req == nil {
 		return "", fmt.Errorf("request is nil")
 	}
 
 	var sortedParams []string
+	// Normalize the URL path
 	path := strings.ToLower(req.URL.Path)
+	// Combine the path with sorted query parameters
 	queryParams := req.URL.Query()
 	for k, v := range queryParams {
 		for _, value := range v {
@@ -133,6 +136,7 @@ func CreateCacheKey(req *http.Request, extraCacheKeyHeaders ...string) (string, 
 
 	sort.Strings(sortedParams)
 	queryString := strings.Join(sortedParams, "&")
+	// Include relevant headers in the cache key
 	var headerParts []string
 	for key, values := range req.Header {
 		_, ok := defaultCacheKeyHeaders[key]
@@ -149,8 +153,10 @@ func CreateCacheKey(req *http.Request, extraCacheKeyHeaders ...string) (string, 
 
 	sort.Strings(headerParts)
 	headersString := strings.Join(headerParts, "&")
+	// Create a unique string for the cache key
 	cacheString := fmt.Sprintf("%s?%s&headers=%s", path, queryString, headersString)
 
+	// Hash the cache string to create a key
 	hash := sha256.New()
 	_, err := hash.Write([]byte(cacheString))
 	if err != nil {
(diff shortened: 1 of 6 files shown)


Bencheng21:
Follow-up fix pushed: including every header exposed a latent bug in `BaseHttpClient.Do` itself.

`Do` computes the cache key from `req` before the round trip (`Get`) and again from the same `req` after it (`Set`). `http.Client.Do` forks the `*http.Request` struct on every call once `Timeout > 0` (which `uhttp.NewClient` always sets) — but that fork only copies the struct, not the `Header` map, so `Header` stays the exact map the caller passed in. `userAgentTripper` then does `req.Header.Set("User-Agent", ...)` on that shared map, mutating the request `Do()` is holding.

Previously this was invisible because `User-Agent` wasn't in the old 4-header allowlist either way. Once the key includes every header, that mutation changes the `Set` key on any request without an explicit `User-Agent` — i.e. nearly all of them — so no future identical request can ever match what was stored: the cache became write-only.

Fix: round-trip on `req.Clone(req.Context())` instead of `req` itself, so whatever the transport chain injects lands on the outbound copy, not on the request `CreateCacheKey` hashes for `Set`. Added `TestWrapper_Do_CachesAcrossTransportHeaderInjection`, confirmed it fails (2 server hits instead of 1) without the fix and passes with it.

laurenleach (changes requested):
I dont think we should have every header. this should be an option that you can pass in from the connector if you want additional headers in the cache key

PR #1086, opened 2026-08-13
Fix the C1File close-vs-write data race

kans:
Closing a C1File while a write is in flight raced on the database handle: closeRawDB nil'd c.db while validateDb read it on another goroutine, so a write could slip past the open-check onto a handle being torn down. TestC1ZConcurrentClose reproduces this on main today — it fails there and passes here.

Closed-ness is now an atomic flag rather than a nil'd pointer: closeRawDB CAS-flips dbClosed before closing rawDb, validateDb loads it, and the pointers stop being written outside construction. The rawDBOpen predicate replaces the bare rawDb != nil checks in Close, finalize, closeWithoutSave, and copyIsolateSync, which had quietly changed meaning once a closed store could still hold a non-nil handle — finalize consulting the stale check could conclude the handle was open and delete the database file out from under a save.

Extracted from the store-teardown branch so the live bug fix is not queued behind that review; the four files here are byte-identical to their versions there, and that branch rebases over this cleanly.

### DIFF
--- a/pkg/dotc1z/c1file.go
+++ b/pkg/dotc1z/c1file.go
@@ -736,8 +736,10 @@ func (c *C1File) finalize(ctx context.Context) error {
 	// operations here and proceeds directly to saveC1z. Checkpointing a
 	// released handle would fail with sql.ErrConnDone and send us down the
 	// cleanupDbDir branch, deleting the working database instead of saving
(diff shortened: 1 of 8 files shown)


kans:
Its probably better to just drop sqlite support than to continue to patch it.

PR #1093, opened 2026-08-13
fix(uhttp): opt-in cache-key headers via CacheOption + WithCacheKeyHeaders

Bencheng21:
## Summary
- `CreateCacheKey` only folded `Accept`, `Content-Type`, `Cookie`, and `Range` into the HTTP response cache key. Any other request header was silently ignored, so two GET requests differing only in an unlisted header (e.g. `Authorization`, a custom API-version or tenant header) collided on the same cache entry and one request could be served the other's cached response.
- `CreateCacheKey(req *http.Request, opts ...CacheOption)` -- a trailing variadic, so every existing zero-arg `CreateCacheKey(req)` / `cache.Get(req)` / `cache.Set(req, resp)` call keeps compiling unchanged. `CacheOption` is an interface rather than a concrete parameter, leaving room for future dimensions (TTL, query-param keying, etc.) without touching these signatures again.
- `WithCacheKeyHeaders(...)` is a `WrapperOption`, configured once at client construction and applied to every request that client makes -- not a per-call option:
  ```go
  cli := uhttp.NewBaseHttpClient(httpClient, uhttp.WithCacheKeyHeaders("Authorization"))
  ```
  `BaseHttpClient` stores the resulting `CacheOption`(s) and forwards them to every `Get`/`Set` call itself makes. `Do`'s signature is completely untouched -- no sibling method, no per-call option, every existing call site keeps working exactly as before.
- The value folded into the key is read from `req.Header` at key-computation time, not supplied by the caller -- so the key can never describe a token that isn't the one actually sent on that request.
- Because the opted-in header's value is read from `req.Header` (same source as the default set), the request-clone fix in `BaseHttpClient.Do` matters here: `http.Client.Do` forks the `*http.Request` struct on every call once `Timeout > 0` (which `uhttp.NewClient` always sets), but that fork is shallow, so `Header` stays the same map the caller passed in -- a `RoundTripper` mutating it between the cache's `Get` and `Set` calls (`userAgentTripper` does this for `User-Agent`) would otherwise make the `Set` key diverge from every future `Get` key for an opted-in header. Fixed by round-tripping on `req.Clone(req.Context())` instead of `req`.

## Alternatives considered
Two other API shapes for the same fix were explored and are open for comparison:
- #1083: header *names* passed via a `CallOption` on `Do` itself (`Do(req, uhttp.WithCacheKeyHeaders(...), ...)`), values still read from `req.Header`.
- #1092: a `DoWithCacheKeyHeaders(req, map[string]string, ...)` sibling method, values supplied directly by the caller rather than read from `req.Header` (traded the request-clone fix away, at the cost of the key being able to describe a value that doesn't match what's on the wire).

This PR's design keeps `Do` and every existing call site fully untouched, matches values to what's actually sent, and leaves room to grow via `CacheOption` without further signature churn -- at the cost of a per-client rather than per-call configuration point.

## Test plan
- [x] `pkg/uhttp/client_test.go`: default set unaffected by headers outside it; a `CacheOption` opts a header in (case-insensitively) without widening the key to everything else. All pre-existing zero-arg `CreateCacheKey`/`Get`/`Set` calls compile and pass unmodified.
- [x] `pkg/uhttp/wrapper_test.go`: `TestWrapper_WithCacheKeyHeaders_DistinguishesRequests` confirms different `Authorization` values no longer collide while the same value still hits cache; `TestWrapper_Do_CachesDespiteRoundTripHeaderInjection` confirms the request-clone fix is load-bearing -- fails without it, passes with it.
- [x] `go test ./pkg/uhttp/...` and `go test ./...` (whole module)
- [x] `golangci-lint run ./pkg/uhttp/...`
- [x] `go build ./...`


kans:
Fable:

# #1093 (cache-option-wrapper) — required changes

1. Drop the unconditional `req.Clone` in `Do` (#1092 already reverted this).
   It doesn't help caller-set headers, and for headers injected mid-flight
   (jar cookies — `Cookie` is in the default key set — or in-place
   RoundTrippers) it turns a dead cache into cross-scope cache hits: the
   CE-1056 bug reintroduced on the default path.

2. Delete `TestWrapper_Do_CachesDespiteRoundTripHeaderInjection` — it pins
   the dangerous behavior (cache hit while ignoring an opted-in header).

3. Export a `CacheOption` constructor, e.g.
   `func CacheKeyHeaders(headers ...string) CacheOption`, and have
   `WithCacheKeyHeaders` wrap it. Today `CacheOption` appears in exported
   signatures (`CreateCacheKey`, `GoCache/DBCache.Get/Set`) but can't be
   constructed outside `uhttp`; the client_test builds the unexported type
   directly.

4. Doc `WithCacheKeyHeaders`: named headers must be set on the request
   before `Do`; transport-injected headers are not seen. Nice-to-have:
   warn in `Do` when a configured header is absent from `req.Header`.

5. Dedupe opted-in names against the default set and each other
   (`WithCacheKeyHeaders("Accept")` currently folds values in twice).

6. Pin the cache env in the new wrapper tests
   (`t.Setenv BATON_DISABLE_HTTP_CACHE=false`, backend=memory) so they
   don't fail under ambient config.

7. Rebase onto main (only XML-helper drift in wrapper.go; merges clean).

8. Minor version bump per repo compat rules (variadic change to exported
   `Get/Set` signatures breaks method-value assignments downstream).

Keep: option-pattern API, values read from `req.Header`, client-level
opt-in. Don't close CE-1056 on merge — the fix lands only when the
header-scoped connectors (sendgrid et al) adopt `WithCacheKeyHeaders`.

laurenleach (approved):

PR #1099, opened 2026-08-20
fix(pebble): preserve resource icon asset refs

jugonzalez12:
Pebble v3 stored ResourceRecord as decomposed fields but had no storage slot for the top-level v2.Resource.icon field. As a result, resource-level icon references were dropped during v2-to-v3 translation and could not be restored when hydrating resources back out of Pebble.

Add icon_asset_external_id to ResourceRecord so the storage shape keeps the asset external ID directly, matching AssetRecord.external_id without adding a one-field v3 ref type. Map v2.Resource.icon.id into that field on write and rebuild v2.AssetRef from it on read.

Extend both the translation round-trip test and the full Pebble persistence path test so profile, status, created_at, and icon all survive the resource write/read flow. Regenerate the storage v3 protobuf outputs.

### DIFF
--- a/pb/c1/storage/v3/records.pb.go
+++ b/pb/c1/storage/v3/records.pb.go
@@ -591,8 +591,10 @@ type ResourceRecord struct {
 	// to 12 when profile/status/created_at (9-11) landed on main first.
 	// No artifact was ever written with the old number.
 	SourceScopeKey string `protobuf:"bytes,12,opt,name=source_scope_key,json=sourceScopeKey,proto3" json:"source_scope_key,omitempty"`
-	unknownFields  protoimpl.UnknownFields
-	sizeCache      protoimpl.SizeCache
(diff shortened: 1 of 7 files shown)


kans:
Do we need to mark the pebble version or update it so thtat we'll know where to look for icons?

jugonzalez12:
> Do we need to mark the pebble version or update it so thtat we'll know where to look for icons?

agents say no:

"The only thing a version marker could tell us is provenance: “this file was written by a version that preserves resource icons.” It would not tell
  us where to recover icons from in old Pebble files, because pre-fix Pebble already dropped the reference. So unless we specifically need diagnostics
  like “icons may be incomplete for artifacts written before X,” I’d avoid adding a version bump."
  
defer to you.

kans (approved):

PR #1115, opened 2026-08-31
Sync Formal Modeling - P & Occult

kans:
# Source-cache replay (phase 6b) + formal verification of sync scheduling

**Status: not for merge as-is.** This branch contains the phase 6b
implementation *and* the formal verification work that characterizes a
structural defect in its composition algorithm. The current algorithm can
seal artifacts containing permanently dead data presented as live (details
below), and that data then replays into subsequent warm syncs. The merge
path is the replacement runtime designed and verified on this branch
(RFC 0011), not the 6b composition as it stands. The branch is published as
the archive of record for both the implementation and its verification.

## What's here

Three related bodies of work.

### 1. Source-cache replay orchestration (phase 6b) — `pkg/`, `internal/`, `proto/`

Warm syncs can reuse partitions of the previous sync's artifact instead of
refetching them. Connectors annotate page responses to either **record** a
partition (fresh rows, tagged with a validator string) or **replay** it
(copy the previous artifact's partition forward, optionally overlaying
changed rows and tombstoning deleted ones). The syncer orchestrates:

- a lookup negotiation (the connector may ask for previous-artifact
  evidence before deciding; bounded to 4 round-trips per request);
- atomic replay units (partition clear + copy + validator publish commit
  together);
- integrity verdicts that fall back to a cold fetch on any doubt;
- crash/resume safety (interrupted replays re-run idempotently).

Touches `pkg/sync` (orchestration), `pkg/dotc1z` / `pkg/sourcecache`
(storage), `proto`/`pb` (annotations), `internal/chaosconnector` (test
harness). Verification evidence: `docs/verification/sync-replay-6b/`.

### 2. Formal verification of the scheduling semantics — `formal/` (no production code)

Two tools, for readers unfamiliar with them:

- **P** is a model checker: you describe a system as communicating state
  machines plus properties that must hold, and it mechanically explores
  very large numbers of crash and interleaving schedules looking for a
  violation, printing the exact schedule when it finds one.
- **Occult** is a proof engine: you write algebraic axioms about the
  system's operations and it proves or refuses claimed identities; it can
  also evaluate invariant-checking functions over recorded event logs.

What was done with them:

- **A calibrated model of the current (6b) design.** The model earns
  trust by mechanically rediscovering known bugs before being believed
  about anything else. Result: with the shipped mitigations ON, the
  checker still seals a **phantom union** — individually truthful
  connector responses composing into an artifact containing rows that no
  longer exist upstream, which then replay warm into later syncs as
  live data with unbounded staleness. This is the defect that blocks
  merging 6b's composition as-is (`formal/walker/CALIBRATION.md`,
  scenario 1; each red cell archives its counterexample schedule).
- **A model of the replacement design** (demand-graph runtime: work is
  admitted by tracked demand, every output carries lineage, seal-time
  obligations are checked against a closure oracle). Model-checking found
  five design bugs during calibration — all fixed in the spec before any
  code exists — and a structured bake-off between two lineage designs
  selected **observable-causal stamps** (`formal/graph/BAKEOFF.md`).
- **Machine-checked proofs** of 15 algebraic laws of the replay/overlay
  composition, a deductive derivation showing the current composition
  rule *manufactures* the phantom union, and conformance proofs for the
  lookup negotiation protocol (`formal/occult/`).

Entry point: `formal/REPORT.md` (verdicts, findings register, and an
explicit statement of what is and is not guaranteed).

### 3. The bridge: checking real executions against the formal invariants

`pkg/sync` gains a test-only event recorder
(`pkg/sync/sync_trace_audit.go`). It is a nil pointer in production —
only test code can enable it, and test files are not compiled into
production binaries; each call site costs one nil check. Chaos tests run
the real syncer, export the recorded commit-order event logs, and an
Occult test suite verifies every exported execution obeys the ordering
and durability invariants (writes only into partitions cleared this sync,
publish before a checkpoint claims quiescence, seal only after
obligations are met — across cold, warm, crash/resume, and tombstone
executions). This instrument also **falsified one piece of documented
resume behavior**: a documented "skip the re-copy on resume" mechanism is
unreachable; the code is safe via idempotent re-copy instead. Comments
were corrected (`pkg/sync/chaos_source_cache_resume_test.go`).

### Also included

`docs/rfcs/0011-demand-graph-sync-runtime.md` — the kickoff design doc
for implementing the verified replacement runtime in `pkg/sync`.

## Reviewer routing

Per `docs/REVIEW_CHECKLIST.md`: the production-behavior changes are item
1 plus the nil-guarded recorder call sites from item 3 — route HIGH
(silent/combinatorial subsystem). Item 2 is additive models and
documentation; the rest of item 3 is test code and fixtures.


kans:
subsumed into other work

PR #1121, opened 2026-09-03
Stamp grant digest state with its hash ABI version and enforce it at Open

mj-palanker:
## Summary

`GrantDigestABIVersion` names the framing of the grant content hash and the principal bucket hash, but nothing on disk recorded which version a file's digest nodes and hash-index rows were computed under. A future bump would have had no way to tell current state from state built by older hash code, short of an index migration, which older binaries can re-pollute after it has run once. This PR adds the durable bookkeeping needed to make such a bump safe on its own, and closes two gaps in how its consequences reach consumers. The constant stays at 1; no hash changes here.

## Changes

**ABI stamp** (`pkg/dotc1z/engine/pebble`)

- A durable stamp key (`rawdb.GrantDigestABIStampKey`, uint32 BE) lives inside the digest keyspace under a reserved meta index id. Every wholesale destroyer of digest state, including those in already-shipped SDKs, erases it without knowing it exists. The presence probe now spans only the node keyspace, so a leftover stamp over an empty file arms nothing.
- Every global-root write site writes the stamp first, so a durable root is never uncertified.
- Open reads the stamp after the presence probe. A writable open drops all digest state whose stamp does not name the current version, and the next `EndSync` rebuilds from scratch through the existing digests-absent path. A read-only open cannot drop, so it sets `grantDigestAbiStale` and the digest root getters report "never built", the same fail-safe shape as the interrupted-build marker.
- A missing stamp over present digest nodes reads as version 1. Every SDK build that predates the stamp hashed at that version, so introducing the stamp costs existing files no rebuild. An unstamped file stays unstamped until its first global-root rewrite, which is fine: absence keeps reading as 1.

**Consumers of the untrusted-state flags**

- The stale flag and the pre-existing build-pending flag are folded into one `grantDigestStateUntrusted` helper, consulted by both root getters and by `ComputeEntitlementBucketDigest`, which previously folded hash-index rows with no check and would have returned digests derived from another hash scheme's content on a read-only open of a stale file.
- The bucket scan is deliberately left ungated. Bucket membership is the principal bucket hash over the encoded principal identity, frozen with the v3 key encoding and unchanged by any ABI bump, so a stale file's bucket placement is still exact.

**Fold compaction** (`pkg/synccompactor`)

- A fold that writes no grants used to leave the base's copied digest state alone unconditionally. That is right when the state is present, but the dest's writable Open may have just dropped it, or the base may never have had any (digest index disabled at seal, or an earlier build failure). Such a fold shipped an output with no digest state and nothing downstream to repair it. When the digest index is enabled and no state is present, the fold now runs the repair, which delegates to the full build: a one-time O(base) cost the first time a digest-less base is folded, after which the output is stamped and later folds pay nothing extra.

## Rollout behavior

- Existing files open with no drop and no rebuild, since a missing stamp reads as the current version 1.
- Files get stamped on their next global-root rewrite (any seal that repairs or rebuilds digests).
- A later bump of the constant is then sufficient by itself: stamped-at-1 and unstamped files alike are dropped and rebuilt at the new version on their next writable use, and a binary rolled back across the bump re-stamps at its own version on its next repair, so the newer binary detects the mixed state instead of trusting it.

## Docs

Three older comments that said a hash framing change "requires an index-migration bump" now point at `GrantDigestABIVersion`; `index_migrations.go` explains why digest ABI bumps do not belong there.

## Tests

- `TestGrantDigestABI*` in `grant_digest_abi_test.go`: stamp written by seal and excluded from the presence probe; stale stamp dropped on writable open and rebuilt at the current version; missing stamp reads as version 1 and is kept, then stamped on the first root rewrite; read-only open sets the stale flag and root getters report not-built; orphan stamp over an empty keyspace is ignored; stale stamp plus interrupted-build marker together; bucket digest gated while bucket scan still yields every grant; plus an oracle test proving the hash-index verifier can detect tampering.
- `TestCompactPebbleFoldRebuildsDigestsWhenBaseHasNone`: a digest-less base folded with a partial that writes no grants ends with a present global root byte-equal to both the original base's root and a from-scratch oracle build.

## Reviewer note

The fold change is a compactor behavior change: the O(base) build now fires on the first fold of any digest-less base, not only after an ABI bump, including a base sealed with the digest index disabled and then folded by an engine with it enabled.

### DIFF
--- a/pkg/dotc1z/engine/pebble/digest.go
+++ b/pkg/dotc1z/engine/pebble/digest.go
@@ -494,12 +494,15 @@ type DigestRoot struct {
 // computeBucketDigest would read that absence as "zero records" — the
 // false-clean trap dirtyPartitionBuckets' doc comment describes.
 func (e *Engine) getPartitionDigestRoot(spec digestIndexSpec, partition string) (DigestRoot, bool, error) {
-	if e.grantDigestBuildPending.Load() {
+	if e.grantDigestBuildPending.Load() || e.grantDigestAbiStale.Load() {
 		// An interrupted digest build's half-committed nodes may be
 		// durable while its hash index never ingested; until the pending
 		// state is consumed (a writable Open drops it; a read-only open
 		// cannot), no stored root may be trusted — report "never built",
-		// which every consumer already treats as "recalculate".
+		// which every consumer already treats as "recalculate". Roots
+		// computed under a different hash ABI (grantDigestAbiStale, only
+		// ever set on read-only opens) are equally untrustworthy: their
+		// hashes come from a different input framing.
 		return DigestRoot{}, false, nil
 	}
 	val, closer, err := e.db.Get(encodeDigestNodeKey(spec.indexID, partition, digestLevelRoot, nil))
--- a/pkg/dotc1z/engine/pebble/digest.go
+++ b/pkg/dotc1z/engine/pebble/digest.go
@@ -494,15 +494,10 @@ type DigestRoot struct {
 // computeBucketDigest would read that absence as "zero records" — the
 // false-clean trap dirtyPartitionBuckets' doc comment describes.
 func (e *Engine) getPartitionDigestRoot(spec digestIndexSpec, partition string) (DigestRoot, bool, error) {
-	if e.grantDigestBuildPending.Load() || e.grantDigestAbiStale.Load() {
-		// An interrupted digest build's half-committed nodes may be
-		// durable while its hash index never ingested; until the pending
-		// state is consumed (a writable Open drops it; a read-only open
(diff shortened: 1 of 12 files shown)


kans:
What happens if we open a new C1z with an old sdk with regard to the ABI stamp?

mj-palanker:
> What happens if we open a new C1z with an old sdk with regard to the ABI stamp?

It should silently ignore the stamp, and if it needs to repair or rebuild, it will do so without the ABI version, un-stamping it in effect.
but it's not fatal or anything.
I need to get this version aware support in so that we can safely upgrade without getting mixed version info.

kans (approved):

PR #1132, opened 2026-09-10
CXP-1085 uhttp: classify OAuth2 token-exchange failures

agustin-conductor:
Classifies `golang.org/x/oauth2` token-exchange failures so `retry.Retryer` can
act on them: a token endpoint that answered 500 or timed out is now
`Unavailable`/`DeadlineExceeded` and gets backed off through, while a rejected
credential stays `Unauthenticated`/`PermissionDenied` and still fails fast.

Fixes [CXP-1085](https://linear.app/ductone/issue/CXP-1085), which surfaced
under [CXP-890](https://linear.app/ductone/issue/CXP-890).

## Why

Token failures reached the syncer as `codes.Unknown`, which `retry.Retryer`
treats as terminal — `pkg/retry/retry.go:61` waits only on `Unavailable` and
`DeadlineExceeded`. The sync action aborted and `IsSyncPreservable` discarded the
artifact, so a sub-second blip at `oauth2.googleapis.com/token` cost hours of
completed grant walking. GCP re-exchanges a service-account JWT roughly hourly,
so longer syncs failed more often.

The classification already existed and was unreachable. `wrapTransientNetworkError`'s
`*oauth2.RetrieveError` branch (added in #1106) only ran from `Transport.RoundTrip`
and `wrapper.Do` — and a non-2xx from a token endpoint is a *successful*
RoundTrip, so x/oauth2 builds the `RetrieveError` above the transport, where that
branch never sees it. This is not GCP-specific: every helper in
`pkg/uhttp/authcredentials.go` had the same gap.

## What changed

- `pkg/uhttp/oauth2.go` (new) exports `ClassifyOAuth2TokenError` and
  `NewClassifyingTokenSource`, for connectors that build their own `jwt.Config`
  rather than going through the SDK helpers (baton-google-cloud-platform does).
- `OAuth2ClientCredentials`, `OAuth2JWT` and `OAuth2RefreshToken` now build their
  token source through the wrapper. `OAuth2RefreshToken` spells out
  `cfg.TokenSource` + `oauth2.NewClient`, which is what `cfg.Client` does.
- `errors.go`'s `RetrieveError` branch delegates to the same function, so the
  transport path and the token-source path share one implementation.

Two x/oauth2 behaviors had to be worked around, both pinned by tests that drive a
real `jwt.Config` against `httptest` — if a future bump fixes either, the test
fails rather than leaving dead code:

1. **The RFC 6749 params are empty.** `jwt.Config`'s token source builds
   `RetrieveError` from the response alone (`jwt/jwt.go:143`), so `error` and
   `error_description` are recovered from the body, JSON or form-encoded. RFC 6749
   §5.2 defines `error` as a string; a body that puts something else there
   (Google's endpoint nests a `code`/`message`/`status` object) carries no param to
   act on, so the code falls back to the HTTP status and the bounded body becomes
   the description. Nothing keys on a vendor's field names.
2. **`%v` flattening.** `jwt/jwt.go:135`, `:140` and `:154` wrap transport failures
   with `%v`, not `%w`, destroying the status the uhttp transport attached before
   any caller sees it; `errors.As` cannot recover it and no dependency bump fixes
   it. `classifyFlattenedOAuth2TokenError` therefore reads message text, gated on
   the `oauth2: cannot fetch token` prefix. It first recovers a grpc status out of
   the flattened text — that code was decided from the typed error — then falls
   back to a substring table whose codes must agree with
   `wrapTransientNetworkError`'s, asserted case by case, since the same failure
   takes either path depending only on whether the caller installed this package's
   transport under `oauth2.HTTPClient`. `pkg/lambda/grpc/util.go:178` already
   classifies serialized errors this way for the same reason.

`pkg/retry`'s retryable set is untouched, per the CXE-1113 anti-goal: that retryer
has no attempt limit at its defaults, so a disabled credential must fail fast
rather than spin for hours.

## Effect on existing connectors

Nothing breaks at compile time — the exported surface only grew — and nothing
breaks in error matching, because the status is attached by `errors.Join` rather
than substitution. `errors.As` still recovers `*oauth2.RetrieveError`, and
`err.Error()` still carries x/oauth2's own text and the response body, so
existing `errors.Is`/`errors.As` checks and message matches keep working.
`TestOAuth2JWT_GetClientPreservesErrorIdentity` pins that end to end.

Three behavior changes worth a reviewer's attention:

- **`IsSyncPreservable` (`pkg/sync/syncer.go:91`) now returns true for these
  errors where it returned false.** An unclassified error fails its
  `status.FromError` check, while `Unavailable`, `Unauthenticated` and
  `PermissionDenied` are all in its preserve list — so a partial artifact
  survives a token failure instead of being discarded. That is the direction
  CXP-890 wants, and it is a retention change, called out here rather than left
  to be discovered. The `InvalidArgument` cases (`invalid_scope`,
  `invalid_request`, NXDOMAIN) still discard, as before.
- **`server_error` and `temporarily_unavailable` flip from terminal to
  retryable.** RFC 6749 §4.1.2.1 defines both as transient; a 400 carrying one
  previously landed on `InvalidArgument`.
- **A permanently failing token endpoint now backs off instead of failing
  fast**, bounded by the caller's `RetryConfig` and run duration. Same exposure
  any upstream 503 already has.

## Testing

`go test ./pkg/...` passes; `golangci-lint run pkg/uhttp/...` is clean (the six
findings the repo-wide run reports are pre-existing on `main`, in packages this
PR does not touch). Tests cover the cross product of source shape (typed
`RetrieveError`, `%v`-flattened text, raw transport error), param location (on
the error, JSON body, form body, absent, non-string) and disposition, each
asserted against `retry.Retryer` rather than the grpc code alone. Every branch in
`oauth2.go` is covered. Ten planted mutants were each caught: dropped body
recovery, dropped flattened fallback, dropped grpc status recovery, `server_error`
made terminal, dropped truncation, dropped existing-status guard, non-string
`error` member used as a code, NXDOMAIN made retryable, dropped body snippet, and
lost `error_description` precedence.

### DIFF
--- a/pkg/uhttp/oauth2.go
+++ b/pkg/uhttp/oauth2.go
@@ -1,0 +1,363 @@
+package uhttp
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"mime"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+
+	"golang.org/x/oauth2"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/conductorone/baton-sdk/pkg/ratelimit"
+)
+
+// oauth2FetchTokenPrefix is the text golang.org/x/oauth2 puts on a token
+// request that failed before a token could be parsed out of the response.
+// It gates classifyFlattenedOAuth2TokenError, the one classifier here that
+// reads message text instead of error identity.
+const oauth2FetchTokenPrefix = "oauth2: cannot fetch token" //nolint:gosec // G101: a message prefix, not a credential
+
+// maxRecoveredDescription bounds an error_description recovered from a
+// response body before it becomes part of a grpc status message. x/oauth2
+// caps the body it reads at 1 MiB, and that whole body can be one JSON
+// string field.
+const maxRecoveredDescription = 256
+
+// ClassifyOAuth2TokenError attaches a grpc status to a golang.org/x/oauth2
+// token-exchange failure, so retry.Retryer can act on it: a token endpoint
+// that answered 500 or timed out is Unavailable/DeadlineExceeded (retryable),
+// while a rejected credential stays Unauthenticated or PermissionDenied
+// (terminal). Without it every token failure reaches the syncer as
+// codes.Unknown, which retry.Retryer treats as terminal, so a sub-second
+// token-endpoint blip discards a whole sync.
+//
+// Connectors that build their own oauth2.TokenSource call this directly or
+// go through NewClassifyingTokenSource; the AuthCredentials helpers in
+// authcredentials.go already do.
+func ClassifyOAuth2TokenError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// An error that already carries a status was classified by a producer
+	// closer to the failure, usually Transport.RoundTrip in this package.
+	// status.Code reads the outermost status, so re-wrapping here would
+	// bury the code the retryer needs behind this one.
+	if status.Code(err) != codes.Unknown {
+		return err
+	}
+
+	var retrieveErr *oauth2.RetrieveError
+	if errors.As(err, &retrieveErr) {
+		return classifyOAuth2RetrieveError(retrieveErr, err)
+	}
+
+	if classified := classifyFlattenedOAuth2TokenError(err); classified != nil {
+		return classified
+	}
+
+	return wrapTransientNetworkError(err)
+}
+
+// NewClassifyingTokenSource returns a TokenSource that runs every error from
+// inner through ClassifyOAuth2TokenError. Wrap the outermost source, the one
+// clientcredentials.Config.TokenSource / jwt.Config.TokenSource /
+// oauth2.Config.TokenSource return, so the reuse-and-refresh caching those
+// build stays in place and only a real token fetch is classified.
+func NewClassifyingTokenSource(inner oauth2.TokenSource) oauth2.TokenSource {
+	if inner == nil {
+		return nil
+	}
+	return &classifyingTokenSource{inner: inner}
+}
+
+type classifyingTokenSource struct {
+	inner oauth2.TokenSource
+}
+
+var _ oauth2.TokenSource = (*classifyingTokenSource)(nil)
+
+func (c *classifyingTokenSource) Token() (*oauth2.Token, error) {
+	token, err := c.inner.Token()
+	if err != nil {
+		return nil, ClassifyOAuth2TokenError(err)
+	}
+	return token, nil
+}
+
+// classifyOAuth2RetrieveError maps a token endpoint's rejection to a grpc
+// code. It is reached from ClassifyOAuth2TokenError, which x/oauth2's token
+// sources feed, and from wrapTransientNetworkError, for the callers that
+// drive a token endpoint through this package's transport directly.
+func classifyOAuth2RetrieveError(retrieveErr *oauth2.RetrieveError, err error) error {
+	tokenErr := oauth2TokenErrorFrom(retrieveErr)
+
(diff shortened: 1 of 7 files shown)


pkg/uhttp/oauth2.go
```diff
@@ -0,0 +1,384 @@
...
+	// A transient token-endpoint status (429/5xx) stays retryable even if
+	// the body also carries a recognized RFC 6749 error param; otherwise
+	// the error param takes priority over the HTTP status, since some
+	// servers report it on a 200 and 400 is the spec default for
+	// invalid_client/invalid_grant, both of which GrpcCodeFromHTTPStatus
+	// alone would misclassify.
+	if tokenErr.resp != nil && isTransientHTTPStatus(tokenErr.resp.StatusCode) {
+		return wrapTransientOAuth2TokenError(tokenErr, err)
+	}
+	if code, ok := oauth2TokenErrorCode(tokenErr.code); ok {
```
mateoHernandez123: 🔴 A terminal HTTP authentication status can be overridden by the response body here. For example, a `RetrieveError` with HTTP 401 and `ErrorCode: "server_error"` becomes `codes.Unavailable`; `retry.Retryer` can then retry indefinitely with its default `MaxAttempts == 0`. CXP-1085 requires 401 to remain `Unauthenticated` so rejected credentials fail fast, and RFC 6749 §5.2 does not define `server_error` or `temporarily_unavailable` for token endpoint responses. Please give 401/403 precedence over these body values and add contradictory status/body test rows.

pkg/uhttp/oauth2.go
```diff
@@ -0,0 +1,384 @@
...
+	if tokenErr.description != "" {
+		msg = fmt.Sprintf("%s: %s", msg, tokenErr.description)
+	}
+	st := status.New(GrpcCodeFromHTTPStatus(tokenErr.resp.StatusCode), msg)
+	if description, rlErr := ratelimit.ExtractRateLimitData(tokenErr.resp.StatusCode, &tokenErr.resp.Header); rlErr == nil {
+		if withDetails, detailsErr := st.WithDetails(description); detailsErr == nil {
+			st = withDetails
+		}
+	}
+	return errors.Join(st.Err(), err)
```
mateoHernandez123: 🔴 `errors.Join(st.Err(), err)` defeats `maxErrorParamLength` at the actual gRPC boundary. When `status.FromError` finds the wrapped status, grpc-go replaces its message with `err.Error()`, which still includes the complete `RetrieveError` description/raw response body (up to 1 MiB). `pkg/tasks/c1api/manager.go` then logs and sends `status.FromError(taskError).Message()` to C1. Please preserve `errors.Is`/`errors.As` with a wrapper whose `Error()` is bounded, and test `status.FromError(result).Message()` with a sensitive marker beyond the cap rather than only inspecting the nested `GRPCStatus()`.

pkg/uhttp/oauth2.go
```diff
@@ -0,0 +1,384 @@
...
+	var retrieveErr *oauth2.RetrieveError
+	if errors.As(err, &retrieveErr) {
+		return classifyOAuth2RetrieveError(retrieveErr, err)
+	}
+
+	if classified := classifyFlattenedOAuth2TokenError(err); classified != nil {
+		return classified
+	}
+
+	return wrapTransientNetworkError(err)
```
mateoHernandez123: 🟠 The fallback does not always preserve the original error identity promised by this exported helper. `ClassifyOAuth2TokenError(context.DeadlineExceeded)` reaches `wrapTransientNetworkError`, whose bare-context branch returns a fresh `status.Error`; afterward `errors.Is(result, context.DeadlineExceeded)` is false. Please join/wrap the status with the original deadline error and assert both `status.Code` and `errors.Is` for canceled/deadline inputs.

ggreer (approved):

PR #1139, opened 2026-09-16
field: add WithMultiline to render a string config field as a textarea

haroonkhan-c1:
## Summary

- Adds `STRING_FIELD_TYPE_MULTILINE` so a connector can declare that a string config field spanning lines (e.g. a PEM-encoded private key) should render as a textarea in the ConductorOne config UI instead of a single-line input.
- `field.MultilineField(name, opts...)` (`pkg/field/fields.go`), mirroring `RandomField`/`Oauth2Field` — identical to `StringField` except for `FieldType`.
- `is_secret` still composes: it lives on `Field`, not `StringField`, so `MultilineField("pem", field.WithIsSecret(true))` renders a masked textarea.

**Update:** this PR originally proposed `bool multiline = 7` on `StringField` instead. [@ggreer](https://github.com/ggreer) and [@laurenleach](https://github.com/laurenleach) pointed out in review that `type` doesn't modify a text box, it selects which widget renders entirely — `FILE_UPLOAD` produces a dropzone, `OAUTH2` an OAuth flow widget, `CONNECTOR_DERIVED_OPTIONS` a select. A `multiline` bool was only meaningful for 1 of the 5 existing values and silently discarded on the other 4, i.e. 4 representable-but-illegal states. This PR now implements their suggestion instead.

## Why a new enum value rather than the bool

Same tradeoff called out originally, minus the part that didn't hold up: `is_secret` composes with any `StringFieldType` regardless of this choice, since it's a `Field`-level property. What doesn't survive scrutiny is composing `multiline` with the *other* string types — there's no such thing as a multiline OAuth2 widget or a multiline file dropzone.

Leaving `FieldType`'s existing consumers alone was the other stated goal, and that still holds with the enum approach: `pkg/config/generate.go`'s `[]byte`-vs-`string` switch keys only on `FieldType == FileUpload`, so `MultilineField` still generates a plain Go `string` and no connector needs consumer-side changes.

**Update:** per follow-up review, extracted that switch into `goTypeForField(f field.SchemaField) string` — a pure function is directly unit-testable, where `Generate()` itself isn't (it panics on error and writes to a hardcoded `conf.gen.go` path rather than a parameter).

## Test plan

- [x] `go build ./...`, `gofmt -l pkg/field/ pkg/config/`
- [x] `go test ./pkg/field/... ./pkg/config/...`
- [x] `TestMultiline` asserts `MultilineField(..., WithIsSecret(true))` marshals to `STRING_FIELD_TYPE_MULTILINE` with `is_secret` preserved, and contrasts against a plain `StringField` staying `TEXT_UNSPECIFIED`
- [x] `TestGoTypeForField` (`pkg/config/generate_test.go`) tables every constructor against its generated Go type — `MultilineField` → `string`, `FileUploadField` → `[]byte`, and the rest — closing the gap where that invariant was previously only checked by hand
- [x] Verified end to end against a consumer: with this change vendored, `baton-okta config` now emits `"type": "STRING_FIELD_TYPE_MULTILINE"` on `okta-private-key` (no `"multiline"` key), `isSecret` preserved, and its generated config struct still types the field as `string`
- [x] Confirmed no breaking-change finding: field `7` (the removed bool) never merged to `main`, so CI's `breaking_against: main` baseline has nothing to flag


proto/c1/config/v1/config.proto
```diff
@@ -189,4 +189,8 @@ message StringField {
   repeated string allowed_extensions = 4;
   repeated StringFieldOption options = 5;
   string suggested_value = 6;
+
+  // Render the input as a multiline textarea. Orthogonal to type and to
+  // Field.is_secret, so a secret field can also be multiline.
+  bool multiline = 7;
```
ggreer: Would it make sense to have this be a new StringFieldType like STRING_FIELD_TYPE_MULTILINE? I don't think multiline can be composed with any of the other string field types like STRING_FIELD_TYPE_FILE_UPLOAD, for example. It would also still be composable with is_secret.

I don't know if my suggestion makes the code simpler or not.

laurenleach: I like this idea ^

laurenleach (approved):

ggreer (approved):

